PR #28, opened 2022-12-06
schema performance tuning

bkmartinjr:
Changes:
* added hooks for using platform_config to tune Census schema (i.e., do not use default SOMA schema)
* first round of schema tuning - primarily filter pipeline configuration

Unrelated, small changes:
* broaden Python versions supported (changes to type static typing and type checks in the GH action)
* parallelize consolidation pass to reduce build time
* disallow python 3.11 support until numba is updated

### DIFF
--- a/cell_census_builder/globals.py
+++ b/cell_census_builder/globals.py
@@ -64,16 +64,69 @@ CENSUS_OBS_TERM_COLUMNS = {
     "tissue_general_ontology_term_id": pa.large_string(),
 }
 
+_RepetativeStringLabelObs = [
+    # these columns are highly repetative string lables and will have appropriate filter
+    "assay",
+    "assay_ontology_term_id",
+    "cell_type",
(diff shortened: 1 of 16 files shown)


cell_census_builder/globals.py
```diff
@@ -64,16 +64,82 @@
     "tissue_general_ontology_term_id": pa.large_string(),
 }
 
+_RepetativeStringLabelObs = [
+    # these columns are highly repetative string lables and will have appropriate filter
```
atolopko-czi, line 68:
```suggestion
    # these columns are highly repetative string labels and will have appropriate filter
```

atolopko-czi (approved):
LGTM! 

Are there any dev artifacts (docs, code) worth linking out to in order to capture how the custom platform config params were tuned? Or even just a few comments in the code (e.g. why zstdfilter levels 5 and 19 are deemed best, per your earlier explanations).

PR #29, opened 2022-12-06
docs: add comp bio to notebook to learn about the census

pablo-gar:
Adds a notebook for comp bios that explains at a high-level the structure of the cell census and how to obtain summary information from it.


bkmartinjr (approved):
This looks great.  Two small suggestions for the future:
* once the bool query bug is fixed in SOMA/TileDB, I would use the SOMA DataFrame query rather than Pandas DataFrame query.  It will be faster as you don't have to read the entire dataframe.
* cell_census and SOMA are slowly getting better docstrings as their prelim documentation.  Maybe make a nod to that in the intro?   The `help()` function works great for that, eg, `help(cell_census.open_soma)` will print out docs

pablo-gar:
> This looks great. Two small suggestions for the future:
> 
> * once the bool query bug is fixed in SOMA/TileDB, I would use the SOMA DataFrame query rather than Pandas DataFrame query.  It will be faster as you don't have to read the entire dataframe.
> * cell_census and SOMA are slowly getting better docstrings as their prelim documentation.  Maybe make a nod to that in the intro?   The `help()` function works great for that, eg, `help(cell_census.open_soma)` will print out docs

Thanks Bruce, I have included the second suggestions.

I will incorporate the first suggestion once #30 is fixed

PR #38, opened 2022-12-12
Add unit tests

### LINKED ISSUE #9: unit tests for cell_census package
The package requires unit tests and pytest configuration.

bkmartinjr:
Additional unit tests to provide minimal coverage of the API.  Tests added for:
* open_soma, get_source_h5ad_uri and download_source_h5ad
* online mean, var
* get_anndata

Tests for experiment_query are deferred as that code is moving into SOMA core.

this PR also fixes a typing bug in `get_anndata`, which was exposed by the addition of a unit test

Fixes #9


atolopko-czi (commented):
Is it worth distinguishing between unit and system tests (against real data) via separate subdirs? It may ease how/when we run these different types of tests.

(Will CR further tomorrow)

bkmartinjr:
> Is it worth distinguishing between unit and system tests (against real data) via separate subdirs? It may ease how/when we run these different types of tests.

Potentially.  However, all of these are intended to be used in the unit test CI (i.e., they are specifically designed to be small & reasonably fast).

Feel free to make a concrete proposal of what you are thinking.  I'm flexible.

atolopko-czi:
> > Is it worth distinguishing between unit and system tests (against real data) via separate subdirs? It may ease how/when we run these different types of tests.
> 
> Potentially. However, all of these are intended to be used in the unit test CI (i.e., they are specifically designed to be small & reasonably fast).
> 
> Feel free to make a concrete proposal of what you are thinking. I'm flexible.

Let's use https://docs.pytest.org/en/7.1.x/example/markers.html. Even without having separate unit/system run triggers, it documents the nature of the test. Just add `@pytest.mark.live_corpus` to each test, as needed, e.g..  We can then document the meaning of `live_corpus` marker in `pytest.ini` per [docs](https://docs.pytest.org/en/7.1.x/example/markers.html#registering-markers).

PR #43, opened 2022-12-13
add census_build_cube.py example code

atolopko-czi:
If you think this example code is worth adding, I'll re-create as a proper notebook.

### DIFF
--- a/api/python/notebooks/census_build_cube.py
+++ b/api/python/notebooks/census_build_cube.py
@@ -41,7 +41,7 @@ if __name__ == '__main__':
             var_df = query.var().to_pandas().set_index("soma_joinid")
             obs_df = query.obs(column_names=["soma_joinid"] + cube_dims_obs).to_pandas().set_index("soma_joinid")
 
-            X_stats: pd.DataFrame = None
+            X_stats_all: pd.DataFrame = None
             for X_tbl in query.X("raw"):
                 print(f"Processing X batch size={X_tbl.shape[0]}")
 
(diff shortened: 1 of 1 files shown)


api/python/notebooks/census_build_cube.py
```diff
@@ -0,0 +1,76 @@
...
+                else:
+                    # Series.add() requires dense array, but maybe there's a better way to do this
+                    X_aggs['sum'] = X_aggs['sum'].sparse.to_dense()
+                    X_stats = X_stats.add(X_aggs, fill_value=0)
+
+                print(f"Batch agg rows={X_aggs.shape[0]}")
+                print(f"Total agg rows={X_stats.shape[0]}")
+
+            X_stats.to_pickle('census_cube.pkl')
+            print(X_stats)
```
atolopko-czi, line 76: If converted to a notebook, will provide some example queries of the cube.

bkmartinjr (approved):
LGTM.

I leave the question of notebook conversion to @pablo-gar

atolopko-czi:
> Who would be the audience for the notebook?

Maybe no one. :) This code was the result of my own experimentation, so did not have a clear audience in mind. However, if we wanted to share how one might query the data akin to our Gene Expression™️ product, this provides the core aggregation logic. It would need rankit and other filtering logic to be added, of course. 

I'm hardly pushing to have this in the notebooks dir, so feel free to just reject this PR if not worth including, and I'll stash the code elsewhere.

> Also naive question here, is the cube materialized in memory and then is gone at the end of the execution?

 Currently, it's just pickled. But the idea is that we could provide a few Gene Expression™️ -like filters for demonstration purposes.

pablo-gar (approved):
Got it, yeah I'm not opposed to it being a notebook. Having what you just described as and intro to the notebook can be helpful. Eventually when go into utilizing the Cell Census for internal cases (e.g. Gene Expression) this can be a good reference but I think ultimately the engs will come to you and ask how to this or that.

I can see how it could also be helpful for other non cellxgene folks to start experimenting with aggregations.

pablo-gar:
@atolopko-czi in light of recent conversations about keeping the API light and reducing maintenance for functions we won't support, I recommend we don't merge this and close it up.

any concerns with that?

atolopko-czi:
Moved to a [gist](https://gist.github.com/atolopko-czi/c786485d0b8e851488b2c3aba219413a) for posterity

PR #45, opened 2022-12-14
docs: Add comp bio notebooks

pablo-gar:
Adds:

`comp_bio_explore_and_load_lung_data.ipynb` -- a notebook demonstrating a full exploration of a tissue of the cell census

`comp_bio_large_smartseq_dataset.ipynb` -- a notebook showcasing gene lengths and how they can be used to normalize data

Updates:

`comp_bio_census_info.ipynb` -- latest cell census build, small improvements


api/python/notebooks/comp_bio_census_info.ipynb
```diff
@@ -1056,13 +1056,13 @@
...
-       "33861074                   colon epithelial cell             True\n",
-       "33861075              enteric smooth muscle cell             True\n",
+       "41487772                         mesodermal cell             True\n",
+       "41487773                   colon epithelial cell             True\n",
+       "41487774                    enteroendocrine cell             True\n",
+       "41487775                   colon epithelial cell             True\n",
+       "41487776              enteric smooth muscle cell             True\n",
        "\n",
-       "[33861076 rows x 2 columns]"
+       "[41487777 rows x 2 columns]"
```
bkmartinjr, line 1065: Need to update the text below which now quotes the wrong number of cells.

bkmartinjr: perhaps write in such a way that it does not go immediately obsolete when we re-run the notebook?  E.g.,  "The above row count is the number of cells ..."

api/python/notebooks/comp_bio_census_info.ipynb
```diff
@@ -1103,7 +1103,10 @@
     }
    ],
    "source": [
-    "human_cell_types = human_cell_types.query(\"is_primary_data == True\")\n",
+    "human_cell_types = census[\"census_data\"][\"homo_sapiens\"].obs.read_as_pandas_all(\n",
+    "    column_names=[\"cell_type\"], value_filter=\"is_primary_data == True\"\n",
```
bkmartinjr, line 1107: Perhaps for the future, but we need to create a notebook that provides a quick tutorial on the `value_filter` syntax.

it is quite simple, and will eventually be covered by the SOMA docs.  But as an interim, perhaps in our notebooks?

bkmartinjr:
Apologies for not putting these inline in the notebook - the GH web view won't display files as large as the lung notebook.

On the lung notebook:

1. Cell [3] splits the value filter into two strings.  Not technically an error, but it is somewhat confusing why this is being done. I would make them a single string, i.e.
```python
lung_obs = census["census_data"]["homo_sapiens"].obs.read_as_pandas_all(
    value_filter="tissue_general == 'lung' and is_primary_data == True"
)
```

2. Quoting the number of results (eg, "You can see that there are a total of 1,194,348 unique lung cells in the Cell Census") will require constant updating of the notebook as these numbers change.  I would either remove the literal number from the text/prose, OR write code that prints it out so it is updated when we re-run the notebook.

There are already places in this notebook with out of date numbers, eg, "You can see that there is a total of 60,554 genes."  (real count: 60638)

4. Cell 14 uses a very inefficient way to count non-zeros in the presence matrix. Suggest code similar to that in cell 15 (ie., just sum across the dimension)

5. I realize you did it intentionally, but the use of a full table scan for 11000+ genes is going to be VERY slow with the current version of TileDB value filters.  It would be dramatically (>10X) faster to do the query based upon the var joinids, already available as `lung_var.soma_joinid`.

api/python/notebooks/comp_bio_census_info.ipynb
pablo-gar: Good idea I'll try to make one that shows the ways we can "easily" access `SOMADataFrame` for pandas operations.

bkmartinjr: You can find the full syntax with:
```
import tiledb
help(tiledb.QueryCondition)
```

api/python/notebooks/comp_bio_normalizing_full_gene_sequencing.ipynb
```diff
@@ -0,0 +1,393 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "liver_dataset_id = \"524179b0-b406-4723-9c46-293ffa77ca81\"\n",
```
bkmartinjr, line 65: I suspect notebook readers will want to know how you determined the dataset_id....   Is that something obvious, or does it warrant elaboration?

api/python/notebooks/comp_bio_normalizing_full_gene_sequencing.ipynb
```diff
@@ -0,0 +1,393 @@
...
+    "liver_adata.X[:5, :5].toarray()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "for i in range(liver_adata.X.shape[1]):\n",
```
bkmartinjr, line 219: This is an anti-pattern - you should use numpy broadcasting to do the division.  By way of example, to divide across axis 1:
```
In [10]: (np.array([[1,2,3],[4,5,6]], dtype=np.float32).T / np.array([2,3])).T
Out[10]: 
array([[0.5       , 1.        , 1.5       ],
       [1.33333333, 1.66666667, 2.        ]])```

In your code, you would do something like:
```
liver_adata.X = (liver_adata.X.T / liver_adata.var.feature_length).T
```

CAVEAT: I did not test the above with the sparse API, so try it first :-)

bkmartinjr (commented):
A number of clean-up suggestions and code simplifications.

api/python/notebooks/comp_bio_census_info.ipynb
pablo-gar: Fixed

api/python/notebooks/comp_bio_normalizing_full_gene_sequencing.ipynb
pablo-gar: I added a section for that clarifies this

api/python/notebooks/comp_bio_normalizing_full_gene_sequencing.ipynb
pablo-gar: I had to do some casting but it works

bkmartinjr:
This all looks good. I wanted to note that the super slow table scan in the lung notebook _could_ be worked around with chanzuckerberg/cell-census#52, if you wish to do so.  I do not have an ETA on when the table scan issue will be fixed, but even when it is complete, it will be slower (i.e., it requires a second scan of the data, when you have already done it in the previous cells).

bkmartinjr (approved):

PR #49, opened 2022-12-15
Refine unit test workflows

bkmartinjr:
Changes:
* Pin the cell_census package dependency for tiledbsoma to 0.5.0a2
* Set the supported python versions (in setup.cfg) to Python 3.8-3.10.  3.11 not supported yet due to numba, and python 3.7 not supported due to tiledbsoma.
* Fix & enable the  unit test workflow for Python matrix [Linux, MacOS] X [Py3.8, Py3.9, Py3.10]
* change package version to 0.0.1-rc0

### DIFF
--- a/api/python/cell_census/setup.cfg
+++ b/api/python/cell_census/setup.cfg
@@ -28,7 +28,7 @@ install_requires =
     numpy
     requests
     tiledb>=0.19.0
-    tiledbsoma==0.2.0rc0
+    tiledbsoma==0.5.0a2
     typing_extensions
     s3fs
     scikit-misc
--- a/api/python/cell_census/setup.cfg
+++ b/api/python/cell_census/setup.cfg
@@ -27,8 +27,8 @@ install_requires =
     numba
     numpy
     requests
-    tiledb>=0.18.1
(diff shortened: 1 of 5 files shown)


api/python/cell_census/setup.cfg
```diff
@@ -16,19 +17,18 @@ classifiers =
...
-python_requires = >= 3.7, < 3.11
+python_requires = >= 3.8, < 3.11
 install_requires =
     numba
     numpy
     requests
-    tiledb>=0.18.1
-    tiledbsoma>=0.5.*
+    tiledb>=0.19.0
+    tiledbsoma==0.5.0a2
```
atolopko-czi, line 31: Did you want `0.2.0rc0` now? (per PR summary)

atolopko-czi (approved):
LGTM, but confirm tiledbsoma version, which is currently 0.5.0a2, not 0.2.0rc0.  Unless those are the same?

api/python/cell_census/setup.cfg
bkmartinjr: summary was incorrect - now fixed

PR #52, opened 2022-12-20
simplify get_anndata API

### LINKED ISSUE #36: get_anndata is slow for large gene lists
Due to an issue with TileDB query conditions, var/obs queries with very large number of values used in an `in` expression will be very slow (time is roughly linear to the number of items in the list).

For example, if this query has a very large list of `lung_genes`, it will be very slow.

```
lung_genes = list(lung_var["feature_id"])
lung_adata = cell_census.get_anndata(
    census,
    organism="Homo sapiens",
    obs_query={"tissue_general": "lung", "is_primary_data": True},
    var_query={"feature_id": lung_genes},
)
```

This expands to a `value_filter` on the `var` DataFrame that looks like `feature_id in ["gene1", "gene2", ..., "geneN"]`, which currently has performance roughly O(N), where N is the number of possible matches (right side of the `in` operator).

a MUCH faster alternative is to directly use the `soma_joinids` (coordinates) and skip the table scan.

The filter issue has been reported to TileDB. Still considering an appropriate work-around for the `get_anndata` API.  It may be helpful to expose the coords that the underlying experiment query supports.

---

~~Update:  ETA from TileDB for a fix to the underlying query condition is late Q1.  Tracking id 24310~~

~~Update [2023-07-11] this work did not make the TileDB embedded 2.16 release train, and is now slated for 2.17 (ETA Q3 2023)~~

Update [2023-09-29] this is now slated for early Q4 2023 in a 1.5.X release.

Update [2023-10-12] this has been fixed via single-cell-data/TileDB-SOMA#1756  Should be in the next release.

---

performance test case:

```
"""
Test case for cellxgene-census#36
"""

import sys
import time

import cellxgene_census
import tiledbsoma


def main():
    print(tiledbsoma.show_package_versions())
    organism = "homo_sapiens"
        exp = census["census_data"][organism]
        var_df = exp.ms["RNA"].var.read().concat().to_pandas()
        test_cases = [
            slice(0, 10),
            slice(0, 100),
            slice(0, 1000, 10),
            slice(0, 250),
            slice(0, 500),
            slice(0, 1000),
            slice(0, 2000),
            slice(0, 4000),
        ]

        print("test,                   coords,    filter")
        for test_slc in test_cases:
            genes = var_df[test_slc]

            # Coordinates test
            var_coords = genes.soma_joinid.to_numpy()
            t_start = time.perf_counter()
            exp.ms['RNA'].var.read(coords=(var_coords,)).concat()
            t_coords = time.perf_counter() - t_start

            # Value filter test
            var_value_filter = f"feature_id in {genes.feature_id.to_list()}"
            t_start = time.perf_counter()
            exp.ms['RNA'].var.read(value_filter=var_value_filter).concat()
            t_value_filter = time.perf_counter() - t_start

            print(f"{str(test_slc):>20}, {t_coords:8.2f}, {t_value_filter:8.2f}")


if __name__ == "__main__":
    sys.exit(main())
```

---

Results of above test case as of 20230927, on main branch and previous release:

```
tiledbsoma.__version__        1.4.3
TileDB-Py tiledb.version()    (0, 22, 3)
TileDB core version           2.16.3
libtiledbsoma version()       libtiledb=2.16.2
python version                3.10.12.final.0
OS version                    Linux 6.2.0-1012-aws
None
The "latest" release is currently 2023-09-25. Specify 'census_version="2023-09-25"' in future calls to open_soma() to ensure data consistency.
test,                   coords,    filter
  slice(0, 10, None),     0.29,     0.33
 slice(0, 100, None),     0.30,     0.70
  slice(0, 1000, 10),     0.25,     0.62
 slice(0, 250, None),     0.36,     1.01
 slice(0, 500, None),     0.24,     1.71
slice(0, 1000, None),     0.24,     3.25
slice(0, 2000, None),     0.29,     6.28
slice(0, 4000, None),     0.29,    12.81


tiledbsoma.__version__        1.5.0rc0.post9.dev3876950273
TileDB-Py tiledb.version()    (0, 23, 0)
TileDB core version           2.17.0
libtiledbsoma version()       libtiledb=2.17.0
python version                3.10.12.final.0
OS version                    Linux 6.2.0-1012-aws
None
The "latest" release is currently 2023-09-25. Specify 'census_version="2023-09-25"' in future calls to open_soma() to ensure data consistency.
test,                   coords,    filter
  slice(0, 10, None),     0.28,     0.44
 slice(0, 100, None),     0.29,     0.68
  slice(0, 1000, 10),     0.24,     0.67
 slice(0, 250, None),     0.34,     1.09
 slice(0, 500, None),     0.26,     2.42
slice(0, 1000, None),     0.28,     3.23
slice(0, 2000, None),     0.25,     6.32
slice(0, 4000, None),     0.31,    13.51
```

bkmartinjr:
PR simplifies the get_anndata API: value_filter naming and syntax normalized/aligned with SOMA. The get_anndata API now uses the same "value_filter" syntax as SOMA.

Example:
```python
    cell_census.get_anndata(
        census, 
        "mus musculus",
        obs_value_filter="is_primary_data == True and tissue_general == 'lung'",
        var_value_filter="feature_name in ['Gm16259', 'Dcaf5', 'Gm53058']",
    })
```

Earlier changes which added more complexit to get_anndata (e.g., joinids-based access) are removed.  The assumption is that this layer is a simple wrapper on ExperimentQuery, for simple/starter use cases. If the user needs more complexity, they should drop down to a lower layer API, where concepts like "joinid" and the full gamut of SOMA API are available.

CC: @pablo-gar - in case you have comments on the API (param) name changes.


atolopko-czi:
It seems that we're pushing the responsibility of obtaining the `joinid`s onto the API user that wants to retrieve an AnnData object, given some set of axes filtering criteria. But wo we expect the user can obtain the `joinid`s in a way that is more efficient than the (slow) query TileDB being used internally by the API (`cell_census.get_anndata._build_query()`)?  If the user can't reasonably get joinids in a more efficient way, is this worth fixing at the CC API level?

If TileDB provides an optimization, are we going to back this out in the future?

bkmartinjr:
> It seems that we're pushing the responsibility of obtaining the joinids onto the API user that wants to retrieve an AnnData object

Even without the short-term performance use case, there are motivations for this change.  And TileDB is working on a fix for the value filter - I don't have a schedule for the fix, but it is coming.

Even after that fix arrives, this change provides the _opportunity_ for the user to get the `joinids` on their own - it is optional, and they can still use the value filter method if they prefer it.  There are several use cases motivating this optionality:
1. The user may wish to do multiple calls to `get_anndata` using those the same obs/var value filter.  This allows them to re-use the joinids from the first result, in subsequent calls to get_anndata.  This will be significantly faster than re-scanning obs/var on each call.
2. The user may have access to the joinid from some other source, and wish to directly address cells/genes by this id.  A concrete example: they have an existing slice of the X matrix, and wish to us it to look up the obs/var dataframe metadata.

You are correct that this change is not an acceptable long-term fix for the slow TileDB value filter performance.  It is motivated by other long-term use cases, *and* it provides a short-term work-around for users who need to slice by a large number of gene feature_ids.

api/python/cell_census/src/cell_census/experiment_query/axis.py
```diff
@@ -6,10 +6,9 @@
 import pyarrow as pa
 from typing_extensions import TypedDict
 
-# Type declaration/helpers local to this file
-#
-Coordinates = Tuple[Union[slice, int, npt.ArrayLike], ...]
-ValueFilter = str
+AxisCoordinate = Union[slice, int, npt.ArrayLike]
+AxisCoordinates = Tuple[AxisCoordinate, ...]
+AxisValueFilter = str
```
atolopko-czi, line 11: When would a user want to grab a _slice_ of `joinid`s? Does the order of `joinid`s ever have significance? Sure someone might want to grab `[0:10]` for  playing around with small subset of an axis, but if that's not a realistic use, why support it at this API level? `list(range(10))` is an easy enough workaround, if needed.

(Even if we change joinids from being ints to strs for `var`, the ordering doesn't seem to have significance.)

If we do away with slices, we can simplify to:
```
AxisCoordinate = Union[int, str, bytes]
AxisCoordinates = List[AxisCoordinate]
```
And now these type names are more accurate.  The singular form `AxisCoordinate` is now _always_ a scalar value, and `AxisCoordinates` can never be a list of lists, which is a case that makes the current name also misleading.

api/python/cell_census/src/cell_census/get_anndata.py
```diff
@@ -104,14 +122,16 @@ def get_anndata(
 
     Examples
     --------
-    >>> get_anndata(census, "Mus musculus", obs_query={"tissue": "brain"})
+    >>> get_anndata(census, "Mus musculus", obs_value_filter={"tissue": "brain"})
 
     >>> get_anndata(census, "Homo sapiens", column_names={"obs": ["tissue"]})
 
+    >>> get_anndata(cesus, "Homo sapiens", obs_joinds=[0,88,222])
```
atolopko-czi, line 129:
```suggestion
    >>> get_anndata(cesus, "Homo sapiens", obs_joinids=[0,88,222])
```

atolopko-czi:
> this change provides the opportunity for the user to get the joinids on their own

My sense is that to maintain ease-of-use, the Cell Census API should hide `joinid`s as much as possible. This changes mixes two levels of abstraction into one place: filtering by domain-relevant values and filtering by "implementation" values. 

While ExperimentQuery  also mixes these two filtering modes, the `get_anndata()` is intended to be the easiest entrypoint to the API users.  Users who call this method are likely to want get straight to the AnnData world, and I doubt they want to think about `joinid`s.  If they are interested in doing joinid-related filtering, they can still use experiment_query.

> The user may wish to do multiple calls to get_anndata using those the same obs/var value filter

If both obs/var are unchanged, they can just reuse an ExperimentQuery object, and call `read_as_anndata()` repeatedly, right? 

If _one of_ obs or var needs to change, we could probably support a mutable ExperimentQuery that allow you to refilter on one axis, thus preserving the result of the other axis query (the joinids ca be maintained internally).

atolopko-czi (changes requested):
One discussion point, one typo

api/python/cell_census/src/cell_census/experiment_query/axis.py
bkmartinjr: I think you are right, at least as the census currently uses the joinids (ie. obs is _only_ indexed on the joinid - which is a census schema assumption, not a SOMA assumption).  I didn't want the API to make too many assumptions which would restrict our ability to evolve the census schema, which is why I exposed the full set of underlying capabilities.

I think we can do either way:
* minimal API surface area that would have to change if we evolve the schema
* or, more flexible API surface area that would be less likely to change

probably simplest to discuss IRT

api/python/cell_census/src/cell_census/get_anndata.py
```diff
@@ -1,88 +1,27 @@
...
 def get_anndata(
     census: soma.Collection,
     organism: str,
     measurement_name: str = "RNA",
     X_name: str = "raw",
-    obs_joinids: Optional[AxisCoordinate] = None,
-    obs_value_filter: Optional[ObsQuery] = None,
-    var_joinids: Optional[AxisCoordinate] = None,
-    var_value_filter: Optional[VarQuery] = None,
+    value_filter: Optional[AxisValueFilters] = None,
```
atolopko-czi, line 24: Any reason not to keep obs and var filters as separate params? It's how ExperimentQuery takes them, and would eliminate the intermediate `AxisValueFilters` type.  We could also have `AxisValueFilter` (singular) as an alias for `str`, if that has any semantic value.

We could then also use `AxisQuery()` (the "null" filter) for obs and var params, cleaning up some of the `... if ... else ...` lines below.

api/python/cell_census/src/cell_census/get_anndata.py
```diff
@@ -122,20 +50,18 @@ def get_anndata(
...
     >>> get_anndata(census, "Homo sapiens", column_names={"obs": ["tissue"]})
 
-    >>> get_anndata(cesus, "Homo sapiens", obs_joinds=[0,88,222])
-
     """
     exp = get_experiment(census, organism)
-    _obs_query = _build_query(obs_joinids, obs_value_filter)
-    _var_query = _build_query(var_joinids, var_value_filter)
+    _obs_value_filter = None if value_filter is None else value_filter.get("obs", None)
+    _var_value_filter = None if value_filter is None else value_filter.get("var", None)
```
atolopko-czi, line 60: Nit (but only needed if we keep the `AxisValueFilters` dict param)
```suggestion
    _obs_value_filter = value_filter.get("obs", None) if value_filter is not None else None
    _var_value_filter = value_filter.get("var", None) if value_filter is not None else None
```

api/python/notebooks/census_query_extract.ipynb
```diff
@@ -13,7 +13,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 8,
```
atolopko-czi, line 16: I'll have to write a pre-commit hook that sets these all to 0. 😆

atolopko-czi (approved):
A couple minor code-level suggestions, but overall simplification LGTM.

atolopko-czi:
Related thought: should we rename `get_anndata.py` to `anndata.py` for consistency with `experiment.py`, which also has only a single `get` method? Or does `anndata` module name cause namespacing problems with `anndata` package?

api/python/cell_census/src/cell_census/get_anndata.py
bkmartinjr: We had them in a dict for `column_names`, so I replicated that pattern.  I don't mind breaking them out into multiple parameters, but we need to do the same for both `value_filter` and `column_names`.

I am personally agnostic so I will change (both) as you suggest.

bkmartinjr:
> Related thought: should we rename `get_anndata.py` to `anndata.py` for consistency with `experiment.py`, which also has only a single `get` method? Or does `anndata` module name cause namespacing problems with `anndata` package?

Yes, all heck breaks loose if we do this (IIRC, with the static type checking)

pablo-gar (approved):
LGTM, I like the consistency with SOMA.

PR #53, opened 2022-12-21
docs: add integration notebook

pablo-gar:
Adds the notebook `comp_bio_data_integration_scvi.ipynb` showcasing the metadata in the cell census that can be used for integration


atolopko-czi:
To play devil's advocate: Considering this is mostly a tutorial for scvi after the AnnData object is obtained, I'm wondering how much value does this notebook have in terms of highlighting Cell Census-specific capabilities? Could it get away with just showing how to get the data needed to address various batch effects, e.g., and otherwise point the user to an scvi tutorial? 

I can see also see the benefit to having the full code in one place that can be easily run and modified by the user. So I'm just raising the question of how the notebooks are used and their overall purpose. 

Maybe we even split notebooks into two directories: cell census API "mechanics" and single-cell analysis "recipes".  In case that helps users judge which notebooks they want to dive into.

atolopko-czi:
`scvi` spits out a bunch of warnings; wondering if we should point out that these are produced by scvi and are ok, just to calm the anxieties of the reader. :)

bkmartinjr:
> `scvi` spits out a bunch of warnings; wondering if we should point out that these are produced by scvi and are ok, just to calm the anxieties of the reader. :)

We also get warnings from ScanPy (et al) on other notebooks. Seems like nature of the beast....

brianraymor:
Is it possible to suppress the warnings? (And would you want to?)

bkmartinjr (approved):
I don't have any strong opinions related to Andrew's questions.  Otherwise, it looks fine to me.

I do think we need to make some "product" decisions around how far / expansive demo notebooks need to be, but putting a bunch of them out there, as documentation, seems generally helpful.

bkmartinjr:
@pablo-gar - I re-triggered the workflows to see if the spurious errors clear.

bkmartinjr:
@pablo-gar - PR #52 will require a slight tweak to this notebook.

pablo-gar:
> Considering this is mostly a tutorial for scvi after the AnnData object is obtained, I'm wondering how much value does this notebook have in terms of highlighting Cell Census-specific capabilities? 

@atolopko-czi did it read like a tutorial for scvi? 

If so I'd like to modify to achieve the intended goal. My intention was to showcase the metadata that could be used for batch information and to demonstrate the implications that it may have in the results, in this case using `dataset_id` + `donor_id` produce the best results for batch selection (parametrization of scvi still needs to happen to optimize these results, with formal clustering metrics)

I intentionally left out all details related to optimizing scvi for integration and pointed to their docs. 

> Maybe we even split notebooks into two directories: cell census API "mechanics" and single-cell analysis "recipes".

I'm a little hesitant to provide recipes at this point, as that would require us to invest more time as to what the **best** way to perform operations is (in this case integration). In the future (maybe next few months) this is likely a direction we want to take.

---- 

Maybe a reduce the amount of text dedicated to scvi to not make it seem like an scvi tutorial?

pablo-gar:
> Maybe we even split notebooks into two directories: cell census API "mechanics" and single-cell analysis "recipes". In case that helps users judge which notebooks they want to dive into. Actually, I see that we use different prefixes for notebooks, cell_census_ and comp_bio_, which is basically the same thing. Sub-directories might still be useful. If nothing else, maybe explain the prefixes in the README [here](https://github.com/chanzuckerberg/cell-census/blob/686f0ab9634cd97263e111fe8212d4cefb3be6fa/api/python/notebooks/README.md#run-notebooks).

Yes this was I suggestion I gave to Bruce and I'm on the hook to separate the notebooks in two different folders.

bkmartinjr:
Note the naming scheme for notebooks:
* `census_` is mechanics demos
* `comp_bio_` is scientific workflow demos

atolopko-czi:
> > Considering this is mostly a tutorial for scvi after the AnnData object is obtained, I'm wondering how much value does this notebook have in terms of highlighting Cell Census-specific capabilities?
> 
> @atolopko-czi did it read like a tutorial for scvi?

"Tutorial" was poor choice of words. I was just considering the amount of content that was anndata/scvi-related versus cell census-related. I don't want to hold this up unnecessarily; it clearly has value, as is. 

> > Maybe we even split notebooks into two directories: cell census API "mechanics" and single-cell analysis "recipes".
> 
> I'm a little hesitant to provide "formal" recipes at this point, as that would require us to invest more time as to what the **best** way to perform operations is (in this case integration). In the future (maybe next few months) this is likely a direction we want to take.

Ah, yes, if a "recipe" implies a more refined or "accepted" process , let's not introduce that term.

I think the prefixes work well enough.

atolopko-czi (approved):
LGTM.  There's one missing space after a period in the intro. 😬 Can't comment directly due to Github diff size limitations

PR #59, opened 2023-01-05
feat: add validations to census builder

### LINKED ISSUE #35: cell_census_builder: validate no dups
Add validations that there are no duplicate indices in all Measurement Layers, for all Experiments:
* X[raw] - all (soma_dim_0, soma_dim_1) pairs are unique
* obs/var - all soma_joinid are unique
Then configure `allow_duplicates=True` (to improve TileDB performance). This requires parameterizing the `allow_duplicates` attribute via `platform_config` in TileDB-SOMA library.

The removal of `allow_duplicates` on the underlying array schemas provides a significant read performance improvement.

atolopko-czi:
Resolves #35 

- no dups in X matrix (unique coords)
- no dups in obs or var soma_joins

Testing in progress, but early review may be useful.

### DIFF
--- a/tools/cell_census_builder/__main__.py
+++ b/tools/cell_census_builder/__main__.py
@@ -120,8 +120,13 @@ def build(
     top_level_collection, filtered_datasets = build_step2_create_axis(
         soma_path, assets_path, datasets, experiment_builders, args
     )
-    assign_soma_joinids(filtered_datasets)
+
+    if not filtered_datasets:
+        logging.error("Zero datasets are suitable for processing. Aborting.")
+        return 1
+
     logging.info(f"({len(filtered_datasets)} of {len(datasets)}) suitable for processing.")
+    assign_soma_joinids(filtered_datasets)
     gc.collect()
 
     # Step 3- create X layers
--- a/tools/cell_census_builder/validate.py
+++ b/tools/cell_census_builder/validate.py
@@ -318,7 +318,7 @@ def validate_X_layers(
                 assert X.schema.field("soma_data").type == CENSUS_X_LAYERS[lyr]
                 assert X.shape == (n_obs, n_vars)
 
-                assert _validate_X_has_no_dups(X, n_obs), f"duplicate coordinates in {eb.name} layer {lyr} matrix"
+                assert _validate_X_has_no_dup_coords(X, n_obs), f"duplicate coordinates in {eb.name} layer {lyr} matrix"
 
     if args.multi_process:
         with create_process_pool_executor(args) as ppe:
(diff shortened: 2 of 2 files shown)


tools/cell_census_builder/__main__.py
```diff
@@ -120,8 +120,13 @@ def build(
     top_level_collection, filtered_datasets = build_step2_create_axis(
         soma_path, assets_path, datasets, experiment_builders, args
     )
-    assign_soma_joinids(filtered_datasets)
+
+    if not filtered_datasets:
+        logging.error("Zero datasets are suitable for processing. Aborting.")
+        return 1
```
atolopko-czi, line 126: Fail here instead of later with an arcane, harder-to-troubleshoot error in code that doesn't expect empty set of datasets. Useful when using `--test-first-n`, which may result in 0 admissible datasets.

tools/cell_census_builder/validate.py
```diff
@@ -311,6 +318,8 @@ def validate_X_layers(
                 assert X.schema.field("soma_data").type == CENSUS_X_LAYERS[lyr]
                 assert X.shape == (n_obs, n_vars)
 
+                assert _validate_X_has_no_dup_coords(X, n_obs), f"duplicate coordinates in {eb.name} layer {lyr} matrix"
```
atolopko-czi, line 321: As it turns out, `_validate_X_layers_contents` catches the dup coords as well (e.g. `AssertionError: homo_sapiens:4 'raw' nnz mismatch 3618682 vs 1809341`, on a test corpus that duplicates every obs row). So this check might not be strictly necessary, though it does make the semantics of the check explicit.

bkmartinjr (commented):
The motivation for this change was to enable use of the TileDB schema `allow_duplicates` feature (for the X sparse matrices).  This schema config (allowing dups at the core DB level) unlocks additional read performance, as the core DB does not have to perform duplicate consolidation upon read.

Is the plan to do that in another PR?  Without this TileDB schema change, we essentially are doing the check twice (once in the core DB because `allow_updates=False`, and once in our own code).

atolopko-czi:
> The motivation for this change was to enable use of the TileDB schema `allow_duplicates` feature (for the X sparse matrices). This schema config (allowing dups at the core DB level) unlocks additional read performance, as the core DB does not have to perform duplicate consolidation upon read.
> 
> Is the plan to do that in another PR? Without this TileDB schema change, we essentially are doing the check twice (once in the core DB because `allow_updates=False`, and once in our own code).

I assumed (incorrectly) that the `allow_dups` setting change was separately planned work. Will take that on as part of the story.

bkmartinjr:
> I assumed (incorrectly) that the allow_dups setting change was separately planned work.

it is fine either way with me.  I'll approve the PR, and you can either land, or hit me up for a second review!

bkmartinjr (approved):
approved assuming enabling allow_dups in a second PR.  If not, I'm happy to re-review!  TY.

atolopko-czi:
Parent issue replaced by https://github.com/chanzuckerberg/cell-census/issues/61. Closing this PR.

We _could_ salvage two minor pieces from this PR:
* https://github.com/chanzuckerberg/cell-census/pull/59/files#diff-84c0d083cd4389aad3dcca3747de5b150f9e9f157074304057471f398df803c7R221-R231
* https://github.com/chanzuckerberg/cell-census/pull/59/files#diff-5eb2ff26a895b5d0440289ad915a57f054370888653a8ee168c34882ca2d4c7fR124-R126

Issue #67, opened 2023-01-11
update Census API and Builder to tilebsoma 0.5.0a4


atolopko-czi:
Make updates to cell_census_builder for TileDB-SOMA changes introduce by:
* https://github.com/single-cell-data/TileDB-SOMA/issues/570 (query API)
* https://github.com/single-cell-data/TileDB-SOMA/issues/644 (explicit context)

- [x] @atolopko-czi complete #77
- [x] @atolopko-czi  will re-run (and fix, as needed) the `comp_bio_explore_and_load_lung_data.ipynb` notebook. This can be done as a follow-up PR to #77

atolopko-czi:
@bkmartinjr Any objection if I also make API changes as needed while I'm fix builder breakages from TileDB-SOMA updates?

bkmartinjr:
None - it would be appreciated.

PR #77, opened 2023-01-20
update to tilebsoma 0.5.0a4

atolopko-czi:
For #67 

Update api, notebooks, and builder to `tiledbsoma==0.5.0a4`

- update api, notebooks, and builder for:
    - fluent query API
    - SOMA context in TileDbObject constructors
- add unit test for highly_variable_genes
- updates include a minor code regression in how we obtain indexer from ExperimentAxisQuery; will be reverted after https://github.com/single-cell-data/TileDB-SOMA/issues/768 in released


api/python/cell_census/src/cell_census/get_anndata.py
```diff
@@ -52,11 +54,7 @@ def get_anndata(
...
     with exp.axis_query(
         measurement_name,
-        obs_query=soma.experiment_query.AxisQuery(value_filter=obs_value_filter)
-        if obs_value_filter is not None
-        else None,
-        var_query=soma.experiment_query.AxisQuery(value_filter=var_value_filter)
-        if var_value_filter is not None
-        else None,
+        obs_query=AxisQuery(value_filter=obs_value_filter) if obs_value_filter is not None else AxisQuery(),
+        var_query=AxisQuery(value_filter=var_value_filter) if var_value_filter is not None else AxisQuery(),
```
atolopko-czi, line 58: cannot pass an explicit `None` to tiledbsoma, so passing `AxisQuery()` instead

pablo-gar (commented):
LGTM!

pablo-gar (approved):

api/python/cell_census/src/cell_census/get_anndata.py
pablo-gar: I understand now your question at slack @atolopko-czi https://czi-sci.slack.com/archives/C02SP5AV7PV/p1674229691525729

I'd like us to minimize the divergence between SOMA API and the convenience functions we have here. So I'd like to define the parameters here similarly as they are defined in SOMA. So either we change SOMA to handle `None` or we change the default of get_anndata here to

```
def get_anndata(
    census: soma.Collection,
    organism: str,
    X_name: str = "raw",
    obs_value_filter: Optional[str] =  AxisQuery(), #### <-----
    var_value_filter: Optional[str] =  AxisQuery(), #### <-----
    column_names: Optional[AxisColumnNames] = None,
) -> anndata.AnnData:
```


To me it seems more reasonable to change the parameters here.

pablo-gar (changes requested):
I had a small request up there

api/python/cell_census/src/cell_census/get_anndata.py
atolopko-czi: Agreed that we should minimize divergence. So for now I'll update here. But I'll aslo remove the `Optional` as shown in your example, since it implies that `None` is ok, when it isn't.  This is how SOMA has it as well (not optional).

atolopko-czi: actually, since the param is the the _string_ filter that is passed _to_ the `AxisQuery()`, I think what we have now makes sense.

atolopko-czi: It might be worth breaking this off as a separate issue, so that we can close out this PR.

api/python/cell_census/pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies= [
     "numpy>=1.21",
     "requests",
     "tiledb>=0.19.0",
-    "tiledbsoma==0.5.0a3",
+    "tiledbsoma==0.5.0a4",
```
bkmartinjr, line 34: at this point, 0.5.0a5 has been released.  update or separate PR?  I'm fine with either.

api/python/cell_census/src/cell_census/compute/highly_variable_genes.py
```diff
@@ -1,25 +1,23 @@
 import numpy as np
 import pandas as pd
-import tiledbsoma as soma
+from somacore.query.query import ExperimentAxisQuery
```
bkmartinjr, line 3: The name is exported at the top level, so this should be:
> from tiledbsoma import ExperimentAxisQuery

Also, Google Python conventions are to never import a naked name like that, but instead import the module.  So a better style would be:

> import tiledbsoma as soma

and then use it as:

> soma.ExperimentAxisQuery

api/python/cell_census/src/cell_census/compute/highly_variable_genes.py
```diff
@@ -1,25 +1,23 @@
...
 
 from .meanvar import OnlineMatrixMeanVariance
 
 
-def highly_variable_genes(query: soma.experiment_query.ExperimentAxisQuery, n_top_genes: int = 10) -> pd.DataFrame:
+def highly_variable_genes(query: ExperimentAxisQuery, n_top_genes: int = 10) -> pd.DataFrame:
```
bkmartinjr, line 8: see above style comment about using unqualified names from imported modules

api/python/cell_census/src/cell_census/open.py
```diff
@@ -5,12 +5,11 @@
 import s3fs
 import tiledb
 import tiledbsoma as soma
+from tiledbsoma.options import SOMATileDBContext
```
bkmartinjr, line 8: See above comment about the unqualified import of names.  E.g.,
> https://google.github.io/styleguide/pyguide.html#22-imports

bkmartinjr (approved):
A handful of non-blocking style issues, and a question about whether you want to update to the latest tiledbsoma release.  But otherwise, it looks great.

api/python/cell_census/src/cell_census/get_anndata.py
bkmartinjr: I agree this should be pushed to another issue/PR.  We need airtime on the latest SOMA API changes.

atolopko-czi: https://github.com/chanzuckerberg/cell-census/issues/80

api/python/cell_census/pyproject.toml
atolopko-czi: Intended to address a5 in a new issue/PR.  With the now quicker tiledbsoma alpha release cadence, I figured this PR might be chasing the latest version indefinitely.

atolopko-czi: https://github.com/chanzuckerberg/cell-census/issues/81

api/python/cell_census/src/cell_census/compute/highly_variable_genes.py
atolopko-czi: Made a set of fixes in c89938e. A couple more fixes will be made after required tiledbsoma updates (https://github.com/single-cell-data/TileDB-SOMA/issues/791).

(I've been waffling on the correct way to import, so thank you for the style advice!)

api/python/cell_census/src/cell_census/compute/highly_variable_genes.py
atolopko-czi: done

api/python/cell_census/src/cell_census/open.py
atolopko-czi: I think we need to export the `tiledbsoma.options` sub-pkg exports as a top-level `tiledbsoma` export?


Usage such as this:
```
import tiledbsoma as soma
...
soma.SOMATileDBContext()
```
seems better than just having an "unaffiliated" `options` package alias:
```
from tiledbsoma import options

options.SOMATileDBContext()
```.

Though this could work:
```
from tiledbsoma import options as soma_options

soma_options.SOMATileDBContext()
```
Noted in https://github.com/single-cell-data/TileDB-SOMA/issues/791

Comment (bkmartinjr) [reply api/python/cell_census/src/cell_census/open.py]: > I think we need to export the tiledbsoma.options sub-pkg exports as a top-level tiledbsoma export?

No, it should just work as is.  I.e.,
```
In [5]: import tiledbsoma as soma

In [6]: soma.options.SOMATileDBContext
Out[6]: tiledbsoma.options.soma_tiledb_context.SOMATileDBContext
```

atolopko-czi: Indeed, and PyCharm has misled me--it complains about `soma.options.SOMATileDBContext`. 🤔

Issue #85, opened 2023-01-24
Reduce resources needed for running comp_bio* notebooks


pablo-gar:
The following notebooks require either too much memory and/or time to run. They should be modified to run in a machine with about 32GB of memory and should take less than 15 min to run each. (these numbers serve for guidance)

@atolopko-czi:
- [x] Benchmark runtime of all notebooks and revise below list, as needed. 

@pablo-gar Revise notebooks as needed:
- [ ] [comp_bio_data_integration_scvi.ipynb](https://github.com/chanzuckerberg/cell-census/blob/main/api/python/notebooks/analysis_demo/comp_bio_data_integration_scvi.ipynb)
- [x] [comp_bio_explore_and_load_lung_data.ipynb](https://github.com/chanzuckerberg/cell-census/blob/main/api/python/notebooks/analysis_demo/comp_bio_explore_and_load_lung_data.ipynb)
- [x] [census_compute_over_X.ipynb](https://github.com/chanzuckerberg/cell-census/blob/main/api/python/notebooks/api_demo/census_compute_over_X.ipynb)
- [x] comp_bio_normalizing_full_gene_sequencing.ipynb

atolopko-czi:
[notebooks-runtime-summary.txt](https://github.com/chanzuckerberg/cell-census/files/10494106/notebooks-runtime-summary.txt)

See `elapsed` (in MM:SS.SSS) and `maxresident` (in kb) to see which notebooks are candidates for improvements.

* `analysis_demo/comp_bio_explore_and_load_lung_data.ipynb` failed outright on a 64GB instance
* `analysis_demo/comp_bio_normalizing_full_gene_sequencing.ipynb` failed due to an apparent data-related issue (error in attached file)


Command to run & time all notebooks:
```
ls api_demo/census_high_variable_genes.ipynb | xargs -n 1 time jupyter-nbconvert --to notebook --execute 2>&1 | tee -a notebook-timings.txt
```

pablo-gar:
Addressed in #101 #98 #92 #91

pablo-gar:
Re opened because of #112

brianraymor:
Closed as complete per triage.

Issue #80, opened 2023-01-23
change get_anndata() axis params  to tiledbsoma.AxisQuery 


atolopko-czi:
[Discussion](https://github.com/chanzuckerberg/cell-census/pull/77#discussion_r1083036031)

`cell_census.get_anndata.get_anndata()` currently takes the following params:
```
    obs_value_filter: Optional[str] = None,
    obs_coords: Tuple[SparseDFCoord, ...] = (slice(None),),
    var_value_filter: Optional[str] = None,
    var_coords: Tuple[SparseDFCoord, ...] = (slice(None),),
```

These params simply "flatten" the `AxisQuery()` constructor params, so we could just as well accept (just 2) `AxisQuery` params instead. This would make this CC API method's signature more consistent with the SOMA method signatures.

pablo-gar:
@atolopko-czi I think this is ok to close because of #102 and #91. Thoughts?

atolopko-czi:
I don't feel too strongly about this change, so I'm okay if you close it. 

Since one intent of CC API is to reduce the complexity of using SOMA (agreed?), keeping AxisQuery() out of the signature helps accomplish that.  A minor gain, but still a gain for the casual user that just wants an AnnData object. One could argue that if we changed this method's signature to take AxisQuery(), we might as well do away with the method entirely.

bkmartinjr:
Has there been a resolution on this?  I can implement if you update the issue to reflect the intended direction.

atolopko-czi:
@pablo-gar? I vote for do nothing.

pablo-gar:
We can leave unchanged. There were recent PRs that addressed it tangentially as well. 

#117 Docstrings updated
#91 `obs_coords` and `var_coords` were added as arguments

https://github.com/chanzuckerberg/cell-census/pull/102#discussion_r1088944589

PR #91, opened 2023-01-25
Improve notebook `comp_bio_explore_and_load_lung_data.ipynb`

pablo-gar:
Part #85 

Improves performance of `comp_bio_explore_and_load_lung_data.ipynb` by only loading 100K randomly selected cells instead of >1mi from lung into an AnnData.

To achieve this the `obs_coors` and `var_coords` arguments where added to `get_anndata()`. Partially related to #80. 

The notebook now can be run a 32GB machine and it runs in <5 min


api/python/cell_census/src/cell_census/get_anndata.py
```diff
@@ -54,7 +57,7 @@ def get_anndata(
     exp = get_experiment(census, organism)
     with exp.axis_query(
         measurement_name,
-        obs_query=soma.AxisQuery(value_filter=obs_value_filter) if obs_value_filter is not None else soma.AxisQuery(),
-        var_query=soma.AxisQuery(value_filter=var_value_filter) if var_value_filter is not None else soma.AxisQuery(),
+        obs_query=soma.AxisQuery(value_filter=obs_value_filter, coords = obs_coords),
+        var_query=soma.AxisQuery(value_filter=var_value_filter, coords = var_coords),
```
pablo-gar, line 61: @atolopko-czi unless I missed something the `if` wasn't needed because the defaults `value_filter=None` and `coords=(slice(None),)` are properly handled by `AxisQuery`

pablo-gar, line 61: @atolopko-czi unless I missed something the `if` wasn't needed because the defaults `value_filter=None` and `coords=(slice(None),)` are properly handled by `AxisQuery`

atolopko-czi: That looks right! 

(While it was true that you couldn't pass in `obs_query=None`, due to a [bug that's being fixed](https://github.com/single-cell-data/SOMA/pull/99), it appears we didn't need that conditional at all. Nice catch!)

atolopko-czi:
Change
```
There the total number of unique lung cells and their composition for:
```
to:
```
The total number of unique lung cells and their composition for:
```

---

```
+  code cell:
+    execution_count: 18
+    source:
+      lung_cell_subsampled_n = 100000
+      lung_cell_subsampled_ids = (
+          lung_obs
+          .query("is_primary_data==True")["soma_joinid"].
+          sample(lung_cell_subsampled_n, random_state=1).
+          tolist()
+      )
```
Is the `is_primary_data==True` necessary? Wasn't `lung_obs` already filtered for that?

PR #92, opened 2023-01-25
Improve notebook `comp_bio_data_integration_scvi.ipynb`

pablo-gar:
Improves performance of `comp_bio_data_integration_scvi.ipynb` by only using 30 hidden layers and 100 epochs for the scvi model.

The notebook now can be run a 32GB machine and it runs in <10 min


atolopko-czi (approved):
LGTM!

PR #101, opened 2023-01-26
Improve  notebook `census_compute_over_X.ipynb`

pablo-gar:
Related to #85 

This notebook is now able to run in a 32 GB machine under 2 mins. The last example in the notebook was narrowed to only explore 5 genes instead of all >60K genes.


atolopko-czi (approved):
LGTM

Issue #78, opened 2023-01-21
Assess the pros and cons of a convenience Cell Census API


pablo-gar:
It has been brought to our attention that the existence of a convenience Cell Census API creates positive and negative impacts on some of the Cell Census requirements as a whole. These requirements are:

<table class="tg">
<thead>
  <tr>
    <th class="tg-0lax">Requirement</th>
    <th class="tg-0lax">Impact of <br>Convenience API </th>
    <th class="tg-0lax">Why</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td class="tg-0lax"><span style="font-weight:400;font-style:normal">The Cell Census must be usable for single-cell research by computational biologists who specialize in </span><br><span style="font-weight:400;font-style:normal">single-cell and are either well versed in R scripting or python scripting</span></td>
    <td class="tg-0lax">Positive</td>
    <td class="tg-0lax"><span style="font-weight:400;font-style:normal">The data can be accessed and loaded easily.</span></td>
  </tr>
  <tr>
    <td class="tg-0lax">Maintenance of the Cell Census should not add unnecessary complexity to the API stack SOMA/tileDB.</td>
    <td class="tg-0lax">Negative</td>
    <td class="tg-0lax">Divergence may occur; an extra layer to maintain</td>
  </tr>
  <tr>
    <td class="tg-0lax">Communicating the Cell Census value prop should be clear and straightforward</td>
    <td class="tg-0lax">Negative</td>
    <td class="tg-0lax">Explaining and extra layer to the user is challenging </td>
  </tr>
</tbody>
</table>

We need to assess the areas where are willing to compromise based on an assessment of priorities. Then we can decide what the appropriate solution is. Current candidates:

1. Convenience `cell-census` API as it currently exists
2. `cell-census` API with with a different class of support/maturity/etc.  Eg., they might live in a sub-package, such as cell_census.unsupported (feel free to pick a better name). suggested by @bkmartinjr 
3. Remove completely `cell-census` API and document recipes for accessing and processing the Cell Census data, suggested by @brianraymor 

Of important note, regardless of the solution for this. we aim to push any important and generalizable processes to the `tilesbsoma` API.

-----

On a final note the following are the current functionalities supported by the convenience census API

#### Census build access

- `open_soma()` Quickly read the Cell Census SOMACollection. 
   - [Location](https://github.com/chanzuckerberg/cell-census/blob/main/api/python/cell_census/src/cell_census/open.py)
[...]

PR #102, opened 2023-01-26
Complete docstrings of `get_anndata()`

pablo-gar:
Addresses #100

### DIFF
--- a/api/python/cell_census/src/cell_census/get_anndata.py
+++ b/api/python/cell_census/src/cell_census/get_anndata.py
@@ -59,7 +59,7 @@ def get_anndata(
     --------
     >>> get_anndata(census, "Mus musculus", obs_value_filter="tissue_general in ['brain', 'lung']")
     
-    >>> get_anndata(census, "Mus musculus", obs_coors=([1, 10, 100],)
+    >>> get_anndata(census, "Mus musculus", obs_coors=([1, 10, 100],))
 
     >>> get_anndata(census, "Homo sapiens", column_names={"obs": ["tissue"]})
 
--- a/api/python/cell_census/src/cell_census/get_anndata.py
+++ b/api/python/cell_census/src/cell_census/get_anndata.py
@@ -59,7 +59,7 @@ def get_anndata(
     --------
     >>> get_anndata(census, "Mus musculus", obs_value_filter="tissue_general in ['brain', 'lung']")
     
-    >>> get_anndata(census, "Mus musculus", obs_coors=([1, 10, 100],))
+    >>> get_anndata(census, "Mus musculus", obs_coords=([1, 10, 100],))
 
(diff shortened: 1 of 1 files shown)


api/python/cell_census/src/cell_census/get_anndata.py
```diff
@@ -50,6 +58,8 @@ def get_anndata(
     Examples
     --------
     >>> get_anndata(census, "Mus musculus", obs_value_filter="tissue_general in ['brain', 'lung']")
+    
+    >>> get_anndata(census, "Mus musculus", obs_coors=([1, 10, 100],))
```
atolopko-czi, line 62:
```suggestion
    >>> get_anndata(census, "Mus musculus", obs_coords=([1, 10, 100],))
```

api/python/cell_census/src/cell_census/get_anndata.py
```diff
@@ -37,9 +37,17 @@ def get_anndata(
     obs_value_filter: str, default None
         Value filter for the ``obs`` metadata. Value is a filter query written in the
         SOMA ``value_filter`` syntax.
+    obs_coords: Tuple[SparseDFCoord, ...]
+        A set of coordinates on the ``obs`` DataFrame index. The tuple must have a length less 
+        than or equal to the number of dimensions, and be of a type supported by 
+        SOMA ``DataFrame``.
```
atolopko-czi, line 43: This is a larger question than just the docs, but shouldn't `get_anndata()` just take a single `SparseDFCoord` rather than a tuple? Since cell census only has 1 dimension, it seems like any CC API method that takes coords should hide this multi-dimension capability of SOMA.

atolopko-czi (changes requested):
API-level question in the comments.

api/python/cell_census/src/cell_census/get_anndata.py
pablo-gar: I agree with you and I'm trying to thread the line between improved UX for comp bios and consistency with tiledbsoma. This will be a continued conversation we'll have for many functionalities of the census -- API related to #78. I'm leaning towards your suggestion for this case.

Do you have any other extra thoughts on this @atolopko-czi ?

atolopko-czi: I think we just have to take each design choice on a case-by-case basis (i.e. "continued conversation"). If we _always_ try to maintain consistency with tiledbsoma APIs, we might restrict our design choices too much.

pablo-gar:
Completed elsewhere https://github.com/chanzuckerberg/cell-census/pull/117

api/python/cell_census/src/cell_census/get_anndata.py
pablo-gar: Tracking here #119

PR #112, opened 2023-01-31
update (most) notebooks for census 2023-01-30

atolopko-czi:
I went through the effort of rerunning the notebooks for the latest census build, but ran into these problems:
* `comp_bio_data_integration_scvi.ipynb` does not run in reasonable time
* `comp_bio_normalizing_full_gene_sequencing.ipynb` failed due to a fault census build, missing dataset metadata

All other notebooks are updated.  

We can merge these, after review, or just wait until next week to get a clean run on the notebooks.


pablo-gar:
Let's update after review if that's ok with you.

I will re-open related to #85 optimizing the notebooks -- the scvi notebook may benefit from running in a multi-core machine which is what I was using for testing? But I need to investigate.

I opened #113 to track re-running of `comp_bio_normalizing_full_gene_sequencing.ipynb` for when it is possible

pablo-gar (approved):

PR #114, opened 2023-02-03
Add git SHA to the metadata field of census

ebezzi:
Adds the git commit SHA field to the `metadata` object of the top level collection.

Tested by running a local build of the census and inspecting the result

### DIFF
--- a/tools/cell_census_builder/__main__.py
+++ b/tools/cell_census_builder/__main__.py
@@ -139,7 +139,7 @@ def build(
     if args.consolidate:
         consolidate(args, top_level_collection.uri)
 
-    add_git_commit_sha(top_level_collection[CENSUS_INFO_NAME])
+    add_git_commit_sha(top_level_collection)
 
     return 0
 
@@ -264,10 +264,9 @@ def build_step3_create_X_layers(
 
     logging.info("Build step 3 - X layer creation - finished")
 
-def add_git_commit_sha(info_collection: soma.Collection) -> None:
+def add_git_commit_sha(top_level_collection: soma.Collection) -> None:
     sha = get_git_commit_sha()
-    info_collection.set(CENSUS_BUILDER_GIT_SHA, sha, relative=True)
-
+    top_level_collection.metadata["git_commit_sha"] = sha
 
 def create_args_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(prog="cell_census_builder")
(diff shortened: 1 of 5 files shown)


tools/cell_census_builder/__main__.py
```diff
@@ -139,6 +139,8 @@ def build(
     if args.consolidate:
         consolidate(args, top_level_collection.uri)
 
+    add_git_commit_sha(top_level_collection)
```
atolopko-czi, line 142: Let's move this to `create_top_level_collections()`, to keep it with the other top-level metadata setters.  Or move all of those here, if you think that's better (which seems reasonable to me).

tools/cell_census_builder/__main__.py
```diff
@@ -262,6 +264,9 @@ def build_step3_create_X_layers(
 
     logging.info("Build step 3 - X layer creation - finished")
 
+def add_git_commit_sha(top_level_collection: soma.Collection) -> None:
+    sha = get_git_commit_sha()
+    top_level_collection.metadata["git_commit_sha"] = sha
```
atolopko-czi, line 269: let's also validate this is set in `validate.validate_all_soma_objects_exist()`

tools/cell_census_builder/util.py
```diff
@@ -112,3 +114,10 @@ def anndata_ordered_bool_issue_853_workaround(df: pd.DataFrame) -> pd.DataFrame:
             df[k] = df[k].cat.set_categories(df[k].cat.categories, ordered=bool(df[k].cat.ordered))
 
     return df
+
+def get_git_commit_sha() -> str:
+    """
+    Returns the git commit SHA for the current repo
+    """
+    repo = git.Repo(search_parent_directories=True)
+    return repo.head.object.hexsha
```
atolopko-czi, line 123: If we want to be super paranoid (and who doesn't want to be that!?), we could also ensure the git workspace is clean (no uncommitted changes).  Not critical, but might help while we're still building the census manually. Discourages build-time "hot fixes" that aren't committed.

atolopko-czi (changes requested):
Just a few minor requests.

tools/cell_census_builder/util.py
ebezzi: Good idea. Added.

atolopko-czi (approved):
👏 LGTM!

Bento007 (approved):

PR #121, opened 2023-02-03
Add Python 3.7 support

### LINKED ISSUE #120: Add Python 3.7 support
`tiledbsoma` supports it, so it would be useful to expand the supported matrix

bkmartinjr:
With this PR, we now support Python versions 3.7 to 3.10.

Python 3.11 support is pending numba support for the same.

Fixes #120


pablo-gar (approved):
LGTM

PR #122, opened 2023-02-03
Coords param change for get_anndata

### LINKED ISSUE #119: `obs_coords` and `var_coords` from `get_anndata()` should just take a single `SparseDFCoord`
Arguments `obs_coords` and `var_coords` from `get_anndata()` should just take a single `SparseDFCoord` rather than a tuple.
 
Since cell census only has 1 dimension, it seems like any CC API method that takes coords should hide this multi-dimension capability of SOMA.

See discussion here https://github.com/chanzuckerberg/cell-census/pull/102#pullrequestreview-1272660967

bkmartinjr:
Because the census has a single index dimension on both axis, we can further refine the get_anndata parameters for simplicity, by assuming a single dim.

Fixes #119

### DIFF
--- a/api/python/cell_census/src/cell_census/get_anndata.py
+++ b/api/python/cell_census/src/cell_census/get_anndata.py
@@ -1,4 +1,4 @@
-from typing import Optional, Tuple
+from typing import Optional
 
 import anndata
 import tiledbsoma as soma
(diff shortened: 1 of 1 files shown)


pablo-gar (approved):
LGTM

Bento007 (approved):

PR #126, opened 2023-02-06
dataset['title'] is a required field

Bento007:


bkmartinjr (approved):
Optional: If you have time, this might be a good time to add a validator for the `datasets` DataFrame in the `census_info` collection.  Minimally just confirm that for all datasets, the required fields are present.

atolopko-czi:
> Optional: If you have time, this might be a good time to add a validator for the `datasets` DataFrame in the `census_info` collection. Minimally just confirm that for all datasets, the required fields are present.

💯.  Note that this validation could be taken on as part of https://github.com/chanzuckerberg/cell-census/issues/111 which asks that all dataset metadata attributes are validated.

atolopko-czi (approved):
LGTM

Bento007:
Since this is already approve I will add this to a future PR

PR #128, opened 2023-02-07
Census builder unit test

ebezzi:
Base unit test for the cell census builder. Currently:
1. Runs a pre-init step to make sure that the environment is ready for the run
2. Runs the builder with a set of fixtures (2 datasets, 4 cells and 4 genes each)
3. Opens the resulting census
4. Performs a set of assertions.

Ideas to improve this: 
1. Add different fixtures with more cases
2. Add more assertions


tools/cell_census_builder/experiment_builder.py
```diff
@@ -435,7 +435,7 @@ def _accumulate_all_X_layers(
         pres_data = raw_X.sum(axis=0) > 0
         if isinstance(pres_data, np.matrix):
             pres_data = pres_data.A
-        pres_data = pres_data[0]
+            pres_data = pres_data[0]
```
ebezzi, line 438: @bkmartinjr can you double check this? In the test case (probably due to the fact that we're using dense matrices), `pres_data` is a ndarray so this line won't work. I think this should not cause any issues.

tools/cell_census_builder/__main__.py
```diff
@@ -91,6 +91,24 @@ def main() -> int:
 
     return cc
 
+def prepare_file_system(soma_path: str, assets_path) -> None:
```
ebezzi, line 94: I moved this code into a function to facilitate testing. In the test, the `setUp` will prepare the environment so at least the datasets folder will be pre-created.

tools/cell_census_builder/experiment_builder.py
bkmartinjr: Given that we are testing the census builder, and it always requires Sparse matrices, the tests should do the same (i.e. the tests should not use dense matrices).

that said, I don't see why this code would break if you use dense matrices, and I suspect something else is going on here.  TL;DR - the change looks incorrect, or at least not addressing the core issue.

bkmartinjr: Which tests fails?

ebezzi: The builder itself fails due to (in my case) `pres_data` being an ndarray and not requiring the additional unpacking. I can rewrite the tests using sparse matrices and see if this addresses the issue.

ebezzi: Moving to a sparse matrix solved the issue. I'm wondering if it's worth adding that as an explicit constraint?

tools/cell_census_builder/tests/test_builder.py
```diff
@@ -1,26 +1,177 @@
 # General unit tests for cell_census_builder. Intention is to add more fine-grained tests for builder.
 import os
+import unittest
```
atolopko-czi, line 3: let's stick with pytest, which is already in use in this repo

tools/cell_census_builder/tests/test_builder.py
```diff
@@ -1,26 +1,177 @@
...
+        mock_prepare_file_system = patch("tools.cell_census_builder.__main__.prepare_file_system")
+        mock_prepare_file_system.start()
+
+        experiment_builders = make_experiment_builders(uricat(self.soma_path, CENSUS_DATA_NAME), [])
+        with patch("tools.cell_census_builder.__main__.build_step1_get_source_assets") as m:
+            m.return_value = self.datasets
+            from types import SimpleNamespace
+
+            args = SimpleNamespace(multi_process=False, consolidate=False, build_tag="test_tag")
+            build(args, self.soma_path, self.assets_path, experiment_builders)
```
atolopko-czi, line 135: should assert result == 0, since this indicates validation passed

atolopko-czi (changes requested):
This is great! A really nice starting point to build off of, when we introduce more granular unit tests.  Two requests before approving.

tools/cell_census_builder/tests/test_builder.py
bkmartinjr: wholeheartedly agree

If you need any help bootstrapping on pytest, just let us know.  For example, your `generate_h5ad` should be a pytest fixture.

ebezzi: I migrated everything to pytest. I think we can improve this further by moving the builder to a class-based approach. Since we have a fair amount of dependencies (ExperimentBuilder, datasets etc), we should be able to implement a relatively clean dependency injection setup using fixtures. I'd defer this work to a separate PR.

atolopko-czi:
> > Couple of additional test ideas
> 
> Alternatively, these could be added to the build-time validator. Not sure which is best - @atolopko-czi might have a viewpoint.

as validator tests sounds great; so we could pick these off as a separate issue

Bento007:
This is a great starting point for the rest of the unit tests. Great work!

tools/cell_census_builder/tests/test_builder.py
```diff
@@ -1,26 +1,189 @@
...
-    This test is not strictly necessary, but it validates the requirements that Cell Census
-    support unicode in DataFrame columns.
-    """
-    with TemporaryDirectory() as d:
-        pd_df = pd.DataFrame(data={"value": ["Ünicode", "S̈upport"]}, columns=["value"])
-        pd_df["soma_joinid"] = pd_df.index
-        s_df = soma.DataFrame(uri=os.path.join(d, "unicode_support")).create(
-            pa.Schema.from_pandas(pd_df, preserve_index=False), index_column_names=["soma_joinid"]
+
+class TestBuilder:
```
atolopko-czi, line 26: note that pytest does not require use of classes (maybe even discourages this?); at least they're not needed in the same way as for unittest

atolopko-czi (approved):
:shipit: (one comment)

tools/cell_census_builder/tests/test_builder.py
```diff
@@ -1,26 +1,189 @@
...
+            Dataset(dataset_id="second_id", corpora_asset_h5ad_uri="mock", dataset_h5ad_path=second_h5ad_path),
+        ]
+
+    def test_base_builder_creation(self, datasets: List[Dataset]) -> None:
+        """
+        Runs the builder, queries the census and performs a set of base assertions.
+        """
+
+        mock_prepare_file_system = patch("tools.cell_census_builder.__main__.prepare_file_system")
+        mock_prepare_file_system.start()
```
Bento007, line 137: Where are you stopping the patches?

tools/cell_census_builder/tests/test_builder.py
Bento007: If you are using the setup and teardown for every test, putting it in a class makes sense.

tools/cell_census_builder/tests/test_builder.py
Bento007: I suggest using a context manager or decorator to make sure the patch is cleaned up.

tools/cell_census_builder/tests/test_builder.py
ebezzi: They can be replaced with `setup_module` and `teardown_module` if we want to get rid of the class. I think both approaches are fine.

tools/cell_census_builder/tests/test_builder.py
Bento007: you don't have to double nest. You can do this:
``` python
with patch("tools.cell_census_builder.__main__.prepare_file_system") as mock_prepare_file_system, patch(
            "tools.cell_census_builder.__main__.build_step1_get_source_assets", return_value=datasets
        ) as mock_get_assets:
   pass
```

Bento007: decorators also work.

ebezzi: Ah nice, let me change to that. Unfortunately decorators conflict with pytest fixtures so we'll have to go with this.

Bento007:
Created https://github.com/chanzuckerberg/cell-census/issues/133 to track validator additions

tools/cell_census_builder/tests/test_builder.py
```diff
@@ -1,26 +1,186 @@
-# General unit tests for cell_census_builder. Intention is to add more fine-grained tests for builder.
 import os
 from tempfile import TemporaryDirectory
+from typing import List
+from unittest.mock import patch
```
bkmartinjr, line 4: No need to change, but I'm curious why you used `unittest.mock` instead of `pytest.monkeypatch`?

It isn't worth changing - we should just make sure we use a consistent style for this (and I don't have a strong opinion on this as pytest and unittest.mock play fine together)

tools/cell_census_builder/tests/test_builder.py
bkmartinjr: I have mostly used module-level fixtures for this kind of stuff, as I (personally) find them a bit clearer.

My only request:  as we expand the test footprint, we should have test style conventions and (mostly) use them.  That will keep the tests explicable for the next generation :-)

This is not a request for a change.

bkmartinjr: As an example of the above -- pytest has built-in support for temp paths and directories, so some of your `setup_class` is replicating built-in functionality.

ebezzi: We can switch to module-level tests. I agree consistency is good.

bkmartinjr (approved):
Seems fine functionality.  I find the style a bit confusing as it is still a hybrid unittest/pytest style (i.e., using classes instead of fixtures), but I'm OK with that as a starting point.

My suggestion:  land it, then have a discussion about test style, then refactor if needed.

tools/cell_census_builder/tests/test_builder.py
bkmartinjr: it isn't a drop-in replacement.  Lots of people prefer doing what you did as `pytest.monkeypatch` is kind of weird.

To be clear: I think your choice was fine!  I just think we should all treat that as our informal standard for future tests so we have consistency.

tools/cell_census_builder/tests/test_builder.py
bkmartinjr: Advocate you land this, and we do it as debt/backlog.  I.e., lets get your useful test landed :-)

ebezzi:
Thanks everyone for the reviews! I'm merging this with two follow-up action items:
1. Some additional test logic will be implemented in a separate validator test.
2. We need to have test style guidelines, especially around usage of module/classes and mocking.

PR #129, opened 2023-02-08
allows_duplicate schema tuning

bkmartinjr:
SOMA and the Cell Census require that all joinids in the obs/var axes, and all coordinates in the X layers, are unique. By default, the SOMA implementation enforces this constraint at _read_ time, incurring a read performance penalty. The Cell Census builder can guarantee this constraint at write time, providing a read-time performance win.

Changes in this PR:
* enhance validation of obs/var axes and the X layers to explicitly confirm this constraint.

WIP still to come:
* set the schema parameter `allows_duplicates` on the axes DataFrames and X SparseNDArrays

### DIFF
--- a/tools/cell_census_builder/__main__.py
+++ b/tools/cell_census_builder/__main__.py
@@ -117,7 +117,7 @@ def build(
         return 1
 
     # Ensure that the git tree is clean
-    if is_git_repo_dirty():
+    if not args.test_disable_dirty_git_check and is_git_repo_dirty():
         logging.error("The git repo has uncommitted changes - aborting build")
         return 1
 
@@ -313,8 +313,12 @@ def create_args_parser() -> argparse.ArgumentParser:
         default=True,
         help="Consolidate TileDB objects after build",
     )
-    # hidden option for testing. Will process only the first 'n' datasets
+    # hidden option for testing by devs. Will process only the first 'n' datasets
     build_parser.add_argument("--test-first-n", type=int, help=argparse.SUPPRESS)
+    # hidden option for testing by devs. Allow for WIP testing by devs.
+    build_parser.add_argument(
+        "--test-disable-dirty-git-check", action=argparse.BooleanOptionalAction, help=argparse.SUPPRESS
+    )
 
     # VALIDATE
     subparsers.add_parser("validate", help="Validate an existing cell census build")
--- a/tools/cell_census_builder/validate.py
+++ b/tools/cell_census_builder/validate.py
@@ -246,7 +246,7 @@ def validate_axis_dataframes(
     return True
 
 
-def _validate_X_layers_contents(args: Tuple[str, str, Dataset, List[ExperimentBuilder]]) -> bool:
+def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[ExperimentBuilder]]) -> bool:
     """
     Validate that a single dataset is correctly represented in the census.
     Intended to be dispatched from validate_X_layers.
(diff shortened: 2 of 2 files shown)


tools/cell_census_builder/__main__.py
```diff
@@ -313,8 +313,12 @@ def create_args_parser() -> argparse.ArgumentParser:
         default=True,
         help="Consolidate TileDB objects after build",
     )
-    # hidden option for testing. Will process only the first 'n' datasets
+    # hidden option for testing by devs. Will process only the first 'n' datasets
     build_parser.add_argument("--test-first-n", type=int, help=argparse.SUPPRESS)
+    # hidden option for testing by devs. Allow for WIP testing by devs.
+    build_parser.add_argument(
+        "--test-disable-dirty-git-check", action=argparse.BooleanOptionalAction, help=argparse.SUPPRESS
```
atolopko-czi, line 320: I'd prefer we don't hide the option. It's just "us" reading the --help.

tools/cell_census_builder/validate.py
```diff
@@ -288,6 +302,35 @@ def count_nonzero(arr: Union[sparse.spmatrix, npt.NDArray[Any]]) -> int:
...
+        return True
+
+    X_layer = se.ms["RNA"].X[layer_name]
+    n_rows, n_cols = X_layer.shape
+    ROW_SLICE_SIZE = 100_000
+
+    for row in range(row_range_start, min(row_range_stop, n_rows), ROW_SLICE_SIZE):
+        # work around TileDB-SOMA bug #900 which errors if we slice beyond end of shape
+        end_row = min(row + ROW_SLICE_SIZE, X_layer.shape[0] - 1)
+        slice_of_X = X_layer.read(coords=(slice(row, end_row),)).tables().concat()
```
atolopko-czi, line 324: just to visually group the comment with just the code it pertains to
```suggestion

        slice_of_X = X_layer.read(coords=(slice(row, end_row),)).tables().concat()
```

tools/cell_census_builder/validate.py
```diff
@@ -288,6 +302,35 @@ def count_nonzero(arr: Union[sparse.spmatrix, npt.NDArray[Any]]) -> int:
...
+        f"validate_no_dups_X start, {experiment_builder.name}, {layer_name}, rows [{row_range_start}, {row_range_stop})"
+    )
+    if layer_name not in se.ms["RNA"].X:
+        return True
+
+    X_layer = se.ms["RNA"].X[layer_name]
+    n_rows, n_cols = X_layer.shape
+    ROW_SLICE_SIZE = 100_000
+
+    for row in range(row_range_start, min(row_range_stop, n_rows), ROW_SLICE_SIZE):
```
atolopko-czi, line 321: A few performance Q's:
* How long does this take to run?
* Might we need to parallelize this operation? 
* Is 100K rows a "tuned" value or a WAG? 😄
* AFAIK, we can't avoid retrieving the matrix values themselves w/the current API. I'm not missing something right?

atolopko-czi (changes requested):
One minor request, and some performance-related Q's before approving.

tools/cell_census_builder/__main__.py
bkmartinjr: I assume you feel the same with the other --test option?

tools/cell_census_builder/validate.py
bkmartinjr: It adds another 30 minutes or so on my test, as written (run with `-mp --max-workers 16` on a _slower_ instance type than you are using)

It is parallelized :-)    See where it is invoked.

WAG, but based upon a memory budget (3 columns, 12 bytes per row, 95% sparsity, 60K columns), so a WAG, but not a SWAG.

Correct - unfortunately IMHO. I think Paul deleted this capability in his refactoring.  I consider it a future (post-1.0) discussion, as the use cases are _very_ narrow.

tools/cell_census_builder/__main__.py
atolopko-czi: yes--didn't realize it was also suppressed.

tools/cell_census_builder/validate.py
atolopko-czi: > It adds another 30 minutes

Great, that's acceptable (not hours).

> It is parallelized :-) See where it is invoked.

Indeed! Forest for the trees... And I should remember who I'm talking to here. 😄 

> I consider [specifying read attributes] a future (post-1.0) discussion, as the use cases are very narrow.

Yeah, this is about the only case I can imagine needing it. If performance was bad, was going to suggest diving down to the TileDB-Py API level for this. All good!

atolopko-czi (approved):
LGTM!

PR #131, opened 2023-02-09
Remove logic duplication in validator

Bento007:
Create the soma_path and asset_path the same way for `builder` and `validate` in `cell_census_builder`

### DIFF
--- a/tools/cell_census_builder/__main__.py
+++ b/tools/cell_census_builder/__main__.py
@@ -72,7 +72,6 @@ def main() -> int:
     process_initializer(args.verbose)
 
     # normalize our base URI - must include trailing slash
-    args.uri = args.uri if args.uri.endswith("/") else args.uri + "/"
     soma_path = uricat(args.uri, args.build_tag, "soma")
     assets_path = uricat(args.uri, args.build_tag, "h5ads")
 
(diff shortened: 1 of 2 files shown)


tools/cell_census_builder/__main__.py
```diff
@@ -72,7 +72,6 @@ def main() -> int:
     process_initializer(args.verbose)
 
     # normalize our base URI - must include trailing slash
-    args.uri = args.uri if args.uri.endswith("/") else args.uri + "/"
```
atolopko-czi, line 75: explain why this is safe? I'm not sure that it's _not_, just want to understand.

atolopko-czi (approved):
One Q, but LGTM, optimistically

tools/cell_census_builder/__main__.py
Bento007: we are doing it [here](https://github.com/chanzuckerberg/cell-census/blob/8ade9277a9235bba952b9120b01676414829cac3/tools/cell_census_builder/util.py#L61). If this change makes the addition of the trailing slash too obscure then we can keep this line.

ebezzi (approved):

PR #132, opened 2023-02-09
Port cell_census and notebooks to tiledbsoma 0.5.0a10

### LINKED ISSUE #106: open_soma needs to become stateful
When the SOMA API moves to stateful handles, we will need to update `open_soma` API to match.  As it just returns a Collection, this primarily is a change in:
* docstrings
* code that uses it
* notebooks

Where we used to say:
```
census = cell_census.open_soma()
```

we will now need to ensure it gets closed, i.e.:
```
with cell_census.open_soma() as census:
    ...
```

### LINKED ISSUE #145: Typo and broken link
## Motivation

Why is this work important to engineers?
- Typo in a demo notebook is misleading to new users; a broken link prevents users from accessing full information regarding cell census schema. 

## Definition of Done

What should the end result look like? What will have been changed?

In https://github.com/chanzuckerberg/cell-census/blob/main/api/python/notebooks/analysis_demo/comp_bio_query_data_and_metadata.ipynb notebook

The following: 

The human gene metadata of the Cell Census is located at census["census_data"]["homo_sapiens"].ms["RNA"].var. Similarly to the cell metadata, it is a SOMADataFrame and thus we can also use its method read().
The mouse gene metadata is at census["census_data"]["homo_sapiens"].ms["RNA"].var.

is replaced with :

The human gene metadata of the Cell Census is located at census["census_data"]["homo_sapiens"].ms["RNA"].var. Similarly to the cell metadata, it is a SOMADataFrame and thus we can also use its method read().
The mouse gene metadata is at census["census_data"]["mus_musculus"].ms["RNA"].var.

A link to the cell census schema (https://github.com/chanzuckerberg/cell-census/blob/main/docs/cell_census_schema_0.0.1.md#cell-metadata--census_objcensus_dataorganismobs--somadataframe) is replaced with a correct link. 

## Tasks

Please see above. 

Detail the specific tasks that can be used to accomplish the desired changes.
If detailed steps cannot be provided at this time, please file a [Tech Proposal](https://docs.google.com/document/d/1o2vuvl-kXwRJN1nBoPzJS_MAQgDGYnjmPZWa4qRDi-I/edit#heading=h.7dvzhm7gqc3v) instead.

- [ ]
- [ ]

bkmartinjr:
Primary changes:
* port to open/close SOMA API.  Fixes #106 
* rework cell_census.open_soma for clarity
* revise all uses of open_soma to open/close or use a context manager
* re-run all notebooks
* revise affected tests

Additional changes:
* notebook typo fix.  Fixes #145 

Partially addresses #151


api/python/cell_census/src/cell_census/open.py
```diff
@@ -3,36 +3,45 @@
...
-# TODO:  there are several changes pending for `open_soma`:
-# * add config or context hook when it is further defined, allowing config overrides.
-# * port to the factory `soma.open()` when it exists
-#
-def open_soma(*, census_version: Optional[str] = "latest", uri: Optional[str] = None) -> soma.Collection:
+    if not context:
+        # if no user-defined context, cell_census defaults take precedence over SOMA defaults
+        tiledb_config = DEFAULT_TILEDB_CONFIGURATION
+        if s3_region is not None:
+            tiledb_config.copy()["vfs.s3.region"] = s3_region
```
ebezzi, line 27: Isn't this a no-op?

bkmartinjr: which part?   The assignment of region is definitely not a noop (i.e., it sets the config to the region specified by the cell census directory), and leaves it unset (ie. use defaults) when the directory lacks a region.

That said, I'm often _non-pythonic_ so if I'm doing something silly here, I would love to learn!

ebezzi: My understanding is that calling `.copy()` will return a copy of the dictionary, which you're modifying, but since you're not assigning to anything the result will be lost.

bkmartinjr: oh my - good call. I'm clearly at the point I can't see forest for trees in my code!

bkmartinjr: OK, I cleaned it up - whew, good catch.

atolopko-czi (approved):
LGTM, but not checking notebooks.

bkmartinjr:
> LGTM, but not checking notebooks.

@pablo-gar - do you want to review notebooks, or are you OK if I just merge?

ebezzi (approved):

pablo-gar (approved):
LGTM

PR #146, opened 2023-02-10
GitHub Action workflow for builder unit test

ebezzi:
Add a Github Action workflow (a step, really) for running the builder test.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Test with pytest (API)
         run: |
           pytest ./api/python/cell_census/tests/
-jobs:
+          
   unit_tests_builder:
         runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/unittests.yml
```diff
@@ -25,6 +25,30 @@ jobs:
...
         run: |
           pytest ./api/python/cell_census/tests/
+
+  unit_tests_builder:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: ["3.9", "3.10"]
```
bkmartinjr, line 38: Tiny, tiny nit:  i'm not sure we need to test on more than one version, as we are the only users.  So strictly from a "how fast is our CI", I suggest a simple test matrix.

bkmartinjr (approved):
LGTM, with one optional nit you can decide to fix or not :-)

.github/workflows/unittests.yml
ebezzi: I only set 3.10

atolopko-czi (approved):
👏

PR #148, opened 2023-02-10
Report cell census changes between versions

ebezzi:
The `census_summary` script can now be used to generate a diff

Example usage: 

```python -m tools.cell_census_builder.scripts.release.census_summary -c 2023-02-06 diff -p 2023-01-30```

Example output:

```
Datasets that were added
d7d7e89c-c93a-422d-8958-9b4a90b69558

3c75a463-6a87-4132-83a8-c3002624394d

5500c673-1610-40a0-86d9-64d987ae50e6

a5d95a42-0137-496f-8a60-101e17f263c8

Previous homo_sapiens cell count: 43246174, current homo_sapiens cell count: 43420131, delta 173957

Previous mus_musculus cell count: 3922090, current mus_musculus cell count: 3922090, delta 0
```

Other cases are computed but only displayed if the difference is not null.

### DIFF
--- a/tools/cell_census_builder/scripts/release/census_summary.py
+++ b/tools/cell_census_builder/scripts/release/census_summary.py
@@ -48,15 +48,13 @@ if __name__ == "__main__":
         removed_datasets = prev_dataset_ids - curr_datasets_ids
         if added_datasets:
             print("Datasets that were added")
-            for d in added_datasets:
-                print(d)
-                print()
+            added_datasets_df = curr_datasets[curr_datasets["dataset_id"].isin(added_datasets)]
+            print(added_datasets_df[["dataset_id", "dataset_title", "collection_name"]])
 
         if removed_datasets:
             print("Datasets that were removed")
-            for d in removed_datasets:
-                print(d)
-                print()
+            removed_datasets_df = prev_datasets[prev_datasets["dataset_id"].isin(removed_datasets)]
+            print(removed_datasets_df[["dataset_id", "dataset_title", "collection_name"]])
 
         # Datasets in both versions but that have differing cell counts
         joined = prev_datasets.join(
--- a/tools/cell_census_builder/scripts/release/census_summary.py
+++ b/tools/cell_census_builder/scripts/release/census_summary.py
@@ -1,3 +1,4 @@
+import argparse
 import sys
 
 import cell_census
@@ -5,10 +6,8 @@ import pandas as pd
 
 from tools.cell_census_builder.globals import CENSUS_DATA_NAME, CENSUS_INFO_NAME
 
-if __name__ == "__main__":
-    census_version = sys.argv[1] if len(sys.argv) > 1 else "latest"
-    previous_census_version = sys.argv[2] if len(sys.argv) > 2 else None
 
+def display_summary(census_version: str) -> int:
     census = cell_census.open_soma(census_version=census_version)
 
     COLS_TO_QUERY = [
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tools/cell_census_builder/scripts/release/census_summary.py [hunk 1]
```diff
@@ -2,10 +2,12 @@
 
 import cell_census
 import pandas as pd
-from cell_census_builder.globals import CENSUS_DATA_NAME
+
+from tools.cell_census_builder.globals import CENSUS_DATA_NAME, CENSUS_INFO_NAME
 
 if __name__ == "__main__":
     census_version = sys.argv[1] if len(sys.argv) > 1 else "latest"
+    previous_census_version = sys.argv[2] if len(sys.argv) > 2 else None
```
ebezzi, line 10: Here we could infer the previous version automatically. Does the Census API expose a list of all available versions?

bkmartinjr: You can't infer it.  "previous" is entirely context dependent at this point, and we will need to run this script manually, i.e.,
> generate_diff census_tag_1 census_tag_2

I'm not sure you can ever automate this for all cases, as it is a human who eventually decides what builds are significant.

tools/cell_census_builder/scripts/release/census_summary.py
```diff
@@ -2,10 +2,12 @@
...
 
     census = cell_census.open_soma(census_version=census_version)
```
bkmartinjr, line 12: please port to latest API.  I aspire that our internal code to reflects latest & best practices of API use, as it will be demo for others.

In this case, for example, I would base this off of PR #132 and use a context manager.

```python
with cell_census.open_soma() as census:
    ...
```

tools/cell_census_builder/scripts/release/census_summary.py
```diff
@@ -28,3 +30,64 @@
...
+
+        # Total cell count deltas by experiment (mouse, human)
+        for organism in ["homo_sapiens", "mus_musculus"]:
+            curr_count = census[CENSUS_DATA_NAME][organism]["obs"].count
+            prev_count = previous_census[CENSUS_DATA_NAME][organism]["obs"].count
+            print(
+                f"Previous {organism} cell count: {prev_count}, current {organism} cell count: {curr_count}, delta {curr_count - prev_count}"
+            )
+
+        # Deltas between summary_cell_counts dataframes
```
ebezzi, line 78: I have the code for this but the dataframe here can be quite big. The current runs shows 65 different entries for total cell counts. Do we want to display it regardless? An alternative is to only show the deltas that are bigger than a certain threshold.

tools/cell_census_builder/scripts/release/census_summary.py
ebezzi: Good point, thanks Bruce!

bkmartinjr: In addition, the `open_soma` will raise if the version is unknown, so I would catch and report that for both open_soma calls

tools/cell_census_builder/scripts/release/census_summary.py
```diff
@@ -28,3 +30,64 @@
...
+        # Datasets removed, added
+        curr_datasets_ids = set(curr_datasets["dataset_id"])
+        prev_dataset_ids = set(prev_datasets["dataset_id"])
+
+        added_datasets = curr_datasets_ids - prev_dataset_ids
+        removed_datasets = prev_dataset_ids - curr_datasets_ids
+        if added_datasets:
+            print("Datasets that were added")
+            for d in added_datasets:
+                print(d)
```
bkmartinjr, line 51: out of curiosity, was there a format requested for this report?  I suspect that dataset ID would be complemented by the dataset title, and maybe collection title?  @pablo-gar - opinions on how fancy you want this to be?

tools/cell_census_builder/scripts/release/census_summary.py
bkmartinjr: ps. alternatively, you could land this on the old API, but it will break soon :-)

tools/cell_census_builder/scripts/release/census_summary.py
```diff
@@ -28,3 +30,64 @@
...
+        datasets_with_different_cell_counts = joined.loc[
+            joined["dataset_total_cell_count_prev"] != joined["dataset_total_cell_count_curr"]
+        ][["dataset_id", "dataset_total_cell_count_prev", "dataset_total_cell_count_curr"]]
+
+        if not datasets_with_different_cell_counts.empty:
+            print("Datasets that have a different cell count")
+            print(datasets_with_different_cell_counts)
+
+        # Total cell count deltas by experiment (mouse, human)
+        for organism in ["homo_sapiens", "mus_musculus"]:
```
bkmartinjr, line 75: there is a slightly more future-proof way to do this, as anything inside `census['census_data']` is an Experiment.  So I suggest:

```
for organism_name in census[CENSUS_DATA]:
    curr_count = census[CENSUS_DATA][organism_name].obs.count
    ...
```

Or even better:
```
for organism_name, experiment in census[CENSUS_DATA].items():
    curr_count = experiment.obs.count
    ...
```

Also of note:  pre-defined fields like `obs` are attributes, so you can use `experiment.obs`, rather than the more verbose `experiment["obs"]`

tools/cell_census_builder/scripts/release/census_summary.py
```diff
@@ -28,3 +30,64 @@
...
+            print(
+                f"Previous {organism} cell count: {prev_count}, current {organism} cell count: {curr_count}, delta {curr_count - prev_count}"
+            )
+
+        # Deltas between summary_cell_counts dataframes
+
+        # Genes removed, added
+        for organism in ["homo_sapiens", "mus_musculus"]:
+            curr_genes = census[CENSUS_DATA_NAME][organism]["ms"]["RNA"]["var"].read().concat().to_pandas()
+            prev_genes = previous_census[CENSUS_DATA_NAME][organism]["ms"]["RNA"]["var"].read().concat().to_pandas()
```
bkmartinjr, line 110: "ms" is predefined in an Experiment,  so you can say `.ms` rather than `["ms"]`

tools/cell_census_builder/scripts/release/census_summary.py
ebezzi: No particular format requests, but they should be easy enough to accommodate. I can add the dataset title and collection title immediately.

tools/cell_census_builder/scripts/release/census_summary.py
```diff
@@ -2,10 +2,12 @@
 
 import cell_census
 import pandas as pd
-from cell_census_builder.globals import CENSUS_DATA_NAME
+
+from tools.cell_census_builder.globals import CENSUS_DATA_NAME, CENSUS_INFO_NAME
 
 if __name__ == "__main__":
```
bkmartinjr, line 8: From a sustainability viewpoint, it would be much better if this code:
* had a real main() function
* returned a system result code (ie.,  `sys.exit(main())`)
* used argparse, so we could add additional features down the road

bkmartinjr: For example, I think that it is _possible_ that future request might include a CLI param to specify what output format is desired:  text, markdown, html, ....

>  python census_summary_diff.py --output-format markdown new-tag old-tag

bkmartinjr (commented):
Suggest giving the program a more descriptive name, as it is really generate a summary of the differences of two census.

So maybe:   census_diff_summary.py   or   summarize_census_diff.py

bkmartinjr (changes requested):
A few changes requested, based on my assumption that this script will become a part of our regular release tooling, and therefore should be made robust and maintainable.  Feel free to reel me in a bit if that assumption is false!

tools/cell_census_builder/scripts/release/census_summary.py
bkmartinjr: that is a Product question - I don't know exactly how they intend to use this.  @pablo-gar - please specify your desired report format.

My own opinion:  the point of this is to summarize _all_ major changes, so we want to list all of the differences.

ebezzi: I added them to the code.

tools/cell_census_builder/scripts/release/census_summary.py
ebezzi: It's certainly worth doing. I will work on it.

tools/cell_census_builder/scripts/release/census_summary.py
ebezzi: I believe we need your PR merged to implement this change

bkmartinjr: Yes, I'm blocked by @pablo-gar.  I have just DM'ed him again and will follow-up by mid-day if we don't have motion.

tools/cell_census_builder/scripts/release/census_summary.py
ebezzi: I addd argparse and reformatted this into two separate subcommands: `summarize` and `diff`. LMK if it makes sense to you.

tools/cell_census_builder/scripts/release/census_summary.py
```diff
@@ -28,3 +29,138 @@
...
+
+    return 0
+
+
+def create_args_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(prog="cell_census_summary")
+    parser.add_argument("-c", "--census-version", default="latest", help="Version of the census. Defaults to latest")
+    subparsers = parser.add_subparsers(required=True, dest="subcommand")
+
+    # BUILD
```
bkmartinjr, line 142: nit: comment doesn't match code

tools/cell_census_builder/scripts/release/census_summary.py
```diff
@@ -28,3 +29,138 @@
...
+    subparsers.add_parser("summarize", help="Summarize the cell census")
+
+    # VALIDATE
```
bkmartinjr, line 145: nit: comment doesn't match code

bkmartinjr (approved):
LGTM - caveat did a quick review.

PR #154, opened 2023-02-10
feature: update builder tiledbsoma release

### LINKED ISSUE #87: Update builder for tiledbsoma create API changes
TileDB-SOMA Issues https://github.com/single-cell-data/TileDB-SOMA/issues/540 and https://github.com/single-cell-data/TileDB-SOMA/issues/638 will introduce API changes for SOMA object creation, allowing for creation of children objects directly from the parent.  Currently children are created independently and then added to the parent.  Builder will need to be updated to accommodate these API changes.

See also:
https://github.com/single-cell-data/SOMA/pull/96 (SOMA spec changes)

### LINKED ISSUE #182: Cell Census builder uses TileDB-SOMA 1.0 RC
If blocked by TileDB-SOMA 1.0 RC, then builder should use latest version by end of sprint.

This issue should be closed when TileDB-SOMA 1.0 RC is used for Cell Census builder.

### LINKED ISSUE #207: Builder validate hangs
The validation step `_validate_X_layers_contents_by_dataset`, running as a child process and invoked by `validate_X_layers`, hangs without completing. 

Originally thought this was due to unreported OOM conditions, but no longer clear if that's the problem.

On branch `atol/validate-memleak-debug`, experimented by trying to free memory potentially held by the AnnData objects. But active and inactive memory still decreased and increased monotonically, respectively.  Changed multiprocessing implementation to use `multiprocessing.Pool` directly, instead of `concurrent.future` wrapping API, as this allowed configuration to _not_ reuse processes (1 task per proceses, each process is created anew).  `vmstat` showed that this did help with memory reclamation. However, the child processses would still eventually hang. Seemingly when only < N tasks remained, where N is the worker (child process) count. E.g., after processing 478 of 490, with 16 workers, the pool would hang. 

Add copious logging output to determine where child processes "gave up", and whether unreported exceptions might be involved. No exceptions seem to be occurring in the child processes. All that can be determined is that the `finally` block is not always reached (log "leaving" message). 

One finding is that removing the AnnData reads _does_ allow the validation step to complete. The TileDB/SOMA reads are therefore not implicated.

Tried other miscellaneous things such as:
* Tried `map` vs `imap_unordered` for polling the results from the process pool. (Did not try "async" methods, however.)
* Replacing use of `reopen_experiment_builders()` and using `raw Experiment.open()` instead, just in case context manager use was somehow related.
* Instead of using open_anndata(), which is a generator function, tried directly opening anndata w/o generator.

atolopko-czi:
Fixes #87 
Fixes #182 
Fixes #207 

Changes:
* Update to tiledbsoma 1.0rc0.  Fixes #182 
* SOMA child objects are now created directly from parents, within context managers
* explicit read/write timestamps are specified as workaround for TileDB-SOMA bug https://github.com/single-cell-data/TileDB-SOMA/issues/919.
* Multiprocessing code reopens necessary TileDBObjects within child processes, since context managers "open" state does not cascade into child processes (this would ill-behaved even if it worked, and it doesn't due to pickling errors).
* Refactored build steps for improved clarity and to accommodate use of context managers. We now have 5 explicit build steps. In particular, separated out the axis & X building steps and the root collection tree creation.
* Remove `BuildState` state machine that tracks progression of build steps, since it would have changed so much (could be reinstated).
* Builder now robust to having an experiment (organism) with 0 datasets/cells, which is useful for dev-time testing.
* Validator performs reads using open() context manager
* Validator assert presence matrix
* test_builder.py test fixture updated to have a dataset for each organism for better test coverage
* update the cell census schema version to latest (`0.1.1`)
* add a codecov.yml that disables status checks
* root cause of #207 was an undetected OOM which occurred in such a manner as to deadlock the process pool. The standard means of detection (a BrokenProessPool error) will not be thrown in this condition. Added logging code to detect and report that this error has likely occured, to help with diagnosis.

Note on out-of-memory handling -- the system remains somewhat fragile when the host system runs out of memory. Most of the time the software will log an error and exit, but there are known (reproducible) cases where the multi-process pool will deadlock and hang.  I have added logging to detect and report this condition, but there is not an obvious means to prevent the hang.

The _best_ solution is to stop running the system on hosts with insufficient memory.  As the average memory utilization is much smaller than the peak memory utilization, an appropriate solution is to mount swap for robust operation as our data set size grows. There are admin scripts available to make this easy when building the census by hand. For future use in automated batch environments, we will need to explore means to configure the instance in the job workflow.


tools/cell_census_builder/__main__.py
```diff
@@ -133,63 +131,52 @@ def build(
...
-    top_level_collection, filtered_datasets = build_step2_create_axis(
-        soma_path, assets_path, datasets, experiment_builders, args
-    )
+    # Step 1 - get all source datasets
+    datasets = build_step1_get_source_datasets(args, assets_path)
+    filtered_datasets = filter_datasets(assets_path, datasets, experiment_builders)
     assign_soma_joinids(filtered_datasets)
+
+    # Step 2 - create root collection, and all child objects, but do not populate X layers
+    root_collection = build_step2_create_root_collection(soma_path, filtered_datasets, experiment_builders)
```
atolopko-czi, line 140: renamed step 2 and moved the axis building to the next step

tools/cell_census_builder/__main__.py
```diff
@@ -133,63 +131,52 @@ def build(
...
+    build_step3_populate_axes_and_X_layers(assets_path, filtered_datasets, experiment_builders, args)
     gc.collect()
 
     # Write out dataset manifest and summary information
-    create_dataset_manifest(top_level_collection[CENSUS_INFO_NAME], filtered_datasets)
-    create_census_summary_cell_counts(
-        top_level_collection[CENSUS_INFO_NAME], [e.census_summary_cell_counts for e in experiment_builders]
-    )
-    create_census_summary(top_level_collection[CENSUS_INFO_NAME], experiment_builders, args.build_tag)
+    build_step4_populate_summary_info(root_collection, experiment_builders, filtered_datasets, args.build_tag)
```
atolopko-czi, line 149: wrapped these into an official step 4, with start/end timing output

tools/cell_census_builder/__main__.py
```diff
@@ -210,80 +197,123 @@ def build_step1_get_source_assets(args: argparse.Namespace, assets_path: str) ->
...
-    Returns: the filtered datasets that will be included. This is simply
-    an optimization to allow subsequent X matrix writing to skip unused
-    datasets.
+    Returns: the root collection.
     """
     logging.info("Build step 2 - axis creation - started")
 
-    top_level_collection = create_top_level_collections(soma_path)
+    with soma.Collection.create(soma_path, context=SOMA_TileDB_Context()) as root_collection:
+        populate_root_collection(root_collection)
```
atolopko-czi, line 211: new tiledbsoma api usage

tools/cell_census_builder/__main__.py
```diff
@@ -210,80 +197,123 @@ def build_step1_get_source_assets(args: argparse.Namespace, assets_path: str) ->
...
+    experiment_builders: List[ExperimentBuilder],
+    filtered_datasets: List[Dataset],
+    build_tag: str,
+) -> None:
+    logging.info("Build step 4 - summary info - started")
+
+    with soma.Collection.open(root_collection[CENSUS_INFO_NAME].uri, "w", context=SOMA_TileDB_Context()) as census_info:
+        create_dataset_manifest(census_info, filtered_datasets)
+        create_census_summary_cell_counts(census_info, [e.census_summary_cell_counts for e in experiment_builders])
+        create_census_summary(census_info, experiment_builders, build_tag)
```
atolopko-czi, line 312: refactoring these steps into a single method allows for shared use of Collection.open() context manager.

tools/cell_census_builder/experiment_builder.py
```diff
@@ -70,29 +78,11 @@ class ExperimentBuilder:
...
-        Initialized = 0
-        Created = 1
-        AxisWritten = 2
-        X_Created = 3
-        X_JoinIdMetadataCreated = 4
-        X_Written = 5
-        X_Presence_Written = 6
-
-        def next(self) -> "ExperimentBuilder.BuildState":
-            return ExperimentBuilder.BuildState(self.value + 1)
```
atolopko-czi, line 87: Due to refactoring, it was going to take work to make this work again; I opted to remove this state machine.  Useful for sanity checks, but the validator is a reasonable check now that census is built correctly. I can be persuaded to reinstate.

tools/cell_census_builder/experiment_builder.py
```diff
@@ -203,22 +168,15 @@ def accumulate_axes(self, dataset: Dataset, ad: anndata.AnnData, progress: Tuple
...
-        assert se.exists()
+        self.populate_obs_axis(obs_df)
 
-        pa_table = pa.Table.from_pandas(
-            obs_df,
-            preserve_index=False,
-            columns=list(CENSUS_OBS_TERM_COLUMNS),
-        )
-        for pa_batch in pa_table.to_batches():
-            se.obs.write(pa_batch)
```
atolopko-czi, line 221: refactored above code into method, `populate_obs_axis`

tools/cell_census_builder/experiment_builder.py
```diff
@@ -203,22 +168,15 @@ def accumulate_axes(self, dataset: Dataset, ad: anndata.AnnData, progress: Tuple
...
+        self.dataset_obs_joinid_start[dataset.dataset_id] = self.n_obs
```
atolopko-czi, line 179: Now accumulating this data "as we go" instead of re-reading the census obs dataframe later to acquire it; this seemed like a safe replacement

tools/cell_census_builder/validate.py
```diff
@@ -66,101 +66,115 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
         |   +-- homo_sapiens: soma.Experiment
         |   +-- mus_musculus: soma.Experiment
     """
-    census = soma.Collection(soma_path, context=SOMA_TileDB_Context())
```
atolopko-czi, line 69: all of the below is just indentation changes due to use of open() context manager

tools/cell_census_builder/validate.py
```diff
@@ -66,101 +66,115 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
...
+                    assert rna.X[lyr].soma_type == "SOMASparseNDArray"
+
+            # and a dataset presence matrix
+            # dataset presence only exists if there are cells in the measurement
+            if eb.n_obs > 0:
+                assert FEATURE_DATASET_PRESENCE_MATRIX_NAME in rna
+                assert soma.SparseNDArray.exists(rna[FEATURE_DATASET_PRESENCE_MATRIX_NAME].uri)
+                assert rna[FEATURE_DATASET_PRESENCE_MATRIX_NAME].soma_type == "SOMASparseNDArray"
+                presence = rna[FEATURE_DATASET_PRESENCE_MATRIX_NAME].read().tables().concat().to_pandas()
+                assert len(presence) > 0
```
atolopko-czi, line 138: I did add this one line to assert presence has some data (caught a bug in refactoring)

tools/cell_census_builder/__main__.py
```diff
@@ -210,80 +197,123 @@ def build_step1_get_source_assets(args: argparse.Namespace, assets_path: str) ->
...
+
+def build_step3_populate_axes_and_X_layers(
     assets_path: str,
     filtered_datasets: List[Dataset],
     experiment_builders: List[ExperimentBuilder],
     args: argparse.Namespace,
 ) -> None:
     """
-    Create and populate all X layers
+    Populate all X layers.
```
bkmartinjr, line 277: and axes (also fix logging statement below)

tools/cell_census_builder/__main__.py
```diff
@@ -210,80 +197,123 @@ def build_step1_get_source_assets(args: argparse.Namespace, assets_path: str) ->
...
+                max_dataset_joinid = max(d.soma_joinid for d in datasets)
+                eb.experiment.ms["RNA"].add_new_sparse_ndarray(
+                    FEATURE_DATASET_PRESENCE_MATRIX_NAME, type=pa.bool_(), shape=(max_dataset_joinid + 1, eb.n_var)
+                )
+
+    return dataset_total_cell_count
 
-def build_step3_create_X_layers(
+
+def build_step3_populate_axes_and_X_layers(
```
bkmartinjr, line 270: this function opens each AnnData twice, which isn't really necessary.  with the soma_joinids already defined, you could refactor this so that:
* for each AnnData, have a sub-process:
    1. open the full AnnData (backed=None)
    2. append obs, var and X[raw]
There is no race as there are no duplicate coordinates.

bkmartinjr: In addition, you could process all AnnData concurrently with this model, or better yet, do all (experiment, AnnData) pairs concurrently.

With your change (which I like), only `build_step2` is single-threaded.

tools/cell_census_builder/__main__.py
```diff
@@ -210,80 +197,123 @@ def build_step1_get_source_assets(args: argparse.Namespace, assets_path: str) ->
...
 
+def build_step4_populate_summary_info(
+    root_collection: soma.Collection,
+    experiment_builders: List[ExperimentBuilder],
+    filtered_datasets: List[Dataset],
+    build_tag: str,
+) -> None:
+    logging.info("Build step 4 - summary info - started")
+
+    with soma.Collection.open(root_collection[CENSUS_INFO_NAME].uri, "w", context=SOMA_TileDB_Context()) as census_info:
```
bkmartinjr, line 309: you could also use `soma.open()` to save some typing....

tools/scripts/requirements.txt
```diff
@@ -5,7 +5,7 @@ numpy
 # NOTE: You can also build this dependency from source, per ./notebooks/README.md.
 # NOTE: The builder's version of tiledbsoma MUST be <= the API's tiledbsoma version, to ensure reader compatibility
 # of TileDB on-disk storage format
-tiledbsoma==0.5.0a6
+tiledbsoma==0.5.0a10
```
bkmartinjr, line 8: a11 is out now (and I've released cell_census on top of it)

tools/cell_census_builder/consolidate.py
```diff
@@ -5,6 +5,7 @@
 
 import tiledbsoma as soma
 
+from .globals import SOMA_TileDB_Context
 from .mp import create_process_pool_executor
 
 if soma.get_storage_engine() == "tiledb":
```
bkmartinjr, line 11: this import of `tiledb` should be moved into the `consolidate_tiledb_object` function (and yeah, this is my latent bug :-)).

tools/cell_census_builder/globals.py
```diff
@@ -216,11 +217,28 @@
...
 def SOMA_TileDB_Context() -> soma.options.SOMATileDBContext:
     global _SOMA_TileDB_Context
     if _SOMA_TileDB_Context is None:
-        _SOMA_TileDB_Context = soma.options.SOMATileDBContext(tiledb_ctx=TileDB_Ctx())
+        # Set write timestamp to "now", so that we use consistent timestamps across all writes (mostly for aesthetic
+        # reasons). Set read timestamps to be same as write timestamp so that post-build validation reads can "see"
+        # the writes. Without setting read timestamp explicitly, the read timestamp would default to a time that
+        # prevents seeing the builder's writes.
+        _SOMA_TileDB_Context = soma.options.SOMATileDBContext(
+            tiledb_ctx=TileDB_Ctx(),
```
bkmartinjr, line 235: what happens if `_TileDB_Ctx` is modified after the initial per-process call to `SOMA_TileDB_Context`?  It would not pick up the change.

I suggest you change line 229 to read:

> if _SOMA_TileDB_Context is None or _SOMA_TileDB_Context.tiledb_ctx != _TileDB_Ctx:

tools/cell_census_builder/tests/test_builder.py
```diff
@@ -9,18 +10,15 @@
...
+from cell_census_builder.datasets import Dataset
+from cell_census_builder.globals import (
     CENSUS_DATA_NAME,
     CENSUS_INFO_NAME,
     CENSUS_X_LAYERS_PLATFORM_CONFIG,
 )
-from tools.cell_census_builder.mp import process_initializer
-from tools.cell_census_builder.util import uricat
+from cell_census_builder.mp import process_initializer
+from scipy.sparse import csc_matrix
```
bkmartinjr, line 21: super tiny nit: I believe we use the Google python style, right?  So you should do `from scipy import sparse` and then spell out `sparse.csc_matrix`

[reference](https://google.github.io/styleguide/pyguide.html#22-imports)

tools/cell_census_builder/tests/test_builder.py
```diff
@@ -46,15 +44,21 @@ def setup_class(cls) -> None:
...
+        CENSUS_X_LAYERS_PLATFORM_CONFIG["raw"]["tiledb"]["create"]["dims"]["soma_dim_0"]["tile"] = 2
+        CENSUS_X_LAYERS_PLATFORM_CONFIG["raw"]["tiledb"]["create"]["dims"]["soma_dim_1"]["tile"] = 2
 
     @classmethod
     def teardown_class(cls) -> None:
         cls.td.cleanup()
 
-    @pytest.fixture
-    def h5ad(self) -> anndata.AnnData:
+    @dataclass
```
bkmartinjr, line 54: suggest `@dataclass(frozen=True)`

bkmartinjr: also, nit:  have been migrating from dataclass to attrs based upon Paul's recommendation. It is definintely an improvement in some ways.

tools/cell_census_builder/tests/test_builder.py
```diff
@@ -46,15 +44,21 @@ def setup_class(cls) -> None:
...
+    class Organism:
+        name: str
+        organism_ontology_term_id: str
+
+    ORGANISMS = [Organism("homo_sapiens", "NCBITaxon:9606"), Organism("mus_musculus", "NCBITaxon:10090")]
+
+    def h5ad(self, organism: Organism) -> anndata.AnnData:
         X = np.random.randint(5, size=(4, 4))
         # The builder only supports sparse matrices
         X = csc_matrix(X)
```
bkmartinjr, line 64: Use [scipy.sparse.random](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.random.html#scipy.sparse.random) to directly create a random sparse matrix.  Will be much better as it will  have sparsity (this code is actually a sparse encoding of a dense matrix)

tools/cell_census_builder/tests/test_builder.py
```diff
@@ -87,13 +91,13 @@ def h5ad(self) -> anndata.AnnData:
         obs = obs_dataframe
 
         # Create vars
-        feature_name = pd.Series(data=["a", "b", "c", "d"])
+        feature_name = pd.Series(data=[f"{organism.name}_{g}" for g in ["a", "b", "c", "d"]])
         var_dataframe = pd.DataFrame(
```
bkmartinjr, line 95: while it works today, if we ever start enforcing the non-uniqueness requirement for feature_reference, this dataframe will fail as every row is identical (except for the index).

Easy to fix - just generate the `data` rows using same method as `feature_name`

tools/cell_census_builder/validate.py
```diff
@@ -66,101 +66,115 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
...
-    assert sorted(census_info[CENSUS_DATASETS_NAME].keys()) == sorted(CENSUS_DATASETS_COLUMNS + ["soma_joinid"])
-    assert sorted(census_info[CENSUS_SUMMARY_CELL_COUNTS_NAME].keys()) == sorted(
-        list(CENSUS_SUMMARY_CELL_COUNTS_COLUMNS) + ["soma_joinid"]
-    )
-    assert sorted(census_info[CENSUS_SUMMARY_NAME].keys()) == sorted(["label", "value", "soma_joinid"])
-
-    # there should be an experiment for each builder
-    census_data = census[CENSUS_DATA_NAME]
-    for eb in experiment_builders:
+    with soma.Collection.open(soma_path, context=SOMA_TileDB_Context()) as census:
```
bkmartinjr, line 69: you could also use `soma.open()` given that you explicitly check the type in the next line

tools/cell_census_builder/validate.py
```diff
@@ -66,101 +66,115 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
...
+        assert census.soma_type == "SOMACollection"
+        assert soma.Collection.exists(census.uri)
+        assert "cxg_schema_version" in census.metadata and census.metadata["cxg_schema_version"] == CXG_SCHEMA_VERSION
```
bkmartinjr, line 72: better style

> assert census.metadata.get("cxg_schema_version") == CXG_SCHEMA_VERSION

tools/cell_census_builder/validate.py
```diff
@@ -66,101 +66,115 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
...
         assert (
-            eb.name in census_data
-            and census_data[eb.name].exists()
-            and census_data[eb.name].soma_type == "SOMAExperiment"
+            "census_schema_version" in census.metadata
```
bkmartinjr, line 74: ibid

tools/cell_census_builder/validate.py
bkmartinjr: any reason to not add the same check for obs/var/X?

You can also use `nnz` to make it faster, rather than having to read the data as you did here.

tools/cell_census_builder/validate.py
```diff
@@ -251,86 +265,87 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
...
-            assert raw_nnz == count_nonzero(
-                ad.raw.X
-            ), f"{eb.name}:{dataset.dataset_id} 'raw' nnz mismatch {raw_nnz} vs {count_nonzero(ad.raw.X)}"
+            raw_nnz = 0
+            if len(soma_joinids) > 0:
+                assert "raw" in se.ms["RNA"].X
+                assert soma.SparseNDArray.exists(se.ms["RNA"].X["raw"].uri)
+
+                def count_elements(arr: soma.SparseNDArray, join_ids: npt.NDArray[np.int64]) -> int:
+                    # TODO XXX: Work-around for regression TileDB-SOMA#473
```
bkmartinjr, line 296: this bug is fixed, so you could remove the work-around

bkmartinjr (changes requested):
A bunch of minor or stylistic stuff, plus:
* change adds an extra pass through all AnnData.  Could be optimized as noted.
* suggest making SOMATileDBContext singleton more robust to changes in the wrapped TileDBCtx singleton
* might as well use tiledbsoma alpha 11
As discussed offline, I'm fine if you want to push some of this to another PR.

tools/cell_census_builder/__main__.py
atolopko-czi: done

tools/cell_census_builder/__main__.py
atolopko-czi: Not sure I follow, since all of these methods are scoped within the Collection.open() ctx mgr now.

tools/cell_census_builder/consolidate.py
atolopko-czi: fixed!

tools/cell_census_builder/tests/test_builder.py
atolopko-czi: done

tools/cell_census_builder/tests/test_builder.py
atolopko-czi: good call, done.  we have 2 other dataclasses still in the code base, but I'm leaving those alone for now w/TODOs

tools/cell_census_builder/__main__.py
bkmartinjr: @atolopko-czi - was this follow-up comment directed at some other comment?  I'm having trouble following what you are "not following" :-)

tools/cell_census_builder/tests/test_builder.py
atolopko-czi: that's a good lesson. since not significant for a 4x4 matrix, leaving as is w/a TODO for now.

tools/cell_census_builder/tests/test_builder.py
atolopko-czi: sorry, what is the "the non-uniqueness requirement for feature_reference"?

tools/cell_census_builder/validate.py
atolopko-czi: fixed

tools/cell_census_builder/validate.py
atolopko-czi: fixed

tools/cell_census_builder/validate.py
atolopko-czi: `nnz` now in use! (didn't know about it)

I think I started to add checks for presence shape, and it started to take time, so I punted and left the below TODO

tools/scripts/requirements.txt
atolopko-czi: yup, will upgrade; just wanted to get things working on a11 before introducing any other problems.

tools/cell_census_builder/validate.py
atolopko-czi: I do see `nnz` available anywhere, but we can do this:
```
                assert sum([c.non_zero_length for c in rna['feature_dataset_presence_matrix'].read().csrs()]) > 0
```
in case this saves time instantiating as a pandas df.

tools/cell_census_builder/tests/test_builder.py
bkmartinjr: I was just pointing out that all of the dataframe rows are identical, and that might not be the best test.  This is a nit - just ignore it.

tools/cell_census_builder/__main__.py
atolopko-czi: I refactored to fold together the iteration through the datasets (anndatas), so that the dataset/cell filtering step and axes processing is done with a single, serialized pass.

I'm not sure we can parallelize this though due to the var accumulation.  Even if it can be, let's save that for another PR!

tools/cell_census_builder/__main__.py
atolopko-czi: OMG, ignore this. 😮‍💨  I did read _my_ "comment to reviewers" as feedback.

tools/cell_census_builder/tests/test_builder.py
```diff
@@ -51,17 +52,18 @@ def setup_class(cls) -> None:
     def teardown_class(cls) -> None:
         cls.td.cleanup()
 
-    @dataclass
+    @attrs.define
```
bkmartinjr, line 55: as these are immutable, suggest you use @attrs.frozen

tools/cell_census_builder/validate.py
```diff
@@ -69,12 +70,12 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
...
         assert soma.Collection.exists(census.uri)
-        assert "cxg_schema_version" in census.metadata and census.metadata["cxg_schema_version"] == CXG_SCHEMA_VERSION
-        assert (
-            "census_schema_version" in census.metadata
-            and census.metadata["census_schema_version"] == CENSUS_SCHEMA_VERSION
-        )
-        assert "created_on" in census.metadata and datetime.fromisoformat(census.metadata["created_on"])
+        assert "cxg_schema_version" in census.metadata
+        assert census.metadata["cxg_schema_version"] == CXG_SCHEMA_VERSION
+        assert "census_schema_version" in census.metadata
```
bkmartinjr, line 75: nano-nit:  as noted earlier, this is redundant (ie. the next line will generate a KeyError if not present)

tools/cell_census_builder/validate.py
```diff
@@ -13,10 +13,11 @@
 import pyarrow as pa
 import tiledbsoma as soma
 from scipy import sparse
+from sklearn.linear_model._ridge import _get_rescaled_operator
```
bkmartinjr, line 16: unused - remove?

bkmartinjr: how did this pass the linter?

tools/cell_census_builder/validate.py
```diff
@@ -361,73 +364,81 @@ def validate_X_layers(
...
-        for eb in experiment_builders:
-            se = census_data[eb.name]
-            assert soma.Collection.exists(se.ms["RNA"].X.uri)
-
-            census_obs_df = se.obs.read(column_names=["soma_joinid"]).concat().to_pandas()
-            n_obs = len(census_obs_df)
-            logging.info(f"uri = {se.obs.uri}, eb.n_obs = {eb.n_obs}; n_obs = {n_obs}")
-            # TODO: Only works when run as builder, not standalone validator, since ExpBuilder is not fully initialized
-            #  in the latter case
+    for eb in reopen_experiment_builders(experiment_builders, 'r'):
```
bkmartinjr, line 367: the need to reopen concerns me.  It implies that there is insufficient isolation between the census "build" code and the "validate" code.

I would have expected that the validation pass would work off a clean, newly opened, for read, set of experiment handles.  And therefore would have no possible SOMA shared state, and therefore no need to re-open.

Anything where there is shared SOMA state, eg., open handles, seems like it is doing to potentially be a suspect validation.

Note: I think it is _fine_ to utilize the parts of the ExperimentBuilder that is declarative input to the builder (ie., the name, filter spec, gene_feature stuff, etc).  But the reopen suggests much more is shared, which makes me not trust the validation.

tools/cell_census_builder/__main__.py
```diff
@@ -5,15 +5,16 @@
 import os.path
 import sys
 from datetime import datetime, timezone
-from typing import List
+from typing import List, Union
 
 import pyarrow as pa
 import tiledbsoma as soma
+from anndata import AnnData
```
bkmartinjr, line 12: unused.  why import?  also, how did this pass lint?

tools/cell_census_builder/__main__.py
```diff
@@ -5,15 +5,16 @@
 import os.path
 import sys
 from datetime import datetime, timezone
-from typing import List
+from typing import List, Union
```
bkmartinjr, line 8: Union is also unused - should not be passing lint!

tools/cell_census_builder/validate.py
bkmartinjr: Thinking about this some more -- what has happened (I started it, you added to it) is that the ExperimentBuilder has grown to be two things:
* declaration of the Experiment "definition" -- filters, names, gene_length, etc.
* accumulated state used to "build" the experiment.  eg., var accumulator, n_obs, etc.

The build phase needs both.  The validator should only use the first item, or it isn't really validating in an isolated manner.

Stuff like use of `reopen...` in the validator is a sign that accumulated build state may be creeping into the validator.  Not ideal IMHO.

Not sure it should hold up this PR, but I think a better split here would be a good idea.  Concretely, we could solve this by splitting the ExperimentBuilder object into:
* ExperimentDefinition - containing name, filter, gene_length stuff, etc. 
* ExperimentBuilder - subclass of ExperimentDefinition, adding the build-time accumulators for of obs/var, open soma.Experiment handles, etc.

the validator would be given only the first (definition).

tools/cell_census_builder/__main__.py
```diff
@@ -210,78 +200,133 @@ def build_step1_get_source_assets(args: argparse.Namespace, assets_path: str) ->
...
+
+        # dataset passes filter if either experiment includes cells from the dataset
         if dataset_total_cell_count > 0:
             filtered_datasets.append(dataset)
+            dataset.dataset_total_cell_count = dataset_total_cell_count
+
+    for eb in experiment_builders:
+        logging.info(f"Experiment {eb.name} will contain {eb.n_obs} cells from {eb.n_datasets} datasets")
+
+
```
bkmartinjr, line 233: formatting issues here (too many lines).

I'm thinking the CI linter is not working for the builder, or it would have caught this....

tools/cell_census_builder/__main__.py
bkmartinjr: OK, new changes to the build phases look good to me. And you are right about `var` accumulation - I "spaced" on that not-so-minor detail....

tools/cell_census_builder/__main__.py
atolopko-czi: I hadn't run pre-commit.  Finally did `pre-commit install`, so this shouldn't happen again.

tools/cell_census_builder/__main__.py
atolopko-czi: fixed

tools/cell_census_builder/__main__.py
atolopko-czi: We don't need no stinkin' linter--we've got Bruce! ;-) I think all is well wrt the linter checks. Linter not running due to conflict w/main.

tools/cell_census_builder/__main__.py
bkmartinjr: that explains why it was "OK" on your own machine.  But how did it pass CI when you pushed the commits?

tools/cell_census_builder/__main__.py
bkmartinjr: doh!  of course.  thank you...

tools/cell_census_builder/validate.py
atolopko-czi: all cleaned up

tools/cell_census_builder/tests/test_builder.py
atolopko-czi: done

tools/cell_census_builder/validate.py
atolopko-czi: understood and agreed; this the root problem of https://github.com/chanzuckerberg/cell-census/issues/179.  linked back to this thread from the issue, so we can address all this properly on that issue.

bkmartinjr: The other place we lack sufficient isolation is the re-use of the SOMA/TileDB context.  The validator should have its own, freshly created context.

bkmartinjr: filed as #215

atolopko-czi:
Reviewed @bkmartinjr changes for hang/oom fix et.al. All LGTM. Thank you for the fixes!

bkmartinjr (approved):

bkmartinjr:
Note: currently blocked by single-cell-data/TileDB-SOMA#988

Bento007 (approved):

PR #199, opened 2023-02-17
Docsite

ebezzi:
Adds a docsite for cell-census, using Sphinx.

Notes: 
1. Right now, the Sphinx conf and sources are in the `doc` folder and the artifacts are in `docs`. For GitHub Pages, there needs to be an `index.html` in `docs` so that folder is best reserved for it. Maybe renaming `doc` to `docs-sphinx` could be a good idea.
2. I will add a script that can be used to regenerate the docs. This should be automatic, but I wouldn't put too much effort until we finalize the hosting.


Bento007 (commented):
Do we need to commit all of the generated files?

doc/python-api.rst
```diff
@@ -0,0 +1,24 @@
...
+.. autofunction:: cell_census.download_source_h5ad
+
+Presence Matrix
+----
+.. autofunction:: cell_census.get_presence_matrix
+
+Versioning
+----
+.. autofunction:: cell_census.get_census_version_description
+.. autofunction:: cell_census.get_census_version_directory
```
pablo-gar, line 24:
```suggestion
Open/retrieve Cell Census data
----
.. autofunction:: cell_census.open_soma
.. autofunction:: cell_census.get_source_h5ad_uri
.. autofunction:: cell_census.download_source_h5ad

Get slice as AnnData
----
.. autofunction:: cell_census.get_anndata

Feature presence matrix
----
.. autofunction:: cell_census.get_presence_matrix

Versioning of Cell Census builds
----
.. autofunction:: cell_census.get_census_version_description
.. autofunction:: cell_census.get_census_version_directory
```

ebezzi:
> Do we need to commit all of the generated files?

Right now we're publishing the docs using GitHub pages on a branch, which requires the files to be in the repo. GitHub also offers another [alternate workflow](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow) that, I believe, does not require the files to be checked in, but since this is just a temporary solution, I went for the simplest one. If we decide to keep GitHub Pages as the hosting solution, I think it's worth exploring.

pablo-gar (approved):
LGTM!

PR #203, opened 2023-02-17
Allow default context to be use in open_soma

Bento007:


api/python/cell_census/src/cell_census/open.py
```diff
@@ -17,11 +17,11 @@
 
 def _open_soma(locator: CensusLocator, context: Optional[soma.options.SOMATileDBContext] = None) -> soma.Collection:
     """Private. Merge config defaults and return open census as a soma Collection/context."""
-    context = context or soma.options.SOMATileDBContext()
     s3_region = locator.get("s3_region")
 
     if not context:
         # if no user-defined context, cell_census defaults take precedence over SOMA defaults
+        context = soma.options.SOMATileDBContext()
```
atolopko-czi, line 24: I think we want this instead (untested):
```
        tiledb_config = {**DEFAULT_TILEDB_CONFIGURATION}
        if s3_region is not None:
            tiledb_config["vfs.s3.region"] = s3_region
        context = soma.options.SOMATileDBContext(tiledb_ctx=tiledb.Ctx(tilecb_config))
```

SOMATileDBContext() is supposed to be frozen, so I wouldn't expect the `replace()` to work.

Also `SOMATileDBContext` takes a tiledb.Ctx object, not a plain dict.

Whatever we end up with, I would verify the default tiledb config args are showing when

atolopko-czi (changes requested):
The dead code path fix here is correct. I'm not sure the implementation is quite right. Suggestion added as comment.

api/python/cell_census/src/cell_census/open.py
Bento007: I added a check for the tileDB defaults and they were set. The code also didn't complain about using replace.

atolopko-czi: Thanks for confirming. I would still consider the revised code above since it's a smidge clearer.

atolopko-czi (approved):
Thanks for catching this.

api/python/cell_census/src/cell_census/open.py
Bento007: `tiledb` is an indirect dependency for the cell_census API. Adding `tiledb.Ctx` would require me to add it as a package requirement. Is that something we want?

https://github.com/chanzuckerberg/cell-census/blob/4485b64579b141b1b94e9005cb9f5fb5e26f638e/api/python/cell_census/pyproject.toml#L30-L42

Bento007: I don't see any other instance in the API where we are importing from the tiledb library directly

bkmartinjr (approved):
good catch!  thanks for fixing.

Please ensure it passes CI before you land it.

Issue #11, opened 2022-11-29
improve validation phase of census builder


bkmartinjr:
The "X" layer validation could be enhanced.  It currently performs checks on shape & nnz, but would benefit from content (value) checks as well.

**[Pablo @pablo-gar] proposal, the following validates the integrity of the data contents using 2 reference datasets ** 

  ~~For 2 Tabula Sapiens~~ All datasets perform the following validation steps:
  - [x] Count values for all cell metadata variables is equal between Cell Census and the original h5ads.
  - [ ] The sum values of X matrix is equal between Cell Census and the original dataset h5ads
  - [ ] The sum number of present genes in each h5ad dataset must be equal to the row sum of the corresponding datasets in the presence matrix.

Additional physical layer checks:
- [x] Verify that the entire census is a single directory tree (ie one root, with all data below that
- [x] Verify that the TileDB groups have all children stored with relative URIs (ie., the Group [is_relative flag](https://tiledb-inc-tiledb.readthedocs-hosted.com/projects/tiledb-py/en/stable/python-api.html#tiledb.Group.is_relative) is set to True)
- [x] Verify that obs, var and X layers are all fully consolidated & vacuumed (i.e., have a fragment count of 1)

bkmartinjr:
A note on how to determine fragment count for a TileDB array -- use `tiledb.array_fragments`.

So for each SOMA object in question (e.g. experiment.obs), do the following:
> assert len(tiledb.array_fragments(experiment.obs.uri)) == 1

Bento007:
> A note on how to determine fragment count for a TileDB array -- use `tiledb.array_fragments`.
> 
> So for each SOMA object in question (e.g. experiment.obs), do the following:
> 
> > assert len(tiledb.array_fragments(experiment.obs.uri)) == 1

Does need to be strictly == 1 or is >= 1 OK.

bkmartinjr:
> Does it need to be strictly == 1 or is <= 1 OK.

Strictly 1

The point is that we want fragments to be consolidated into a single fragment, i.e., "==1"

bkmartinjr:
> Strictly 1

note on this - I neglected to consider the case where the array was empty.  In that corner case, the fragment count would be zero.  I'll make a fix for this in my RC2 PR.  Apologies for the misdirect!

bkmartinjr:
> Should I take this literally (as in sum all the values) or is there more nuance to this check?

[...]

Issue #108, opened 2023-01-31
title is not optional in dataset schema


brianraymor:
https://github.com/chanzuckerberg/cell-census/blob/7152b68c0b3e0debd1fc2b5200f1131bf26a00a0/tools/cell_census_builder/manifest.py#LL63C11-L63C86

`"dataset_title": dataset.get("title", ""),  # title is optional in schema
`

The comment is incorrect and the default value unnecessary. `title` is required in [uns](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/3.0.0/schema.md#uns-dataset-metadata) because it is included in **Curators MUST annotate the following keys and values in uns.**

pablo-gar:
Done

PR #204, opened 2023-02-17
Additional validation checks

Bento007:
- Verify that the TileDB groups have all children stored with relative URIs
- Validate required dataset metadata is present
- Verify that obs, var and X layers are all fully consolidated & vacuumed
- Add consolidate to builder unit test.
- split consolidate function to make testing easier.

Relates to #11


tools/cell_census_builder/consolidate.py
```diff
@@ -17,12 +17,20 @@ def consolidate(args: argparse.Namespace, uri: str) -> None:
         return
 
     logging.info("Consolidate: started")
+    uris_to_consolidate = gather(uri)
+    run(args, uris_to_consolidate)
+    logging.info("Consolidate: finished")
+
 
+def gather(uri: str) -> List[str]:
```
bkmartinjr, line 25: if these are module-private functions, please add an underscore prefix (e.g., _gather).  If public functions, please give them a more descriptive name.

Ditto for run/_run

tools/cell_census_builder/validate.py
```diff
@@ -74,58 +78,85 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
         assert datetime.fromisoformat(census.metadata["created_on"])
         assert "git_commit_sha" in census.metadata
 
-        for name in [CENSUS_INFO_NAME, CENSUS_DATA_NAME]:
-            assert soma.Collection.exists(census[name].uri)
-            assert census[name].soma_type == "SOMACollection"
+        with tiledb.Group(census.uri) as census_group:
```
bkmartinjr, line 81: Using `tiledb` directly is fine for now, but it would be better if we used the SOMA API for the `is_relative` test.  This points out a gap in in the SOMA API!

Please:  a) file a feature request for the SOMA equivalent of `is_relative`,    b) add a TODO/reminder in this code to port to that new API when available.

tools/cell_census_builder/validate.py
```diff
@@ -74,58 +78,85 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
...
+            for name in [CENSUS_INFO_NAME, CENSUS_DATA_NAME]:
+                assert soma.Collection.exists(census[name].uri)
+                assert census[name].soma_type == "SOMACollection"
```
bkmartinjr, line 84: this check is redundant and can be removed.  The `soma.Collection.exists()` peforms the same test....

tools/cell_census_builder/validate.py
```diff
@@ -74,58 +78,85 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
...
         census_info = census[CENSUS_INFO_NAME]
-        for name in [CENSUS_DATASETS_NAME, CENSUS_SUMMARY_NAME, CENSUS_SUMMARY_CELL_COUNTS_NAME]:
-            assert name in census_info, f"`{name}` missing from census_info"
-            assert soma.DataFrame.exists(census_info[name].uri)
-            assert census_info[name].soma_type == "SOMADataFrame"
+        with tiledb.Group(census_info.uri) as census_info_group:
+            for name in [CENSUS_DATASETS_NAME, CENSUS_SUMMARY_NAME, CENSUS_SUMMARY_CELL_COUNTS_NAME]:
+                assert name in census_info, f"`{name}` missing from census_info"
+                assert soma.DataFrame.exists(census_info[name].uri)
+                assert census_info[name].soma_type == "SOMADataFrame"
```
bkmartinjr, line 92: redundant - this test is already performed by the `soma.DataFrame.exists()` call

tools/cell_census_builder/validate.py
bkmartinjr: ditto for the same code below.

tools/cell_census_builder/validate.py
```diff
@@ -74,58 +78,85 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
...
+                assert census_info_group.is_relative(name), error_tiledb_group
 
         assert sorted(census_info[CENSUS_DATASETS_NAME].keys()) == sorted(CENSUS_DATASETS_COLUMNS + ["soma_joinid"])
         assert sorted(census_info[CENSUS_SUMMARY_CELL_COUNTS_NAME].keys()) == sorted(
             list(CENSUS_SUMMARY_CELL_COUNTS_COLUMNS) + ["soma_joinid"]
         )
         assert sorted(census_info[CENSUS_SUMMARY_NAME].keys()) == sorted(["label", "value", "soma_joinid"])
 
+        # verify required dataset fields are set
+        census_info[CENSUS_DATASETS_NAME]
```
bkmartinjr, line 102: not sure what this line is doing - delete?

Bento007: yep, probably missed when resolving a conflict.

tools/cell_census_builder/validate.py
```diff
@@ -74,58 +78,85 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
...
+        df: pd.DataFrame = census_info[CENSUS_DATASETS_NAME].read().concat().to_pandas()
+        assert not all(df["collection_id"].isin([""]))
```
bkmartinjr, line 104: I personally found this test a bit confusing. Totally optional, but do you think it would be clearer as:

```python
    (df["collection_id"] != "").all()
```

bkmartinjr: alternatively, you could do all four tests together as:
```python
(df[["collection_id","collection_doi","collection_name","dataset_title"]] != "").all().all()
```

tools/cell_census_builder/validate.py
Bento007: do I file it in the [TileDB-SOMA](https://github.com/single-cell-data/TileDB-SOMA) repo?

tools/cell_census_builder/validate.py
```diff
@@ -74,58 +78,85 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
...
+        assert not all(df["collection_name"].isin([""]))
+        assert not all(df["collection_doi"].isin([""]))
```
bkmartinjr, line 106: Last I checked, DOI is not required.  would be good to confirm all of these fields are mandatory.  

The current census build (for example) definitely has several datasets with no DOI

tools/cell_census_builder/validate.py
bkmartinjr: this pattern is repeated below - good to clean up all of them.

tools/cell_census_builder/validate.py
```diff
@@ -74,58 +78,85 @@ def validate_all_soma_objects_exist(soma_path: str, experiment_builders: List[Ex
...
-                # TODO(atolopko): validate 1) shape, 2) joinids exist in datsets and var
-
+        with tiledb.Group(census_data.uri) as census_data_group:
+            for eb in experiment_builders:
+                assert soma.Experiment.exists(census_data[eb.name].uri)
+                assert census_data[eb.name].soma_type == "SOMAExperiment"
+                assert census_data_group.is_relative(eb.name), error_tiledb_group
+
+                e = census_data[eb.name]
+                with tiledb.Group(e.uri) as e_group:
```
bkmartinjr, line 118: I think this code might be cleaner if you factored out the relative check into a stand-alone function, which would isolate all of the TileDB-specific code, and partition the "concern" being tested.  For example, reading this, it is a bit hard to figure out if you are testing the `relative` attribute on all objects, or have missed a few.

you could simply do a recursive, top-down walk of the root Collection, and check every child below the root for this flag.  Much less code, and (I think) easier to understand.

tools/cell_census_builder/validate.py
```diff
@@ -447,6 +478,44 @@ def validate_manifest_contents(assets_path: str, datasets: List[Dataset]) -> boo
     return True
 
 
+def validate_consolidation(soma_path: str, experiment_builders: List[ExperimentBuilder]) -> bool:
+    """Verify that obs, var and X layers are all fully consolidated & vacuumed"""
```
bkmartinjr, line 482: this code should walk all elements the collections, not just the pre-defined names, and check that _any_ DataFrame/NDArray is consolidated.

In other words, do something like:
```python
for name in collection:
    if isinstance(collection[name], (soma.DataFrame, soma.SparseNDArray, soma.DenseNDArray):
        # check fragment count
    else:
        # recursively call on collection[name]
```

In other words, check everything, and don't hardwire in magic names as we will change them over time.

bkmartinjr (changes requested):
Some suggestions on how the code could be simplified and made robust to future name changes.

tools/cell_census_builder/validate.py
bkmartinjr: Yep, that would be perfect. It is a case where we just forgot to add the ability to introspect this property.  Nice dogfooding find!!!

tools/cell_census_builder/validate.py
brianraymor: DOI is optional
Dataset title - see #108

tools/cell_census_builder/validate.py
Bento007: I think the first suggestion is cleaner.

tools/cell_census_builder/validate.py
Bento007: https://github.com/single-cell-data/TileDB-SOMA/issues/999

tools/cell_census_builder/validate.py
```diff
@@ -447,6 +445,45 @@ def validate_manifest_contents(assets_path: str, datasets: List[Dataset]) -> boo
...
+def validate_relative_path(soma_path: str) -> bool:
+    """
+    Verify the census objects are stored in the same relative path
+    :param soma_path:
+    :return:
+    """
+    # TODO use SOMA API. See https://github.com/single-cell-data/TileDB-SOMA/issues/999
+
+    def _walk_tree(name: str, parent: Any) -> None:
+        if parent.soma_type in ["SOMACollection", "SOMAExperiment", "SOMAMeasurement"]:
```
bkmartinjr, line 474: There is absolutely _nothing_ wrong with this code, but I have small suggestion to make it a bit more elegant/concise.  Completely optional!

You can do the same by using the Python type system and more concisely, written as:

> if isinstance(parent, soma.Collection):

This is because Experiment & Measurement are Collection sub-classes in the Python typing.

bkmartinjr (approved):
Nice job - thanks for doing this!

PR #206, opened 2023-02-19
[R] CellCensus package MVP

mlin:
Provides `CellCensus::open_soma(census_version='latest')` to get the top-level `tiledbsoma::SOMACollection`, as well as underlying helper methods for loading the release directory JSON.

- [Initial README](https://github.com/chanzuckerberg/cell-census/tree/mlin/bootstrap-r-pkg/api/r/CellCensus#readme) guidance is to install the package directly from GitHub for now.
- Initially depends on bleeding-edge builds of [tiledbsoma from r-universe](https://tiledb-inc.r-universe.dev/tiledbsoma) per @eddelbuettel suggestion.
- Packaging strategy follows the [R Packages book](https://r-pkgs.org/). 
- Includes GH CI workflow using some [nice R Actions](https://github.com/r-lib/actions).
- Some early roxygen docstrings, mostly copy pasta from the Python equivalents.


.github/workflows/Rcheck.yml
```diff
@@ -0,0 +1,28 @@
+name: cell_census R package checks
```
bkmartinjr, line 1: now that we are multi-lingual, we may want to reorganize the other workflows (or at least rename them so it is clear they are Python-specific).

*Optional* idea: rename each with a language prefix?   

@atolopko-czi - any thoughts or preferences on this?  

I'm also OK if we defer this to a future PR/project.

bkmartinjr:
@mlin @pablo-gar - how will demo/doc notebooks be organized now that we are multi-lingual?  Will they still exist in language-specific sub-folders alongside the cell-census package, or be promoted into another location?

I ask because this decision may effect how the contents of api/{r,python}/* are organized.

api/r/CellCensus/CellCensus.Rproj
```diff
@@ -0,0 +1,22 @@
+Version: 1.0
```
bkmartinjr, line 1: I don't pretend to know R ecosystem versioning conventions, but should an early release such as this start as a 1.0?  Or 0.?

api/r/CellCensus/tests/testthat/test-open.R
```diff
@@ -0,0 +1,12 @@
+test_that("open_soma", {
+  coll <- open_soma("2023-02-13")
```
bkmartinjr, line 2: this build tag will eventually go away (weeks to months from now). The only durable tag (currently) is the `latest` tag.

bkmartinjr (approved):
A couple of minor nits (most significant is the version number - do we want to use `1.0` at this early date?)

But I think it is completely sufficient as a first bootstrap.  Thank you!

.github/workflows/Rcheck.yml
```diff
@@ -0,0 +1,28 @@
...
+
+on:
+  pull_request:
+    paths-ignore:
+      - "apis/python/**"
```
ebezzi, line 6: We can probably add the builder here: `tools/cell_census_builder`

api/r/CellCensus/CellCensus.Rproj
ebezzi: Also should we sync this to the Python versioning? We should probably consolidate the behavior with tiledb-soma anyway.

.github/workflows/Rcheck.yml
atolopko-czi: either language-specific prefixes or subdirs (if that's supported) seems reasonable to me

ebezzi (approved):
Looks great! Added a couple of nitpicks but they can be addressed at a later time (if needed).

api/r/CellCensus/tests/testthat/test-open.R
atolopko-czi: IMO, if tests depend upon having _specific_ data, those tests should use a test fixture (dynamically built, ideally) rather than the live census.  And we should use `latest` if it's the latter case.

pablo-gar:
> @mlin @pablo-gar - how will demo/doc notebooks be organized now that we are multi-lingual? Will they still exist in language-specific sub-folders alongside the cell-census package, or be promoted into another location?
> 
> I ask because this decision may effect how the contents of api/{r,python}/* are organized.

@bkmartinjr The main location for users to access the notebooks should be in the doc-site. As such the python notebooks can stay as they are, and the R tutorials should live in the vignettes folder of the R package (cc @mlin) -- these get automatically render in the doc-site via `pgkdown` (cc @ebezzi)

api/r/CellCensus/tests/testthat/test-open.R
ebezzi: I think fixtures are ideal but they will probably take quite a bit of time to write and we should probably add them after we ship this MVP. I think using this as a sanity check is a good idea at least for now. Using `latest` might be dangerous since a build could cause tests to start failing in main after a build.

atolopko-czi: Agreed that fixtures are fine for future consideration.

I think it's acceptable if tests fail even if the census data is the cause. These are essentially system tests, rather than unit tests. Consider that many of the Python "unit" tests, which are really system tests, are using pytest markers (annotations) to denote that they depend upon live data.   If we ever have a failure that is census build-specific, it's probably an indication that we're missing an important builder validation. And we can then improve the validator as needed.

.github/workflows/Rcheck.yml
mlin: Will do a small follow-up PR on this.

api/r/CellCensus/CellCensus.Rproj
mlin: This `1.0` is a red herring -- it describes the *format/schema version* of the `.Rproj` file, not of the project/package itself. The package version is written in the `DESCRIPTION` file and defaulted to `0.0.0.9000` (I don't know where the 9000 comes from!). Agree we will want to some some coordinated version numbers, hopefully controlled by git tags, when we're settled enough to make coherent release versions. Looking forward to that =)

api/r/CellCensus/tests/testthat/test-open.R
mlin: Agree all. I changed some things to use `latest` where it didn't really matter, and where needed (e.g. checking the respective row of the release directory dataframe), I changed the several occurrences of `2022-02-13` to refer to a single hardcoded constant so that it'll be easy to update if/when that version goes away.

Issue #208, opened 2023-02-21
Revert builder PIT hack


atolopko-czi:
Revert timestamp hacks that were needed as a workaround for https://github.com/single-cell-data/TileDB-SOMA/issues/919. See TODOs in `cell_census_builder.globals.SOMA_TileDB_Context()`

PR #210, opened 2023-02-21
revert builder pit hack (FAILS)

atolopko-czi:
#208 

Upgraded to TileDB-SOMA `1.0rc0`, which was expected to allow us to specify an explicit write timestamp. Alas, the unit test for the builder continues to fail (it passes without specifying the explicit write timestamp, as before).

Unit test failure stack trace (from unit test report output):
```
test_builder.py:156: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../__main__.py:142: in build
    build_step4_populate_X_layers(assets_path, filtered_datasets, experiment_builders, args)
../__main__.py:295: in build_step4_populate_X_layers
    populate_X_layers(assets_path, filtered_datasets, experiment_builders, args)
../experiment_builder.py:463: in populate_X_layers
    presence += _accumulate_X(assets_path, dataset, experiment_builders, progress=(n, len(datasets)))
../experiment_builder.py:430: in _accumulate_X
    return _accumulate_all_X_layers(
../experiment_builder.py:349: in _accumulate_all_X_layers
    experiment.ms[ms_name].X[layer_name].write(pa.SparseCOOTensor.from_scipy(X_remap))
```


mlin:
@atolopko-czi I can't quite tell from the stack trace -- is this the same nondeterministically inconsistent view of the data reported in https://github.com/single-cell-data/TileDB-SOMA/issues/919 ? or some different pathology?

thetorpedodog:
taking a quick dip into here to see if I can understand how this would interact with the change linked above and:

https://github.com/chanzuckerberg/cell-census/blob/cab8bc053939afaeb91e28647a4003da4f95e2be/tools/cell_census_builder/globals.py#L229

↑ won’t this always be `True`?

mlin:
I made a regression test in https://github.com/single-cell-data/TileDB-SOMA/pull/983 to at least be confident we fixed the specific problem in https://github.com/single-cell-data/TileDB-SOMA/issues/919 (not to exclude that others may exist). Agree with @thetorpedodog that condition looks sus.

bkmartinjr:
@atolopko-czi - I believe this is obsolete. Can you close if that is true?

atolopko-czi:
Obsoleted by later PRs

PR #211, opened 2023-02-22
docs: Clarifies text for Feature Dataset Presence Matrix, rename schema file

pablo-gar:
Changes:

- Clarifies 'Feature Dataset Presence Matrix' specification, it was recently noted by @atolopko-czi that the spec was ambiguous and text was added to clarify it
- The schema minor version is bumped.
- Removes version number from schema file to reduce maintenance cost of other docs referring to it.


bkmartinjr (approved):
LGTM.

PR #217, opened 2023-02-23
Add SOMA page and links

ebezzi:
Add a `What is SOMA` (title TBD) page with description and links to SOMA. Links are temporary and will change in the future.

I also added `intersphinx` so that links to external documentation can be set. Right now, the `tiledbsoma` documentation isn't up to date so they won't work, but they will in the future.


pablo-gar (approved):
LGTM

PR #227, opened 2023-02-27
Update python package and builder to tiledbsoma 1.0.0rc2

### LINKED ISSUE #208: Revert builder PIT hack

### LINKED ISSUE #221: Cell Census builder uses TileDB-SOMA 1.0 RC1
There a few bug fixes in RC1 of TileDB-SOMA, when the release is available the builder need to use it

### LINKED ISSUE #228: update to tiledbsoma RC2
Update builder and python package

bkmartinjr:
Update cell census package *and* builder to utilize `tiledbsoma==1.0.0rc2`

Other changes:
* revert the census builder to create the census with default TileDB timestamp behavior.  Specifically, the objects comprising the census build will not share a single timestamp.
* fix small bug in validator which is triggered by test builds containing zero cells for a given organism.

Fixes #208 
Fixes #221 
Fixes #228


tools/cell_census_builder/globals.py
```diff
@@ -220,32 +228,22 @@
...
-# The logical timestamp at which all builder data should be recorded
-WRITE_TIMESTAMP = int(time.time() * 1000)
-
-# Using "end of time" for read_timestamp means that all writes are visible, no matter what write timestamp was used
-END_OF_TIME = 0xFFFFFFFFFFFFFFFF
-
 
 def SOMA_TileDB_Context() -> soma.options.SOMATileDBContext:
     global _SOMA_TileDB_Context
     if _SOMA_TileDB_Context is None or _SOMA_TileDB_Context != TileDB_Ctx():
```
atolopko-czi, line 234: To fix a bug I introduced on an earlier PR:
```suggestion
    if _SOMA_TileDB_Context is None or _SOMA_TileDB_Context.tiledb_ctx != TileDB_Ctx():
```

tools/cell_census_builder/globals.py
```diff
@@ -220,32 +228,22 @@
...
         _SOMA_TileDB_Context = soma.options.SOMATileDBContext(
             tiledb_ctx=TileDB_Ctx(),
-            # TODO: Setting an explicit write timestamp causes later reads to fail!
-            # write_timestamp=write_timestamp,
-            # TODO: We *should* be able to set this equal to WRITE_TIMESTAMP, but as specifying a write_timestamp is
-            #  problematic, we must use "end of time" for now
-            read_timestamp=END_OF_TIME,
+            # `None` opts out of explicit timestamp consistency, and uses the underlying
+            # TileDB per-object write consistency. This is sufficient given that this
+            # builder explicitly orders / synchronizes all reading/writing.
```
atolopko-czi, line 239: As "timestamp consistency [is] the default" (per [this comment](https://github.com/single-cell-data/TileDB-SOMA/pull/981#issuecomment-1442438931)), we could remove this comment. Minimally, update it to be accurate or reword so that it doesn't become outdated if/when TileDB-SOMA semantics change again.

tools/cell_census_builder/globals.py
```diff
@@ -220,32 +228,22 @@
...
+            timestamp=None,
         )
     return _SOMA_TileDB_Context
 
 
 def TileDB_Ctx() -> tiledb.Ctx:
+    # See `DEFAULT_TILEDB_CONFIG` above
```
atolopko-czi, line 246: It _could_ be set to something other than `DEFAULT_TILEDB_CONFIG`. But if we're never going to do that in builder code, I suppose we could remove the ability to set it altogether, in which case this comment is fine.

tools/cell_census_builder/globals.py
atolopko-czi: But maybe it's time to just eliminate support for custom `tiledb_ctx` values (remove the setter), since we don't use it, allowing us to remove this check.

atolopko-czi (approved):
LGTM. Some comments re: comments, one unrelated bug fix, and a suggestion to remove support for custom setting of `tiledb_ctx`.

tools/cell_census_builder/globals.py
bkmartinjr: Yeah, for now, I'll just make it a cached value.  Good call.

tools/cell_census_builder/globals.py
bkmartinjr: you are right that this is confusing.  I meant not using PIT consistency across the entire census collection.  I'll delete.

tools/cell_census_builder/globals.py
bkmartinjr: resolved by deleting setter

PR #232, opened 2023-02-28
rename CI workflows

mlin:
More-clearly separate Python and R workflows.

Just renames the .yml files with py- and r- prefixes. I tried making `.github/workflows/{python,r}` subdirectories, but unfortunately this didn't work, GitHub seemed to ignore them.


.github/workflows/py-build.yml
```diff
@@ -1,4 +1,4 @@
-name: cell_census build
+name: Python cell_census build
```
bkmartinjr, line 1: this file name is referenced in `release_process.md` - could you update the reference to point to the updated file name as part of this PR?

atolopko-czi (approved):

PR #233, opened 2023-03-01
Change docstrings to use Google conventions

ebezzi:
Notes: 
1. The docstrings at the beginning of each module could use some work. That said, the modules generally only contain 1-2 functions, and the examples are defined in the functions.
2. I removed type annotations from `Args`. They are only recommended if the functions lack type annotations, which is not our case. I think removing them helps with the readability, but we can add them back.

### DIFF
--- a/api/python/cell_census/src/cell_census/_experiment.py
+++ b/api/python/cell_census/src/cell_census/_experiment.py
@@ -21,14 +21,16 @@ def _get_experiment(census: soma.Collection, organism: str) -> soma.Experiment:
     [lifecycle: experimental]
 
     Args:
-        census: soma.Collection, The census
-        organism: str, The organism name, eg., ``Homo sapiens``
+        census: soma.Collection
+            The census.
+        organism: str
+            The organism name, eg., ``Homo sapiens``.
 
     Returns:
         A soma.Experiment object with the requested experiment.
 
     Raises:
-        ValueError: if unable to find the specified organism
+        ValueError: if unable to find the specified organism.
 
     Examples:
 
--- a/api/python/cell_census/src/cell_census/_experiment.py
+++ b/api/python/cell_census/src/cell_census/_experiment.py
@@ -1,33 +1,42 @@
+# Copyright (c) 2022-2023 Chan Zuckerberg Initiative
+#
+# Licensed under the MIT License.
+
+"""
+Experiments handler.
+
+Contains methods to retrieve SOMA Experiments.
+"""
+
 import re
 
 import tiledbsoma as soma
 
 
 def _get_experiment(census: soma.Collection, organism: str) -> soma.Experiment:
-    """
-    Given a census ``soma.Collection``, return the experiment for the named organism.
+    """Given a census ``soma.Collection``, return the experiment for the named organism.
     Organism matching is somewhat flexible, attempting to map from human-friendly
-    names to the underlying collection element name.  Will raise a ``ValueError`` if
-    unable to find the specified organism [lifecycle: experimental].
+    names to the underlying collection element name.
 
-    Parameters
-    ----------
-    census - ``soma.Collection``
-        The census
-    organism - ``str``
-        The organism name, eg., ``Homo sapiens``
+    [lifecycle: experimental]
 
-    Returns
-    -------
-    ``soma.Experiment`` - the requested experiment.
+    Args:
+        census: soma.Collection, The census
+        organism: str, The organism name, eg., ``Homo sapiens``
 
-    Examples
-    --------
-    >>> human = get_experiment(census, 'homo sapiens')
+    Returns:
+        A soma.Experiment object with the requested experiment.
 
-    >>> human = get_experiment(census, 'Homo sapiens')
+    Raises:
+        ValueError: if unable to find the specified organism
 
-    >>> human = get_experiment(census, 'homo_sapiens')
+    Examples:
+
+        >>> human = get_experiment(census, 'homo sapiens')
+
+        >>> human = get_experiment(census, 'Homo sapiens')
+
+        >>> human = get_experiment(census, 'homo_sapiens')
     """
     # lower/snake case the organism name to find the experiment name
     exp_name = re.sub(r"[ ]+", "_", organism).lower()
--- a/api/python/cell_census/src/cell_census/_get_anndata.py
+++ b/api/python/cell_census/src/cell_census/_get_anndata.py
@@ -1,3 +1,11 @@
+# Copyright (c) 2022-2023 Chan Zuckerberg Initiative
+#
+# Licensed under the MIT License.
+
+"""Get slice as AnnData
+
+Methods to retrieve slices of the census as AnnData objects.
+"""
 from typing import Optional
 
 import anndata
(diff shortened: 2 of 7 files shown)


api/python/cell_census/src/cell_census/_get_anndata.py
```diff
@@ -25,45 +33,32 @@ def get_anndata(
...
+        X_name: The ``X`` layer to query. Defaults to `raw`.
+        obs_value_filter: Value filter for the ``obs`` metadata. Value is a filter query written in the
+            SOMA ``value_filter`` syntax.
+        obs_coords: Coordinates for the ``obs`` axis, which is indexed by the ``soma_joinid`` value.
+            May be an ``int``, a list of ``int``, or a slice. The default, ``None``, selects all.
+        var_value_filter: Value filter for the ``var`` metadata. Value is a filter query written in the
+            SOMA ``value_filter`` syntax.
+        var_coords: Coordinates for the ``var`` axis, which is indexed by the ``soma_joinid`` value.
+            May be an ``int``, a list of ``int``, or a slice. The default, ``None``, selects all.
+        column_names: Colums to fetch for ``obs`` and ``var`` dataframes.
```
bkmartinjr, line 53: Columns is misspelled ("Colums")

api/python/cell_census/src/cell_census/_experiment.py
```diff
@@ -1,33 +1,42 @@
...
+    """Given a census ``soma.Collection``, return the experiment for the named organism.
     Organism matching is somewhat flexible, attempting to map from human-friendly
-    names to the underlying collection element name.  Will raise a ``ValueError`` if
-    unable to find the specified organism [lifecycle: experimental].
+    names to the underlying collection element name.
+
+    [lifecycle: experimental]
+
+    Args:
+        census: soma.Collection, The census
```
bkmartinjr, line 24: Suggestion only - I think the args with a linebreak reads better in pydoc and in an IDE.  It is explicitly allowed per the Google style guide - the linebreak goes just before the description of the arg.

In other words:

```
    Args:
        census: soma.Collection
            The census
        organism: str
            The organism name, eg., ``Homo sapiens``
```

api/python/cell_census/src/cell_census/_experiment.py
```diff
@@ -1,33 +1,42 @@
...
 
-    Returns
-    -------
-    ``soma.Experiment`` - the requested experiment.
+    Examples:
 
-    Examples
-    --------
-    >>> human = get_experiment(census, 'homo sapiens')
+        >>> human = get_experiment(census, 'homo sapiens')
```
bkmartinjr, line 36: some examples (e.g., this file) have an empty line between them, some (e.g., get_anndata) do not.  I like the extra whitespace, but please use your own discretion about which we should adopt.

api/python/cell_census/src/cell_census/_open.py
```diff
@@ -103,27 +110,25 @@ def open_soma(
...
+    does not guarantee that the H5AD exists or is accessible to the user.
+
+    [lifecycle: experimental]
+
+    Args:
+        dataset_id: The ``dataset_id`` of interest.
+        census_version: The census version.
+
+    Returns:
+        A :class:`CensusLocator` object that contains the URI and optional S3 region for the source H5AD.
```
bkmartinjr, line 123: I noticed some of these have the tag `:py:class:` and some have `:class:` - are both correct?

api/python/cell_census/src/cell_census/_open.py
```diff
@@ -42,54 +51,52 @@ def open_soma(
...
 
-    Open a specific Cell Census by version:
+    Returns:
+        A SOMA Collection object containing the top-level census.
+        It can be used as a context manager, which will automatically close upon exit.
 
-    >>> with cell_census.open_soma("2022-12-31") as census:
-            ...
+    Raises:
+        ValueError: if the census cannot be found, the URI cannot be opened, or neither a URI
```
bkmartinjr, line 69: Should the `ValueError` have a `:class:` tag?

bkmartinjr (approved):
overall, looks great!  Couple of minor suggestions inline.

Approving on the expectation that those are easy to fix.

bkmartinjr:
@ebezzi - oh, this also needs to be clean on the CI checks.  Apologies, I missed that before I hit approved!

api/python/cell_census/src/cell_census/_experiment.py
ebezzi: I think it's a good idea, especially since some of the later methods have a lot of arguments. I will change it.

api/python/cell_census/src/cell_census/_open.py
ebezzi: They're both correct - I used `:py:class:` for external docs since that's the prefix exported by Sphinx. But since both work, I will settle on `:class:`.

api/python/cell_census/src/cell_census/_open.py
ebezzi: Not necessary - Sphinx can parse this and add the hyperlink without additional annotations.

api/python/cell_census/src/cell_census/_open.py
```diff
@@ -1,3 +1,12 @@
+# Copyright (c) 2022-2023 Chan Zuckerberg Initiative
+#
+# Licensed under the MIT License.
+
+"""Open census and related datasets
+
+Contains methods to open the Cell Census object and access its datasets.
```
pablo-gar, line 7:
```suggestion
Contains methods to open  publicly hosted versions of Cell Census object and access its source datasets.
```

api/python/cell_census/src/cell_census/_presence_matrix.py
```diff
@@ -1,3 +1,12 @@
+# Copyright (c) 2022-2023 Chan Zuckerberg Initiative
+#
+# Licensed under the MIT License.
+
+"""Presence matrix methods
+
+Methods to retrieve the feature presence matrix.
```
pablo-gar, line 7:
```suggestion
Methods to retrieve the feature dataset presence matrix.
```

api/python/cell_census/src/cell_census/_presence_matrix.py
```diff
@@ -9,31 +18,31 @@ def get_presence_matrix(
     organism: str,
     measurement_name: str = "RNA",
 ) -> sparse.csr_matrix:
-    """
-    Read the gene presence matrix and return as a SciPy sparse CSR array
-    (``scipy.sparse.csr_array``). The returned sparse matrix is indexed on the
+    """Read the gene presence matrix and return as a SciPy sparse CSR array.
```
pablo-gar, line 21:
```suggestion
    """Read the feature dataset presence matrix and return as a SciPy sparse CSR array.
```

api/python/cell_census/src/cell_census/_release_directory.py
```diff
@@ -1,3 +1,12 @@
+# Copyright (c) 2022-2023 Chan Zuckerberg Initiative
+#
+# Licensed under the MIT License.
+
+"""Versioning of Cell Census builds
+
+Methods to retrieve information about the available Cell Census versions.
```
pablo-gar, line 7:
```suggestion
Methods to retrieve information about versions of the publicly hosted Cell Census object.
```

pablo-gar (approved):

PR #236, opened 2023-03-01
fix presence matrix wrong shape

ebezzi:
Reverts a [change](https://github.com/chanzuckerberg/cell-census/commit/85bf654ebdfae67a64135b086a078b27bf041715#diff-32913cbb027d8f5e52f9ab54e24145dcb765c97d8ad2e6165e45625fade18e39R268) that introduced a bug that led the presence matrix to have an incorrect shape.

Per [the cell-census schema specification](https://github.com/chanzuckerberg/cell-census/blob/main/docs/cell_census_schema.md#feature-dataset-presence-matrix--census_objcensus_dataorganismmsrnafeature_dataset_presence_matrix--somasparsendarray), the presence matrix needs to have N rows where N is the number of datasets belonging to the experiment. Since this is a sparse matrix, its _domain_ will extend to the total number of datasets (i.e. `max(d.soma_joinid for d in datasets)`), therefore the shape of the presence matrix (for each experiment) should equal the total number of datasets.

### DIFF
--- a/tools/cell_census_builder/experiment_builder.py
+++ b/tools/cell_census_builder/experiment_builder.py
@@ -254,7 +254,7 @@ class ExperimentBuilder:
                     platform_config=CENSUS_X_LAYERS_PLATFORM_CONFIG[layer_name],
                 )
 
-    def populate_presence_matrix(self) -> None:
+    def populate_presence_matrix(self, datasets: List[Dataset]) -> None:
         """
         Save presence matrix per Experiment
         """
@@ -265,7 +265,7 @@ class ExperimentBuilder:
             # sanity check
             assert len(self.presence) == self.n_datasets
 
-            max_dataset_joinid = max(self.presence.keys())
+            max_dataset_joinid = max(d.soma_joinid for d in datasets)
 
             # LIL is fast way to create spmatrix
             pm = sparse.lil_array((max_dataset_joinid + 1, self.n_var), dtype=bool)
--- a/tools/cell_census_builder/tests/test_builder.py
+++ b/tools/cell_census_builder/tests/test_builder.py
@@ -82,6 +82,8 @@ def test_base_builder_creation(
             # Presence matrix should exist with the correct dimensions
             for exp_name in ["homo_sapiens", "mus_musculus"]:
                 fdpm = census[CENSUS_DATA_NAME][exp_name].ms[MEASUREMENT_RNA_NAME][FEATURE_DATASET_PRESENCE_MATRIX_NAME]
+                fdpm_matrix = fdpm.read().coos().concat()
+                assert fdpm_matrix.shape[0] == 4  # 4 datasets
                 fdpm_df = fdpm.read().tables().concat().to_pandas()
                 n_datasets = fdpm_df["soma_dim_0"].nunique()
                 n_features = fdpm_df["soma_dim_1"].nunique()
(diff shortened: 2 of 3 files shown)


tools/cell_census_builder/tests/test_builder.py
```diff
@@ -82,6 +82,8 @@ def test_base_builder_creation(
             # Presence matrix should exist with the correct dimensions
             for exp_name in ["homo_sapiens", "mus_musculus"]:
                 fdpm = census[CENSUS_DATA_NAME][exp_name].ms[MEASUREMENT_RNA_NAME][FEATURE_DATASET_PRESENCE_MATRIX_NAME]
+                fdpm_matrix = fdpm.read().coos().concat()
+                assert fdpm_matrix.shape[0] == 4  # 4 datasets
```
atolopko-czi, line 86: Each presence matrix should only have rows for the datasets that the experiment it is associated with. So 2 (distinct) datasets in each presence matrix for this test, 4 features each, 8 rows total.  See [discussion](https://czi-sci.slack.com/archives/C048BN7K6U8/p1676562041122019).  Let me know if I'm misunderstanding.

atolopko-czi (changes requested):
Requesting changes, if only to ensure we're doing the right thing here before approving. My take is that this change is not needed.

tools/cell_census_builder/tests/test_builder.py
bkmartinjr: Yeah, this assert is wrong.  Andrew has it right as I read the conftest code.

Remember, per Pablo's spec, it is the number of datasets that contribute data to the _specific_ experiment, not the number in the total census (i.e., union of all organisms)

tools/cell_census_builder/experiment_builder.py
```diff
@@ -265,7 +265,7 @@ def populate_presence_matrix(self) -> None:
             # sanity check
             assert len(self.presence) == self.n_datasets
 
-            max_dataset_joinid = max(self.presence.keys())
+            max_dataset_joinid = max(d.soma_joinid for d in datasets)
```
bkmartinjr, line 268: this is wrong - it is going to add datasets which do not exist in the organism's experiment.

tools/cell_census_builder/tests/test_builder.py
ebezzi: As mentioned in the (edited) description, this wasn't the previous behavior which is what this PR is trying to restore. Since that is not compliant with the SOMA abstract spec, we'll probably want to close this and work towards a more permanent fix.

bkmartinjr: Yeah, I was misunderstanding the issue at hand. 

Here are some suggested constraints to assert - ie. what we would expect to be true:

1. For all experiment's presence matrix, all dimension 0 coordinates (aka the dataset DataFrame's soma_joinid) should be > 0 and <= the max soma_joinid in the datasets table.
2. for all experiment's presence matrix, the shape[0] == max(datasets soma_joinid) + 1 

Important assumption above:  the rows in the datasets DataFrame are assigned 0..n_datasets.  This is just a convention (they could be anything, but we choose to sequentially assign them starting at zero).

So I would change your assertion to read:

```
assert fdpm_matrix.shape[0] == max(returned_datasets.soma_joinid) + 1
```

bkmartinjr (approved):
This looks good.  

One requests before you land this:  [lease document in the PR description the assumed data design, which is that the _shape_ of the presence matrix on the first dimension must match the shape of the `datasets` dataframe.

Thank you!

tools/cell_census_builder/tests/test_builder.py
bkmartinjr: BTW, to clear up the above comment - the spec is _silent_ on the shape of the presence matrix, and only specifies its contents.  I believe that fixing the shape as proposed by this PR is consistent with the spec.

bkmartinjr (approved):
LGTM

PR #237, opened 2023-03-03
refactor ExperimentBuilder and validator for improved validator isolation

### LINKED ISSUE #179: validator skips some assertions in standalone mode
If the Validator is invoked in standalone mode, ExperimentBuilders will not have all of their instance variables initialized. As the validator logic uses some of these (n_obs, n_var), the validator must skip some of these checks for now. We should have these checks re-compute these missing values.

### LINKED ISSUE #215: [census builder] improve state isolation in the validator sub-system
The census builder validator is not sufficiently isolated from the rest of the builder. This makes it more complex than necessary (e.g., reopen of experiments), and leaves it open to possible state contamination from the build process

I recommend:
- the validator should be driven by a subset of the ExperimentBuilder object which minimizes exposure to bugs in the build process itself.  For example, the `experiment` handle should be off limits, as should build-time accumulators (eg., n_obs).
- the validator should not re-use the SOMA/TileDB context, and should instead use a clean/new context, with default timestamp, so as to mimic reading behavior of (most) other users.

See [discussion here](https://github.com/chanzuckerberg/cell-census/pull/154#discussion_r1107709939).

See related #179

bkmartinjr:
Fixes #215 
Fixes #179 

Changes:
* Break the ExperimentBuilder into two parts:
    * ExperimentSpecification - read-only declaration of Experiment configuration.  May not contain information derived from the AnnData files.
    * ExperimentBuilder - the accumulated state used during the build process, derived from AnnData input files.
* Refactor the validator to _only_ use the ExperimentSpecifications, increasing isolation from the build process. As a side effect, this fixes #179
* Minor fixes to test_builder which was incorrectly creating AnnData's (this cleans up all of the warnings generated during the test execution)
* Added `attrs` dependency (preferred over dataclasses, and used to create ExperimentSpecification)

### DIFF
--- a/tools/cell_census_builder/__main__.py
+++ b/tools/cell_census_builder/__main__.py
@@ -149,9 +149,6 @@ def build(
     # Step 5- write out dataset manifest and summary information
     build_step5_populate_summary_info(root_collection, experiment_builders, filtered_datasets, args.build_tag)
 
-    for eb in experiment_builders:
-        eb.build_completed = True
-
     # consolidate TileDB data
     if args.consolidate:
         consolidate(args, root_collection.uri)
--- a/tools/cell_census_builder/__main__.py
+++ b/tools/cell_census_builder/__main__.py
@@ -13,7 +13,12 @@ from .anndata import open_anndata
 from .census_summary import create_census_summary
 from .consolidate import consolidate
 from .datasets import Dataset, assign_dataset_soma_joinids, create_dataset_manifest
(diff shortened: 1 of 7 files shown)


bkmartinjr:
*Note:* tests are failing due to the mismatch between the current state of tests in `main` and the current census build, specifically around the shape of `presence`.  Will be resolved when we do a new build.

atolopko-czi (approved):
Very nice cleanup of numerous hacks!

PR #243, opened 2023-03-06
Add validation data checks

### LINKED ISSUE #11: improve validation phase of census builder

Bento007:
Fixes #11 

# features
- Verify the contents of the X matrix are EQUAL for all var values present in the AnnData
- Verify the contents of the X matrix are EMPTY for all var ids NOT present in the AnnData. Test by asserting that
  no col IDs contain a joinid not in the AnnData.
- Verify the contents of the presence matrix match the features present
  in the AnnData (where presence is defined as having a non-zero value)
- Modify the test dataset generated to add more variability.

# fixes
- remove duplicate check for path names
- add missing types to validate_all_soma_objects_exists docstring
- Change feature_name to feature_id when generating a dataset. Feature_id better represent how it is used.
- clean up `_validate_X_layers_contents_by_dataset` to run faster and improve readability.
- remove unused parameter from `validate_consolidation`

# DOD
- the validation steps must be correct
- the new validation step shall not significantly increase the run time of the validation step.


tools/cell_census_builder/validate.py
```diff
@@ -281,30 +305,29 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
...
-                    ad.X
-                ), f"{eb.name}:{dataset.dataset_id} 'raw' nnz mismatch {raw_nnz} vs {count_nonzero(ad.X)}"
+                ad_x = ad.X
             else:
-                assert raw_nnz == count_nonzero(
-                    ad.raw.X
-                ), f"{eb.name}:{dataset.dataset_id} 'raw' nnz mismatch {raw_nnz} vs {count_nonzero(ad.raw.X)}"
-
+                ad_x = ad.raw.X
+            h5ad_sum = sum(ad_x.data)  # calculate the actual sum of the X matrix values
```
bkmartinjr, line 323: Why test the sum and not test for actual per-element equality?  Equality of sums still allows for a  number of errors (e.g., incorrectly assigned coordinates).  Given that both matrices are loaded and in memory, it is a small step to just compare for actual equality.

In other words, something like:
```
raw_coo = exp.ms[MEASUREMENT_RNA_NAME].X["raw"].read().coos().concat().to_scipy()
assert ad_x.shape == raw_coo.shape
assert (ad_x.tocoo() != raw_coo).nnz == 0
```

It will use some extra memory, but it seems worth it.

A direct equality comparision will provide a super set of both the sum and nnz tests.

tools/cell_census_builder/tests/conftest.py
```diff
@@ -60,20 +63,20 @@ def get_h5ad(organism: Organism) -> anndata.AnnData:
...
             "feature_reference": organism.organism_ontology_term_id,
         },
-        index=feature_name,
+        index=feature_id,
     )
     var = var_dataframe
 
     # Create embeddings
-    random_embedding = np.random.rand(4, 4)
+    random_embedding = np.random.rand(cells, genes)
```
bkmartinjr, line 79: tiny nit: this is using an obsolete numpy API.  New code would ideally use the Generator API.

(optional to fix)

bkmartinjr (changes requested):
Primary concern is the new test for X matrix sum & nnz is still weaker than an element equality check, which isn't that much more expensive (primarily costs more memory), and offers much stronger guarantees.

Happy to discuss if there is an argument against a full equality check, but based on what I know today, I'd like to see us go the extra mile and _really_ validate.

tools/cell_census_builder/validate.py
bkmartinjr: BTW, if you aren't familiar with the pitfalls of comparing sparse matrices, LMK.  Comparison is an odd beast:
* if the shapes or types differ, it returns True/False
* otherwise it does an element-wise compare (which allocates another matrix of the same size)
So the use of `!=` is a trick - if you expect the compared matrices to be equal, this will allocate a result that is all `False`, which is `0`, which is a very, very sparse matrix (i.e., no memory used).

Bento007: Here's what I tried:
```
raw_coo = exp.ms[MEASUREMENT_RNA_NAME].X["raw"].read((cell_soma_joinids, slice(None))).coos().concat().to_scipy()
assert ad_x.shape == raw_coo.shape
assert (ad_x.tocoo() != raw_coo).nnz == 0
```

The problem is the shape of the ad_x, is different from the raw_coo, and i'm not sure why or how to correct it.

Everything else about the matrixes appears to be correct, nnz, and manually comparing the matrix values looks correct.

tools/cell_census_builder/validate.py
```diff
@@ -281,30 +305,29 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
...
+            assert raw_sum == h5ad_sum, error.format(eb.name, dataset.dataset_id, "sum", raw_sum, h5ad_sum)
+            h5ad_nnz = _count_nonzero(ad_x.data)
+            assert raw_nnz == h5ad_nnz, error.format(eb.name, dataset.dataset_id, "nnz", raw_nnz, h5ad_nnz)
+            h5ad_gene_count = np.count_nonzero(ad_x.sum(axis=0))
+            assert h5ad_gene_count == fdpm_gene_count, error.format(
+                eb.name, dataset.dataset_id, "gene_count", fdpm_gene_count, h5ad_gene_count
```
atolopko-czi, line 329: nit: prefix or suffix test value variables with `expected`  and `actual`, for readability

tools/cell_census_builder/tests/conftest.py
```diff
@@ -104,13 +107,13 @@ def soma_path(tmp_path: pathlib.Path) -> str:
 def datasets(assets_path: str) -> List[Dataset]:
     datasets = []
     for organism in ORGANISMS:
-        for dataset_id in range(2):
-            h5ad = get_h5ad(organism)
-            h5ad_path = f"{assets_path}/{organism.name}_{dataset_id}.h5ad"
+        for i in range(2):
```
atolopko-czi, line 110: add `NUM_DATASETS` constant and
```suggestion
        for i in range(NUM_DATASETS):
```

atolopko-czi (commented):
A few optional nits, while you're continuing to work on all-value comparison logic.

tools/cell_census_builder/validate.py
```diff
@@ -278,56 +278,49 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
...
+                var_query=soma.AxisQuery(value_filter=f"feature_id in {ad_genes}"),
+            ) as query:
+                fdpm_gene_count = 0
+                raw_coo = None
+                if query.n_obs > 0:
+                    assert soma.SparseNDArray.exists(exp.ms[MEASUREMENT_RNA_NAME].X["raw"].uri)
+                    fdpm_gene_count = _count_elements(
+                        exp.ms["RNA"][FEATURE_DATASET_PRESENCE_MATRIX_NAME], [dataset.soma_joinid]
+                    )
+                    raw_coo = query.to_anndata("raw").X.tocoo()  # TODO: slow, find a faster method.
```
Bento007, line 314: @bkmartinjr is there a better way of converting this?

atolopko-czi: Maybe try to avoid the `to_anndata` conversion: `query.X('raw').coos().concat().to_scipy()`

Bento007: This gives the wrong shape I run this with `test_base_builder_creation`. Converting to anndata has been the only way I've found so far to get the h5ad and cell_census to have the same shape.

Bento007:
The speed up commit above actually runs slower than converting to anndata which was the original approach.

tools/cell_census_builder/validate.py
```diff
@@ -255,56 +256,80 @@ def validate_axis_dataframes(
...
-                assert raw_nnz == count_nonzero(
-                    ad.raw.X
-                ), f"{eb.name}:{dataset.dataset_id} 'raw' nnz mismatch {raw_nnz} vs {count_nonzero(ad.raw.X)}"
-
+                expected_ad_x = ad.raw.X
+
+            with exp.axis_query(
+                measurement_name="RNA",
+                obs_query=soma.AxisQuery(value_filter=f"dataset_id == '{dataset.dataset_id}'"),
+                var_query=soma.AxisQuery(value_filter=f"feature_id in {[*ad.var.index.array]}"),
```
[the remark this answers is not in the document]
Bento007: > It will also be faster to find the dataset's soma_joinid from census['census_summary']['census_datasets'] and pass that in for obs_query= via coords and not value_filter

This will the accuracy of this check to the accuracy of the  `census['census_summary']`. The `value_filter` while slower is less dependent on previous assumptions.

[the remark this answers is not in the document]
atolopko-czi: Fair concern. Though we could validate separately that a dataset soma_joinid return all and only rows for the given datasets (maybe we already are somewhere?).

Bento007:
Performance of builder test from github actions
main: 25.81s
not_directly_comparing: 27.00s
convert_to_anndata: 28.34s
convert_to_coo: 36.81s
remove var filter: 27.17s

tools/cell_census_builder/validate.py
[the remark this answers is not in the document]
Bento007: Removing the var value_filter, speed up the tests significantly

atolopko-czi:
> Is it correct that the validator should only be run on a local snap-shot?

Yes. My suggestion to try it on the s3 census directly was hopeful in nature. :). Downloading to an ec2 instance should be simple enough for this testing. (Faster and slightly cheaper than downloading to a local dev machine).

tools/cell_census_builder/validate.py
```diff
@@ -255,56 +256,77 @@ def validate_axis_dataframes(
...
+                    actual_x_coo = query.X("raw").coos().concat().to_scipy()
+                    # translating the rows
+                    actual_x_coo.row = actual_x_coo.row - actual_x_coo.row.min()
+                    # mapping the var_soma_joinid to the anndata feature_ids
+                    actual_genes = query.var(column_names=["soma_joinid", "feature_id"]).concat().to_pandas()
+                    gene_mapping = ad.var.join(actual_genes.set_index("feature_id"))
+                    gene_mapping["ad_id"] = range(gene_mapping.shape[0])
+                    gene_mapping = gene_mapping.set_index("soma_joinid")
+                    # actual_x_coo.col.map()
+                    for i in range(actual_x_coo.col.size):
```
Bento007, line 309: This loops is slow

tools/cell_census_builder/validate.py
```diff
@@ -271,39 +274,82 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
...
+            if len(obs_joinids) == 0:
+                continue
 
-            def count_nonzero(arr: Union[sparse.spmatrix, npt.NDArray[Any]]) -> int:
-                """Return _actual_ non-zero count, NOT the stored value count."""
-                if isinstance(arr, (sparse.spmatrix, sparse.coo_array, sparse.csr_array, sparse.csc_array)):
-                    return np.count_nonzero(arr.data)
-                return np.count_nonzero(arr)
+            X_raw = (
+                exp.ms[MEASUREMENT_RNA_NAME]
```
bkmartinjr, line 296: super small item:  you might want to change all of the hard-coded `"RNA"` strings to use the global `MEASUREMENT_RNA_NAME`, so the code is consistent.

I apologize in advance for my inconsistent coding!

tools/cell_census_builder/validate.py
```diff
@@ -271,39 +274,82 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
...
+
+            # Assertion 3- the contents of the presence matrix match the features present
+            # in the AnnData (where presence is defined as having a non-zero value)
+            presence = (
+                exp.ms[MEASUREMENT_RNA_NAME][FEATURE_DATASET_PRESENCE_MATRIX_NAME]
+                .read((dataset.soma_joinid,))
+                .tables()
+                .concat()
+            )
+            assert var_joinid_in_adata[presence["soma_dim_1"].to_numpy()].all()
```
bkmartinjr, line 349: I forgot to add messages to these assertions.  

Optional:  you could add them if you want to improve the code.

bkmartinjr (approved):
Two minor and optional comments.  Otherwise it LGTM.

Suggest waiting for @atolopko-czi to also approve, as I'm essentially approving some of my own code, which isn't best practice :-)

tools/cell_census_builder/validate.py
```diff
@@ -271,39 +276,89 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
...
+            # Assertion 1 - the contents of the X matrix are EQUAL for all var values
+            # present in the AnnData
+            assert (
+                sparse.coo_matrix((data, (rows_by_position, cols_by_position)), shape=expected_ad_x.shape)
+                != expected_ad_x
+            ).nnz == 0, f"{eb.name}:{dataset.dataset_id} the X matrix elements are not equal."
+
+            # Assertion 2 - the contents of the X matrix are EMPTY for all var ids
+            # NOT present in the AnnData. Test by asserting that no col IDs contain
+            # a joinid not in the AnnData.
```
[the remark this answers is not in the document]
atolopko-czi: I'm probably missing something here, since I'm not seeing the case where assertion 1 (full value quality) passes and assertion 2 fails. In other words, if X contained data for an "extra" gene wouldn't that cause the full coo_matrix inequality check, above, to fail? Doesn't the inequality check consider the coords?

E.g, this case, where `ccX` has a value at col1, and `adX` does not, fails the original equality check:
```
In [1]: ccX = sc.sparse.coo_matrix(([0, 1, 2], ([0, 0, 0], [0, 1, 2])), shape=(1,4))

In [2]: adX = sc.sparse.coo_matrix(([0, 2], ([0, 0], [0, 2])), shape=(1,4))

In [3]: (ccX != adX).nnz
Out[3]: 1
```

atolopko-czi (approved):
LGTM! I do have one question re: necessity of an assertion, and one nit (var rename suggestion).

tools/cell_census_builder/validate.py
```diff
@@ -261,8 +263,11 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
     Intended to be dispatched from validate_X_layers.
 
     Currently, implements weak tests:
-    * the nnz is correct
-    * there are no zeros explicitly saved (this is mandated by cell census schema)
+    * 1 - the contents of the X matrix are EQUAL for all var values present in the AnnData
+    * 2 - the contents of the X matrix are EMPTY for all var ids NOT present in the AnnData. Test by asserting that
+      no col IDs contain a joinid not in the AnnData.
```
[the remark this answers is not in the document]
bkmartinjr: I think your proposed text is also incorrect/misleading - see the code in Assertion 2

What the test does is get all var soma_joinids corresponding to feature_ids *not* in the AnnData, and then verify that the X slice does *not* contain these joinids.  So the correct comment would be something like:   

> Test by asserting that no the X census slice does not contain dim_1 IDs associated with features _not_ in the AnnData var dataframe.

tools/cell_census_builder/validate.py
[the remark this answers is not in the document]
bkmartinjr: Your assumption fails to catch the case where the Census incorrectly has values in a var (feature_id) which is not present in the AnnData.

The first test restricts itself to the feature_ids present in the AnnData, and confirms that the values and coordinates in the Census are "eq".  But this leaves it blind to the case where the census has garbage values outside of those feature ids.

This approach is used because we need to reindex the matrix using the AnnData var dataframe index (i.e., using it's ordering/iloc).  As that index does not contain var feature_ids not present in itself (:-)), it was simplest to split this test into two separate assertions.

atolopko-czi (approved):
Latest changes LGTM

PR #244, opened 2023-03-06
add unit tests for python cell_census package

bkmartinjr:
Changes - more/better unit tests for the cell_census python package:
* add error condition unit test coverage for `open_soma` and other API
* add tests to cover all `open_soma` parameters
* add missing absolute URI test for `uri_join`

### DIFF
--- a/api/python/cell_census/tests/test_directory.py
+++ b/api/python/cell_census/tests/test_directory.py
@@ -33,6 +33,9 @@ DIRECTORY_JSON = {
             "s3_region": "us-west-2",
         },
     },
+    # An explicitly dangling tag, to confirm we handle correct
+    # Underscore indicates expected failure to test below
+    "_dangling": "no-such-tag",
 }
 
 
@@ -50,7 +53,7 @@ def test_get_census_version_directory(directory_mock: Any) -> None:
     assert all((type(v) == dict for v in directory.values()))
 
     for tag in DIRECTORY_JSON:
-        assert tag in directory
+        assert tag[0] == '_' or tag in directory
         if isinstance(DIRECTORY_JSON[tag], dict):
             assert directory[tag] == DIRECTORY_JSON[tag]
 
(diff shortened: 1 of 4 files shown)


api/python/cell_census/tests/test_open.py
```diff
@@ -37,16 +71,42 @@ def test_get_source_h5ad_uri() -> None:
...
         census_datasets = census["census_info"]["datasets"].read().concat().to_pandas()
 
     small_dataset = census_datasets.nsmallest(1, "dataset_total_cell_count").iloc[0]
+    assert isinstance(small_dataset.dataset_id, str)
+    return small_dataset.dataset_id
+
+
+@pytest.mark.live_corpus
+def test_download_source_h5ad(tmp_path: pathlib.Path, small_dataset_id: str) -> None:
+    # with cell_census.open_soma(census_version="latest") as census:
```
ebezzi, line 91: Can the commented code be removed?

api/python/cell_census/tests/test_directory.py
```diff
@@ -60,6 +63,11 @@ def test_get_census_version_directory(directory_mock: Any) -> None:
         assert directory[tag] == cell_census.get_census_version_description(tag)
 
 
+def test_get_census_version_description_errors() -> None:
+    with pytest.raises(KeyError):
```
ebezzi, line 67: Not strictly related to the tests, but I was wondering if maybe we don't want to specialize these exceptions? Right now everything is `KeyError` or `ValueError` but these exceptions map to very specific issues (for instance, non existent census version.

ebezzi (approved):
LGTM, one nit and one more generic comment (not strictly related to this PR).

api/python/cell_census/tests/test_open.py
bkmartinjr: oops!  Yes, definitely.

api/python/cell_census/tests/test_directory.py
bkmartinjr: sure, we could do that.

I agree it is a separate issue and I'll file it.

PR #250, opened 2023-03-10
[r] add source h5ad accessor functions

mlin:
Adding `CellCensus::{get_source_h5ad_uri,download_source_h5ad}` with tests and (minimal) docstrings


api/r/CellCensus/R/open.R
```diff
@@ -5,26 +5,30 @@ DEFAULT_TILEDB_CONFIGURATION <- c(
...
+#' @param s3_region Optional AWS region accompanying uri.
 #'
 #' @return Top-level `tiledbsoma::SOMACollection` object
 #' @importFrom tiledbsoma SOMACollection
 #' @importFrom tiledb tiledb_ctx
 #' @export
 #'
 #' @examples
-open_soma <- function(census_version = "latest", uri = "") {
+open_soma <- function(census_version = "latest", uri = "", s3_region = "") {
```
bkmartinjr, line 19: is `s3_region` a temporary placeholder for passing a full `context`?   At least in the Python API, the equivalent function accepts a context, which contains any necessary TileDB configuration.

If a placeholder, please disregard this comment.  If not, should it instead be a `context`?

bkmartinjr (approved):
caveat: cursory review by non-R-thonic person!  One open issue about the signature of `open_soma`, but otherwise lgtm.

api/r/CellCensus/R/open.R
mlin: `SOMATileDBContext` was *just* added to the R API over the weekend, so I'll do a follow-up to receive and pass it around!

PR #251, opened 2023-03-10
[ci] fix paths-ignore in workflow YMLs

mlin:
I must have goofed these originally.


atolopko-czi (approved):
LGTM.  Alternately, [we could rename the directory](https://czi-sci.slack.com/archives/C048BN7K6U8/p1674825932582749).

PR #253, opened 2023-03-13
X contents validation

bkmartinjr:
Proposed revision to test full contents of the X matrix in validation pass, plus presence contents.

Tested on a full census build, on the standard config, and it ran 5 hrs 10 minutes.

### DIFF
--- a/tools/cell_census_builder/validate.py
+++ b/tools/cell_census_builder/validate.py
@@ -281,54 +281,112 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
             anndata_cell_filter = make_anndata_cell_filter(eb.anndata_cell_filter_spec)
             ad = anndata_cell_filter(unfiltered_ad, retain_X=True)
             if ad.raw is None:
-                expected_ad_x = ad.X
+                expected_ad_x, expected_ad_var = ad.X, ad.var
             else:
-                expected_ad_x = ad.raw.X
+                expected_ad_x, expected_ad_var = ad.raw.X, ad.raw.var
 
-            with exp.axis_query(
-                measurement_name="RNA", obs_query=soma.AxisQuery(value_filter=f"dataset_id == '{dataset.dataset_id}'")
-            ) as query:
-                actual_census_gene_count = 0
-                actual_x_coo = None
-                if query.n_obs > 0:
-                    assert soma.SparseNDArray.exists(exp.ms[MEASUREMENT_RNA_NAME].X["raw"].uri)
-                    actual_census_gene_count = _count_elements(
-                        exp.ms["RNA"][FEATURE_DATASET_PRESENCE_MATRIX_NAME], [dataset.soma_joinid]
-                    )
-
-                    # mapping the var_soma_joinid to the anndata feature_ids
-                    actual_genes = query.var(column_names=["soma_joinid", "feature_id"]).concat().to_pandas()
-                    gene_mapping = ad.var.join(actual_genes.set_index("feature_id"))[["soma_joinid"]]
-                    gene_mapping["ad_id"] = range(gene_mapping.shape[0])
-                    gene_mapping = gene_mapping.set_index("soma_joinid")
-                    actual_x = (
-                        query.X("raw")
-                        .tables()
-                        .concat()
-                        .to_pandas()
-                        .join(gene_mapping, on="soma_dim_1")[["soma_dim_0", "ad_id", "soma_data"]]
-                    )
-                    # translating the rows
-                    actual_x.soma_dim_0 = actual_x.soma_dim_0 - actual_x.soma_dim_0.min()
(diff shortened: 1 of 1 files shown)


tools/cell_census_builder/validate.py
```diff
@@ -276,59 +276,86 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
...
+                continue
+
+            X_raw = (
+                exp.ms[MEASUREMENT_RNA_NAME]
+                .X["raw"]
+                .read((obs_joinids.soma_joinid.to_numpy(), slice(None)))
+                .tables()
+                .concat()
+            )
+            data = X_raw["soma_data"].to_numpy()
```
Bento007, line 307: Isn't it expensive to expand out the whole sparse matrix?

tools/cell_census_builder/validate.py
```diff
@@ -276,59 +276,86 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
...
+            # positionally re-index the vars/cols using the feature_id as the join key.
+            cols_by_position = (
+                expected_ad_var.join(all_var_ids.set_index("feature_id"))
+                .set_index("soma_joinid")
+                .index.get_indexer(cols)
+            )
+
+            # Assertion 1 - the contents of the X matrix are EQUAL for all var values
+            # present in the AnnData
+            assert (
```
Bento007, line 333: I [read](https://stackoverflow.com/questions/30685024/check-if-two-scipy-sparse-csr-matrix-are-equal) that both shape and nnz should be equal before doing this comparison. Shapes makes sense. However I don't see how asserting nnz help.

tools/cell_census_builder/validate.py
```diff
@@ -276,59 +276,86 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
...
+                sparse.coo_matrix((data, (rows_by_position, cols_by_position)), shape=expected_ad_x.shape)
+                != expected_ad_x
+            ).nnz == 0, f"{eb.name}:{dataset.dataset_id} the X matrix elements are not equal."
+
+            # Assertion 2 - the contents of the X matrix are EMPTY for all var ids
+            # NOT present in the AnnData. Test by asserting that no col IDs contain
+            # a joinid not in the AnnData.
+            assert (
+                var_joinid_in_adata.all()
+                or not pd.Series(cols).isin(all_var_ids[~var_joinid_in_adata].soma_joinid).any()
```
Bento007, line 343: Can you explain this check for me and why we have an `or`?

Bento007 (commented):
Looks good

tools/cell_census_builder/validate.py
```diff
@@ -276,59 +276,86 @@ def _validate_X_layers_contents_by_dataset(args: Tuple[str, str, Dataset, List[E
...
-                expected_ad_x_coo = expected_ad_x.tocoo()
-                assert expected_ad_x_coo.nnz == actual_x_coo.nnz, error.format(
-                    "nnz", actual_x_coo.nnz, expected_ad_x.nnz
-                )
-                assert expected_ad_x_coo.shape == actual_x_coo.shape, error.format(
-                    "shape", actual_x_coo.shape, expected_ad_x_coo.shape
+                expected_ad_x, expected_ad_var = ad.raw.X, ad.raw.var
+
+            # get the joinids for the obs axis
+            obs_joinids = (
```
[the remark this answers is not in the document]
bkmartinjr: Yes -- it is simpler and has fewer dependencies. The query layer is a more complex class that wraps this same operation (plus some additional overhead).  It introduces additional dependencies into the builder.  I don't see any benefits.

tools/cell_census_builder/validate.py
bkmartinjr: sure, but if you want to compare every value withe AnnData, you don't have any choice.  You must read it all in order to check for equality.

tools/cell_census_builder/validate.py
bkmartinjr: You may be misunderstanding the code -- this an equality test implemented with `__ne__` and an nnz check.  This pattern prevents the equality check from having to materialize an full array of bools (remember, the equality check is per-element).

See for example, this [stack overflow answer](https://stackoverflow.com/questions/30685024/check-if-two-scipy-sparse-csr-matrix-are-equal).  

Highly suggest you study the sparse array implementation and documentation to get a better understanding of how it works.

tools/cell_census_builder/validate.py
bkmartinjr: consider the case where _all_ of the values in the mask are True.  In that case, the second clause will fail (the arg to `isin` will be an empty set), so it is protected with the first clause.

Bento007 (approved):

PR #254, opened 2023-03-13
open_anndata unit test

ebezzi:
Unit test for the `open_anndata` function.

The way I implemented fixtures for this test isn't very elegant but it's functional. Suggestions on how to improve it are welcome.

### DIFF
--- a/tools/cell_census_builder/anndata.py
+++ b/tools/cell_census_builder/anndata.py
@@ -6,7 +6,7 @@ import numpy as np
 import pandas as pd
 
 from .datasets import Dataset
-from .globals import CXG_SCHEMA_VERSION, CXG_SCHEMA_VERSION_IMPORT, FEATURE_REFERENCE_IGNORE, RNA_SEQ
+from .globals import CXG_SCHEMA_VERSION, CXG_SCHEMA_VERSION_IMPORT, FEATURE_REFERENCE_IGNORE
 from .util import uricat
 
 AnnDataFilterSpec = TypedDict(
@@ -114,7 +114,7 @@ def make_anndata_cell_filter(filter_spec: AnnDataFilterSpec) -> AnnDataFilterFun
         if organism_ontology_term_id is not None:
             obs_mask = obs_mask & (ad.obs.organism_ontology_term_id == organism_ontology_term_id)
         if assay_ontology_term_ids is not None:
-            obs_mask = obs_mask & ad.obs.assay_ontology_term_id.isin(RNA_SEQ)
+            obs_mask = obs_mask & ad.obs.assay_ontology_term_id.isin(assay_ontology_term_ids)
 
         # multi-organism dataset cell filter - exclude any cells where organism != feature_reference
         feature_references = set(ad.var.feature_reference.unique()) - FEATURE_REFERENCE_IGNORE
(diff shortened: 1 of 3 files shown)


tools/cell_census_builder/tests/anndata/test_anndata.py
```diff
@@ -0,0 +1,62 @@
+from typing import List
```
bkmartinjr, line 1: any reason to not include a simple test for the schema version?  (ie., that we drop any unsupported schema versions)?

bkmartinjr: I would also add a test to ensure it correctly handles the case where the AnnData lacks a raw (i.e., adata.raw is None)

ebezzi: Good idea, I will add both.

bkmartinjr (changes requested):
No concerns with the tests, but recommend adding a test for the other conditions the code must handle:
* incorrect cellxgene schema version (drops the AnnData)
* no raw layer

tools/cell_census_builder/tests/anndata/test_anndata.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # The X matrix should have the same size as raw.X
+    assert h5ad.X.shape == h5ad.raw.X.shape
+
+    # The 4th column (correspnding to the added gene) is all zeros
+    added_col = h5ad.X.todense()[:, 3]
+    assert np.array_equal(added_col.A1, np.zeros(4))
+
+
+import pytest
```
Bento007, line 66: why is this import down here?

ebezzi: Was temporary. The fixtures have been moved to conftest.py

ebezzi:
> No concerns with the tests, but recommend adding a test for the other conditions the code must handle:
> 
> * incorrect cellxgene schema version (drops the AnnData)
> * no raw layer

@bkmartinjr added a test for the incorrect schema version. Turns out the default datasets already have no raw layer, so that's already tested. I explicitly added a statement in the test description.

tools/cell_census_builder/anndata.py
```diff
@@ -114,7 +114,7 @@ def _filter(ad: anndata.AnnData, retain_X: Optional[bool] = True) -> anndata.Ann
         if organism_ontology_term_id is not None:
             obs_mask = obs_mask & (ad.obs.organism_ontology_term_id == organism_ontology_term_id)
         if assay_ontology_term_ids is not None:
-            obs_mask = obs_mask & ad.obs.assay_ontology_term_id.isin(RNA_SEQ)
+            obs_mask = obs_mask & ad.obs.assay_ontology_term_id.isin(assay_ontology_term_ids)
```
ebezzi, line 117: This is a bug in the current version of the code, as per discussion with @bkmartinjr yesterday.

bkmartinjr:
> Turns out the default datasets already have no raw layer

Just want to confirm that we also test datasets _with_ a raw layer as well?  I believe you do, but confirming....

ebezzi:
> > Turns out the default datasets already have no raw layer
> 
> Just want to confirm that we also test datasets _with_ a raw layer as well? I believe you do, but confirming....

Yes, that is tested by `test_open_anndata_equalizes_raw_and_normalized`. If you think it's worth it, I can add one test where the raw layer exists and has the same shape as the normalized one.

bkmartinjr (approved):
LGTM!  Thanks for the enhancements.

PR #258, opened 2023-03-14
Unit tests for staging datasets

ebezzi:
Add unit tests for these 3 functions:
1. build_step1_get_source_datasets
2. `source_assets`
3. `load_manifests`

### DIFF
--- a/tools/cell_census_builder/tests/test_manifest.py
+++ b/tools/cell_census_builder/tests/test_manifest.py
@@ -1,0 +1,134 @@
+import io
+import pathlib
+import re
+from unittest.mock import patch
+
+import pytest
+
+from tools.cell_census_builder.manifest import CXG_BASE_URI, load_manifest
+
+
+@pytest.fixture
+def manifest_csv(tmp_path: pathlib.Path) -> io.TextIOWrapper:
+    # assets_path = f"{tmp_path}/h5ads"
+    # os.mkdir(assets_path)
+    manifest_content = """
+    dataset_id_1, data/h5ads/dataset_id_1.h5ad
+    dataset_id_2, data/h5ads/dataset_id_2.h5ad
+    """
+    path = f"{tmp_path}/manifest.csv"
+    h5ad_path = f"{tmp_path}/data/h5ads/"
+    pathlib.Path(h5ad_path).mkdir(parents=True, exist_ok=True)
+    pathlib.Path("data/h5ads/dataset_id_1.h5ad").touch()
+    pathlib.Path("data/h5ads/dataset_id_2.h5ad").touch()
+    with open(path, "w+") as f:
+        f.writelines(manifest_content.strip())
+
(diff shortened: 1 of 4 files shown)


tools/cell_census_builder/source_assets.py
```diff
@@ -11,6 +11,10 @@
 from .mp import cpu_count, create_process_pool_executor
 
 
+def _get_n_workers() -> int:
```
ebezzi, line 14: Required for testing. In GHA, running with more than 1 worker causes the test to fail.

bkmartinjr: Can you point me at more information on this constraint?  I'm trying to understand _why_ it would fail on GHA.

If it is the case that GHA processes may not fork more than some limited number of child processes, we would be better off just testing this with the command line arg (`args` param) with a `--max-workers 1`.  This is the only code that calls `create_process_pool_executor` with the optional override, and it only affects the behavior when `args` lacks a `max_workers` value.

tools/cell_census_builder/tests/test_manifest.py
```diff
@@ -0,0 +1,113 @@
+import io
```
bkmartinjr, line 1: you might consider a test that loads from manifest, where the manifest has a duplicate, thereby testing the `dedup_datasets` code.

This is not critical, but I added the dedup code when I accidentally created a file-based manifest with the same dataset ID!

bkmartinjr (changes requested):
Super minor changes requested.  Overall it is looking good.

tools/cell_census_builder/source_assets.py
ebezzi: The error is [here](https://github.com/chanzuckerberg/cell-census/actions/runs/4419756699/jobs/7748726289): 
```concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.```
I didn't notice that the max_workers can be specified in args. Will use that. I'm not sure it's worth investigating the issue above - in my experience testing doesn't always play well with multiprocessing, so since this test isn't specifically meant to test multiprocessing, I think we can just disable it.

tools/cell_census_builder/tests/test_manifest.py
ebezzi: 👍 will add.

ebezzi: Added.

tools/cell_census_builder/source_assets.py
bkmartinjr: Ah, this means something killed the child process.  Commonly this is an out-of-memory condition on the system, or something else that causes an uncontrolled failure (segv, etc).  Super hard to diagnose in that environment.

bkmartinjr (approved):
LGTM!

PR #261, opened 2023-03-15
Util unit tests

ebezzi:
Unit tests for:
- `is_positive_integral`
- `array_chunker`
- `uricat`

### DIFF
--- a/tools/cell_census_builder/tests/test_util.py
+++ b/tools/cell_census_builder/tests/test_util.py
@@ -1,0 +1,30 @@
+import numpy as np
+from scipy.sparse import csr_matrix
+
+from tools.cell_census_builder.util import is_positive_integral
+
+
+def test_is_positive_integral() -> None:
+    X = np.array([1, 2, 3, 4])
+    assert is_positive_integral(X)
+
+    X = np.array([-1, 2, 3, 4])
+    assert not is_positive_integral(X)
+
+    X = np.array([1.2, 0, 3, 4])
+    assert not is_positive_integral(X)
+
+    X = np.zeros((3, 4))
+    assert is_positive_integral(X)
+
+    X = csr_matrix([[1, 2, 3], [4, 5, 6]])
+    assert is_positive_integral(X)
+
+    X = csr_matrix([[-1, 2, 3], [4, 5, 6]])
+    assert not is_positive_integral(X)
+
+    X = csr_matrix([[1.2, 0, 3], [4, 5, 6]])
+    assert not is_positive_integral(X)
+
+    X = csr_matrix([0, 0, 0])
+    assert is_positive_integral(X)
--- a/tools/cell_census_builder/tests/test_util.py
+++ b/tools/cell_census_builder/tests/test_util.py
@@ -1,7 +1,8 @@
 import numpy as np
-from scipy.sparse import csr_matrix
+import pytest
+from scipy.sparse import coo_matrix, csr_matrix, triu
 
-from tools.cell_census_builder.util import is_positive_integral
+from tools.cell_census_builder.util import array_chunker, is_positive_integral, uricat
 
 
 def test_is_positive_integral() -> None:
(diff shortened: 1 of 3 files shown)


tools/cell_census_builder/tests/test_util.py
```diff
@@ -0,0 +1,106 @@
+import numpy as np
+import pytest
+from scipy.sparse import coo_matrix, csr_matrix, triu
+
+from tools.cell_census_builder.util import array_chunker, is_positive_integral, uricat
+
+
+def test_is_positive_integral() -> None:
```
ebezzi, line 8: To be pedantic, this method should be called `is_nonnegative_integral` since 0 is an accepted value. If that looks better I can make the change.

bkmartinjr: feel free to improve it!

bkmartinjr: If you do this, please also add a test of empty array, e.g.,

> assert not is_nonnegative_integral(np.array([], dtype=np.float32))

and equivalent for sparse.

tools/cell_census_builder/tests/test_util.py
```diff
@@ -0,0 +1,106 @@
...
+    X = np.array([1, 2, 3, 4])
```
bkmartinjr, line 9: these result in an array of dtype np.int64. The function only accepts floats (both float32 and float64), although it does not raise an error if the input violates the type signature.

These tests should explicitly create arrays of dtype float32 or float64.  Prefer float32 for most tests, as that is what we typically receive.

tools/cell_census_builder/tests/test_util.py
```diff
@@ -0,0 +1,106 @@
+import numpy as np
+import pytest
+from scipy.sparse import coo_matrix, csr_matrix, triu
+
+from tools.cell_census_builder.util import array_chunker, is_nonnegative_integral, uricat
+
+
+def test_is_nonnegative_integral() -> None:
```
bkmartinjr, line 8: need to add a few tests:
* empty (zero length) ndarray
* 2D ndarray (not just 1D)
* spmatrix of shape (0, 0)

tools/cell_census_builder/tests/test_util.py
```diff
@@ -0,0 +1,106 @@
...
+    chunked = list(array_chunker(X, nnz_chunk_size=10))
+    assert len(chunked) == 5
+    assert chunked[0].nnz == 3
+    assert chunked[1].nnz == 7
+    assert chunked[2].nnz == 11
+    assert chunked[3].nnz == 15
+    assert chunked[4].nnz == 19
+
+    # Other formats are rejected by the method
+    X = triu(np.ones(100).reshape(10, 10)).tocoo()
```
bkmartinjr, line 98: recommend testing this by passing something other than COO (eg., LIL or BSR), as it is fairly likely we will add COO support the the array_chunker at some point.  Just change the final call from `tocoo()` to `tolil()` (or anything other than csr/csc/coo)

tools/cell_census_builder/tests/test_util.py
```diff
@@ -0,0 +1,106 @@
...
+    with pytest.raises(NotImplementedError):
+        list(array_chunker(X))
+
+
+def test_uricat() -> None:
```
bkmartinjr, line 103: need to add a test where there are multiple absolute paths, e.g.,

> assert uricat("file://foo", "file://bar") == "file://bar"

tools/cell_census_builder/util.py
```diff
@@ -10,11 +10,13 @@
 from scipy import sparse
 
 
-def array_chunker(arr: Union[npt.NDArray[Any], sparse.spmatrix]) -> sparse.coo_matrix:
+def array_chunker(
+    arr: Union[npt.NDArray[Any], sparse.spmatrix],
+    nnz_chunk_size: Optional[int] = 256 * 1024**2,  # goal (~2.4GiB for a 32-bit COO)
+) -> Iterator[sparse.coo_matrix]:
     """
     Return the array as multiple chunks, each a coo_matrix.
```
bkmartinjr, line 18: Add param comment, e.g.,

nnz_chunk_size:  approximate number of elements in each chunk. 

I would also add a note that chunking is done by row, and will *never* divide a row across two chunks

bkmartinjr (changes requested):
couple of change requests. hit me up for a second review when ready.

tools/cell_census_builder/util.py
bkmartinjr: oh, doh - you are absolutely correct.

tools/cell_census_builder/tests/test_util.py
```diff
@@ -0,0 +1,133 @@
...
+    X = np.array([[1.2, 0, 3], [4, 5, 6]], dtype=np.float32)
+    assert not is_nonnegative_integral(X)
+
+    X = np.zeros((3, 4), dtype=np.float32)
+    assert is_nonnegative_integral(X)
+
+    X = csr_matrix([[1, 2, 3], [4, 5, 6]])
+    assert is_nonnegative_integral(X)
+
+    X = csr_matrix([[-1, 2, 3], [4, 5, 6]])
```
bkmartinjr, line 30: these are also int64 - same issue as with ndarray, they need a float dtype to be specified.

bkmartinjr (approved):
one more dtype issue on the `is_nonnegative_integral` test.

I'm marking as approved, on the assumption you will add those explicit dtypes....

tools/cell_census_builder/tests/test_util.py
ebezzi: Apologies, for some reason I thought scipy.sparse created float matrices by default. Fixing it.

PR #263, opened 2023-03-15
assertion to confirm correct MP configuration

bkmartinjr:
Adding assertion to the process pool creation utility to confirm that the system is correctly configured. This will prevent chasing of obscure bugs on systems which default to `fork` rather than `spawn`


ebezzi (approved):
Thanks Bruce!

PR #264, opened 2023-03-15
Update lifecycle annotations to be consistent with tiledbsoma

ebezzi:
The lifecycle annotations are now using their own directive. This will help when rendering the docs using Sphinx, and they look better in general.


pablo-gar (approved):
LGTM

PR #267, opened 2023-03-16
Dockerfile for census builder

ebezzi:
Dockerfile that can be used for platform independent census builds.

Usage:

1. Build with:
```
docker build --build-arg=COMMIT_SHA=$(git rev-parse --short HEAD) . -t census-builder
```

2. Modify the `build-census.yaml` file to include options and build commands. An example is included that runs the builder with a manifest file.

3. Run with:
```
docker run --mount type=bind,source="$(pwd)/data",target=/data census-builder
```
  The mount is necessary so that the built census can be retrieved and uploaded to S3 later.


entrypoint.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
ebezzi, line 1: This file is no longer necessary, but I included it as an example if we wanted a simplified build process (without the yaml file)

entrypoint.py
```diff
@@ -0,0 +1,31 @@
...
+    builder = config["census-builder"]
+    uri = builder["uri"]
+    add_args(builder, args)
+    commands = builder["commands"]
+    for cmd, opts in commands.items():
+        subcommand_args = args.copy()
+        subcommand_args.append(cmd)
+        add_args(opts, subcommand_args)
+        print(subcommand_args)
+        subprocess.call(subcommand_args)
```
ebezzi, line 31: Instead of calling `subprocess`, we could modify the builder so that it could be called directly. This would require, at minimum, to modify the `main` function to accept a `Namespace`, but I think for now I'd like to validate this approach before we make major changes.

bkmartinjr (approved):
This looks good to me as a starting point.  I can easily wire it up to the builder package.

thanks!

PR #275, opened 2023-03-20
Enable anonymous access to S3 bucket

### LINKED ISSUE #466: [Python] Update to tiledbsoma==1.2.3, remove tiledb pin
The current `cellxgene-census==1.0.0` contains a pin for `tiledb`, added in PR #454, as a work-around for a missing tiledbsoma dependency.  This tiledbsoma issue was fixed in single-cell-data/TileDB-SOMA#1296, and release in `tiledbsoma==1.2.3.

We need to do three things:
* Update our minimum `tiledbsoma` dependency to something later than `1.2.2`
* Remove the explicit pin of `tiledb`
* and release

ebezzi:
Fixes https://github.com/chanzuckerberg/cell-census/issues/163
Fixes #466


ebezzi:
Added a few unit tests. These set some fake credentials and try to do an S3 call. Currently they fail because the unsigned request isn't enabled - they should pass when enabled (tested using `aws s3 ls`).

Note that unsetting the two variables won't work since `boto` can get credentials from other sources (e.g. instance metadata). One would need to hack boto so that the credential resolution is bypassed altogether, but this simplified approach should be good enough.

api/python/cellxgene_census/pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies= [
     # NOTE: the tiledbsoma version must be >= to the version used in the Census builder, to
     # ensure that the assets are readable (tiledbsoma supports backward compatible reading).
     # Make sure this version does not fall behind the builder's tiledbsoma version.
-    "tiledbsoma==1.2.2",
+    "tiledbsoma==1.2.3",
```
bkmartinjr, line 34: given this update, please remove the `tiledb` pin below, and also update the PR description to note that this PR will _also_ fix #466

bkmartinjr: Also, is there an equivalent update required for the R build?

bkmartinjr (approved):
with one ask, this looks good.  Approving on the condition that you make the update requested to api/python/cellxgene_census/pyproject.toml

api/python/cellxgene_census/tests/test_open.py
```diff
@@ -188,3 +189,14 @@ def test_download_source_h5ad_errors(tmp_path: pathlib.Path, small_dataset_id: s
...
+
+def test_can_open_with_anonymous_access() -> None:
+    """
+    With anonymous access, `open_soma` must be able to access the census even with bogus credentials
+    """
+    os.environ["AWS_ACCESS_KEY_ID"] = "fake_id"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "fake_key"
+    with cellxgene_census.open_soma(census_version="latest") as census:
+        assert census is not None
+        assert isinstance(census, soma.Collection)
```
atolopko-czi, line 202: curious, does the test in fact _fail_ when `tiledb_config["vfs.s3.no_sign_request"] = "false"`? (I suppose we could assert that case as well.)

atolopko-czi (approved):
LGTM. 2 minor comments.

api/python/cellxgene_census/pyproject.toml
ebezzi: Removed. R has a similar upgrade needed but it's not straightforward. I'll do that as a separate PR.

api/python/cellxgene_census/tests/test_open.py
ebezzi: Yes. added a test.

PR #286, opened 2023-03-24
Refine release policy

pablo-gar:
Refines the existing guidelines and policy. Adds policy for long-term supported and weekly releases.

@bkmartinjr and @atolopko-czi please review for clarity and make suggestions


docs/cell_census_storage_and_release_policy.md
```diff
@@ -1,52 +1,73 @@
...
+**Last edited**: March, 2023.
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED" "MAY", and "OPTIONAL" in this document are to be interpreted as described in [BCP 14](https://tools.ietf.org/html/bcp14), [RFC2119](https://www.rfc-editor.org/rfc/rfc2119.txt), and [RFC8174](https://www.rfc-editor.org/rfc/rfc8174.txt) when, and only when, they appear in all capitals, as shown here.
 
----
+## Definitions
 
-The cell census MUST be stored in the following S3 root bucket:
+* **Cell Census build**: a SOMA collection with the Cell Census data [as specified in the Cell Census schema](https://github.com/chanzuckerberg/cell-census/blob/main/docs/cell_census_schema.md#data-encoding-and-organization). 
+* **Cell Census source h5ads**: the set of h5ad files used for create a Cell Census build.
```
atolopko-czi, line 10:
```suggestion
* **Cell Census source H5AD files**: the set of H5AD files used to create a Cell Census build.
```

docs/cell_census_storage_and_release_policy.md
```diff
@@ -1,52 +1,73 @@
...
+* **Cell Census release**: a Cell Census build that is officially and publicly hosted online.
```
atolopko-czi, line 11: one might ask "what does 'officially' mean?"; perhaps clarify by referencing below sections, otherwise just remove?

docs/cell_census_storage_and_release_policy.md
```diff
@@ -1,52 +1,73 @@
...
-This folder MUST contain all public builds and releases defined as:
+### Weekly Cell Census release (latest)
 
-* Census "build" is a SOMA collection / container containing cell data.
-* Census "RELEASE" is a "build" which has been officially blessed and released in a formal manner.
-* A "build" is named with a `tag`, which is a string of printable ASCII (eg, "foo", "2022-11-01", "1-rc2.9"):
-   `./cell-census/[tag]/`
-* a RELEASE is named with a tag following the conventional format `release-[counter]`:
-   `./cell-census/release-[counter]/`
+* It MUST be created and made available for public access every week on Mondays except for CZI holidays.
```
atolopko-czi, line 25: I would make this `SHOULD`, if that's reasonable. I figure either we're fully automating it, in which case holidays don't matter (barring build failures), or we should be less strict about exactly when it appears: e.g. "once per week, best effort"

docs/cell_census_storage_and_release_policy.md
```diff
@@ -1,52 +1,73 @@
...
+* Its date of publication MUST be used as the `release_date` (see [below](#Cell-Census-release-information)). 
+* All data from CELLxGENE Discover hosted at the `release_date` that meets the requirements [according to the schema](https://github.com/chanzuckerberg/cell-census/blob/main/docs/cell_census_schema.md#data-included) MUST be included in it.
+* It MUST be deleted the following Monday and after the next weekly Cell Census release is published.
```
atolopko-czi, line 28: Again, I think SHOULD is more appropriate, or even MAY

docs/cell_census_storage_and_release_policy.md
```diff
@@ -1,52 +1,73 @@
...
 
-All h5ads used to create the Cell Census MUST be copied within the release folder into a folder named `h5ads`:	
-`./cell-census/release-[counter]/h5ads/`
+The following S3 bucket MUST be used as the root to store Cell Census data:
 
-Any dataset changes, additions, or deletions per release MUST be documented in the following human-readable changelog file name `changelog.txt`:
+`s3://cellxgene-data-public/`
 
-`./cell-census/changelog.txt`
+Cell Census data MUST deposited under a folder named `cell-census` of the root S3 bucket:
```
atolopko-czi, line 37:
```suggestion
Cell Census data MUST be deposited under a folder named `cell-census` of the root S3 bucket:
```

Technically "folder" is "prefix", but this is interpretable as is.

Could combine the above two bullets

docs/cell_census_storage_and_release_policy.md
```diff
@@ -1,52 +1,73 @@
...
+
+The Cell Census **release** MUST be deposited in a folder named `soma`:
 
+`./cell-census/[tag]/soma/`
+
+All Cell Census **source h5ads** used to create a specific Cell Census **release** MUST be copied into a folder named `h5ads`:	
+`./cell-census/[tag]/h5ads/`
+
+
+## Cell Census release information `json`
```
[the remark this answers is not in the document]
bkmartinjr: If it matters, I intended it to be an implementation artifact *only*, and be replaced at a later date by an API surface.  This is why the API contains routines to access the same info, rather than just telling people to fetch the file.

bkmartinjr (commented):
What is the purpose of the "release policy" doc?  It combines:
* operational policy
* data access commitment for end users
* internal implementation details

Which makes future evolution of the implementation and user commits very hard to evolve when we get new or extended requirements.

Can you clarify the document intent?  I could imagine:
1. A document that describes our current operational process, subject to changes, and not written like an IETF RFC (which implies we will never change our operational processes).  Example: the release schedule is _currently_ every six months ,but that is a plan, not a long-term contractual obligation.
2. A description of internal design choices, which _will_ change
3. A means by which end-users can get "permalinks" suitable for publishing and a _promise_ to not break them
etc.

The other minor changes are all good.  It is the "policy" doc I'm confused about.

atolopko-czi:
> What is the purpose of the "release policy" doc? It combines:

I agree with these points and I think 2 documents could suffice: 1) engineering requirements (which may or may not be worth having in the repo, but seems fine if labeled as such), and 2) user-oriented doc explaining what we're committing to provide

bkmartinjr:
> user-oriented doc explaining what we're committing to provide

And on this point: I suggest we keep it as narrow as is usable/friendly/practical (from the end-user perspective), as anything else constrains our ability to evolve and progress the schema, implementation and API.

pablo-gar:
Thank you for all the comments understood the feedback and it makes sense -- too many goals combined in one doc. I will likely make this a user-facing document that has clear information to be used a data hosting requirements and I can sync up with @atolopko-czi  to look into operational policies and implementations

pablo-gar:
Addressed in new PR #314

Issue #240, opened 2023-03-06
Create a requirements.txt for notebooks


atolopko-czi:
## Motivation

* Notebook virtualenv is easier to setup
* README instructions do not need to be updated if/when package reqs change.

## Definition of Done
- [x] New api/python/notebooks/requirements.txt file contains all necessary Python pkgs for notebooks (now listed in api/python/notebooks/README.md)
- [x] api/python/notebooks/README.md updated to reference requirements.txt and not explicit packages
- [x] Cell Census Manual Build instructions (GDoc) updated to reference requirements.txt and not explicit packages

## Tasks

Obvious per DoD, above

atolopko-czi:
All tasks now done, closing!

PR #308, opened 2023-03-28
Add requirements.txt for notebooks

ebezzi:
Issue #240 

1. Add a requirements.txt for notebooks
2. Update README.md to reference this dependency files.

### DIFF
--- a/api/python/notebooks/README.md
+++ b/api/python/notebooks/README.md
@@ -26,15 +26,10 @@ as it will keep you from running out of RAM.
   $ python -m venv ./venv
   $ source ./venv/bin/activate
 ```
-2. Install the `cell_census` package using `pip`:
+2. Install the required dependencies:
 ```shell
-  $ pip install -U cell-census
+  $ pip install -r cell-census/api/python/notebooks/requirements.txt
 ```
-3. Install other third-party packages needed to run the notebooks:
-```shell
-  $ pip install scikit-misc scvi-tools
-```
-
 ## Verify your installation
 Check that your installation works - this make take a few seconds, as it loads metadata from S3:
 ```shell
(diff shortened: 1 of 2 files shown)


api/python/notebooks/requirements.txt
```diff
@@ -0,0 +1,3 @@
+cell-census
+scikit-misc 
+scvi-tools
```
atolopko-czi, line 3: I also install `jupyter`. While we might expect that jupyter is already installed for those commonly using notebooks, a pre-existing Jupyter installation outside of this venv prevent cell-census from being found in the python path.

api/python/notebooks/README.md
```diff
@@ -26,15 +26,10 @@ as it will keep you from running out of RAM.
   $ python -m venv ./venv
   $ source ./venv/bin/activate
 ```
-2. Install the `cell_census` package using `pip`:
+2. Install the required dependencies:
 ```shell
-  $ pip install -U cell-census
+  $ pip install -r cell-census/api/python/notebooks/requirements.txt
```
atolopko-czi, line 31: while `-U` is not necessary for a first-time installation, I don't think there's any harm in keeping it; if a user repeats the command later they'll get a new cell-census. I'm 50/50 on whether to keep it. :)

atolopko-czi (changes requested):
LGTM, but suggest adding `jupyter`

api/python/notebooks/requirements.txt
ebezzi: I was thinking about that. I don't think Jupyter is mandatory for running the notebooks though (for instance, I think you can also run them with vscode).

api/python/notebooks/README.md
ebezzi: Let's keep it.

PR #339, opened 2023-03-31
Python API name changes

ebezzi:


.pre-commit-config.yaml
```diff
@@ -10,8 +10,8 @@ repos:
         args: ["--config", "./api/python/notebooks/pyproject.toml"]
       - id: black
         name: black-cell-census
```
bkmartinjr, line 12: should these names change as well for clarity?

api/python/cellxgene_census/release_process.md
```diff
@@ -1,8 +1,8 @@
-# cell_census Python package release process
+# cellxgene-census Python package release process
```
bkmartinjr, line 1: underscore or dash?   We seem to mostly use underscore....

bkmartinjr: AFAIK, the only thing that should use the dash is the repo name.  Python packages get aliased, so we might as well use the "import" name, which is underscored

bkmartinjr (changes requested):
No major concerns.  The only thing thing I'd like to resolve is the use of dash vs underscore in the package name when it is referenced outside of an import statement.  Suggest we just use underscore for consistency.

bkmartinjr (approved):
LGTM

PR #340, opened 2023-03-31
name migration for tools/builder

bkmartinjr:
Changes:
* file and directory name changes:  cell_census* -> cellxgene_census*
* rename package to cellxgene_census_builder
* internal naming cleanup along the same lines

*Note to reviewers:* the `cellxgene_census_builder` intentionally imports the _old_ `cell_census` package because it wants to pull in an older version of `tiledb`, for format compatibility with old `cell_census` releases.  Over time, when we are ready to move to a more recent TileDB encoding version, this will migrate to importing `cellxgene_census`.


ebezzi (approved):
LGTM

PR #357, opened 2023-04-05
[r] use new SOMATileDBContext$context() accessor

mlin:
cf. https://github.com/single-cell-data/TileDB-SOMA/commit/8691ce07da3847eb5a31830deb30593181165738

also incidentally fix deferred FIXME regarding UTF-8 error in value_filter

### DIFF
--- a/api/r/cellxgene.census/tests/testthat/test-release_directory.R
+++ b/api/r/cellxgene.census/tests/testthat/test-release_directory.R
@@ -1,6 +1,6 @@
 # A known-good Cell Census version. This may need updating if the version used
 # is withdrawn for any reason.
-KNOWN_CENSUS_VERSION <- "2023-02-13"
+KNOWN_CENSUS_VERSION <- "2023-04-04"
 KNOWN_CENSUS_URI <- paste(
   "s3://cellxgene-data-public/cell-census/",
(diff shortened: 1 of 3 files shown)


api/r/cellxgene.census/tests/testthat/test-release_directory.R
```diff
@@ -1,6 +1,6 @@
 # A known-good Cell Census version. This may need updating if the version used
 # is withdrawn for any reason.
-KNOWN_CENSUS_VERSION <- "2023-02-13"
+KNOWN_CENSUS_VERSION <- "2023-04-04"
```
atolopko-czi, line 3: Ah! This will be less of a problem once have a LTS (5 year) release.

atolopko-czi (approved):
LGTM

api/r/cellxgene.census/tests/testthat/test-release_directory.R
bkmartinjr: Why not just designate one build to be a "long term test build", give it a tag, and not delete it?   I don't think we need a "real" LTS release for this purpose....

bkmartinjr (approved):

Issue #325, opened 2023-03-31
Update repository documents to reflect product name Census


pablo-gar:
Update the following documents and fix corresponding links:

- [x] `cell-census/README.md`
- [x] `cell-census/api/r/CellCensus/README.md`
- [x] `cell-census/api/python/notebooks/README.md`
- [x] `cell-census/api/python/cell_census/tests/README.md`
- [x] `cell-census/docs/cell_census_storage_and_release_policy.md`
- [x] `cell-census/docs/cell_census_schema.md`
   - **Document any changes to the data schema that affect the Census TileDB-SOMA object**. 

Any references to "Cell Census" should be modified accordingly to the new feature name "Census":

Non-code redaction

- CZ CELLxGENE Discover Cell Census --> CZ CELLxGENE Discover Census
- Cell Census --> Census

All other code related if and when it appears:

- CellCensus --> cellxgene.census (R package)
- cell_census --> cellxgene_census (Python package)
- cell-census --> cellxgene-census (Python package)

pablo-gar:
Re opening a because the schema was not updated. In addition the READMEs have to be re-named following this guideline for non-code text

> Non-code redaction
> 
> CZ CELLxGENE Discover Cell Census --> CZ CELLxGENE Discover Census
> Cell Census --> Census

pablo-gar:
completed

PR #374, opened 2023-04-11
Updates schema to 1.0 

pablo-gar:
Part of #325 and it also adds pending information to the schema.

All changes include:

* Updates text to reflect official name: CZ CELLxGENE Discover Census.
* Updates `census["census_info"]["summary"]` to reflect official name in the column `label`:
   * From `"cell_census_build_date"` to `"census_build_date"`.
   * From `"cell_census_schema_version"` to `"census_schema_version"`.
* Adds the following rows to `census["census_info"]["summary"]`:
   * `"census_build_git_commit_sha"`
   * `"census_r_api_version"`
   * `"census_python_api_version"`

@atolopko-czi  please take close attention to the new rows added to the `"summary"` data frame

### DIFF
--- a/docs/cellxgene_census_schema.md
+++ b/docs/cellxgene_census_schema.md
@@ -271,7 +271,7 @@ Census metadata MUST be stored as a `SOMADataFrame` with two columns:
 </tbody>
 </table>
 
-This `SOMADataFrame` MUST have the following rows::
+This `SOMADataFrame` MUST have the following rows:
  
 1. Census schema version:
 	1. label: `"census_schema_version"`
--- a/docs/cellxgene_census_schema.md
+++ b/docs/cellxgene_census_schema.md
@@ -28,7 +28,7 @@ The following terms are used throughout this document:
 
 The Census Schema follows [Semver](https://semver.org/) for its versioning:
 
-* Major: any schema changes that make the Census incompatible with public API(s) or SOMA. Examples:
+* Major: any schema changes that make the Census incompatible with the Census API or SOMA API. Examples:
 	* Column deletion in Census `obs`
 	* Addition of new modality
 * Minor: schema additions that are compatible with public API(s) and SOMA. Examples:
--- a/docs/cellxgene_census_schema.md
+++ b/docs/cellxgene_census_schema.md
@@ -320,7 +320,7 @@ An example of this `SOMADataFrame` is shown below:
     <td>2022-11-30</td>
   </tr>
   <tr>
-    <td>ccensus_build_git_commit_sha</td>
+    <td>census_build_git_commit_sha</td>
     <td>v1.0.0-9-g40266c8</td>
   </tr>
   <tr>
--- a/docs/cellxgene_census_schema.md
+++ b/docs/cellxgene_census_schema.md
@@ -1,19 +1,15 @@
-# CELLxGENE Cell Census Schema 
+# CZ CELLxGENE Discover Census Schema 
 
-**Version**: 0.1.1.
+**Version**: 1.0.0.
 
-**Last edited**: Feb, 2023.
+**Last edited**: Apr, 2023.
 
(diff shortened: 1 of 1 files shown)


docs/cellxgene_census_schema.md
```diff
@@ -307,13 +312,25 @@ An example of this `SOMADataFrame` is shown below:
...
+    <td>census_schema_version</td>
     <td>0.0.1</td>
   </tr>
   <tr>
-    <td>cell_census_build_date</td>
+    <td>census_build_date</td>
     <td>2022-11-30</td>
   </tr>
+  <tr>
+    <td>ccensus_build_git_commit_sha</td>
```
atolopko-czi, line 323:
```suggestion
    <td>census_build_git_commit_sha</td>
```

docs/cellxgene_census_schema.md
```diff
@@ -277,22 +273,31 @@ Census metadata MUST be stored as a `SOMADataFrame` with two columns:
 
 This `SOMADataFrame` MUST have the following rows::
```
atolopko-czi, line 274:
```suggestion
This `SOMADataFrame` MUST have the following rows:
```

atolopko-czi (approved):
One typo found (see suggestion), but approving.

Other considerations for discussion:
* We have some [metadata](https://github.com/chanzuckerberg/cellxgene-census/blob/284a4a478781d097e698f045e7f56bffdf989f06/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/build_soma.py#L102-L104) in the root SOMA collection, stored under the SOMA `metadata` accessor.  We have other [metadata](https://github.com/chanzuckerberg/cellxgene-census/blob/284a4a478781d097e698f045e7f56bffdf989f06/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/census_summary.py#L19-L20) stored in `census_info/summary`. Note the `CENSUS_SCHEMA_VERSION` is repeated in both places.
Do we want to store these new items only in `summary` DataFrame or in both? 
* To programmatically set the API versions, I realized we will need to do this: https://github.com/chanzuckerberg/cellxgene-census/issues/382, which I just added. But I also now realize we'll need an issue to set these new metadata, so perhaps #382 can be subsumed by that. Let me know if you want me to create the issue.

pablo-gar:
@atolopko-czi I updated as per our conversation, the only significant change to be aware from the builder perspective is:

> * Adds the following row to `census["census_info"]["summary"]`:
>   * `"dataset_schema_version"`

atolopko-czi:
> @atolopko-czi I updated as per our conversation, the only significant change to be aware from the builder perspective is:
> 
> > * Updates `census["census_info"]["summary"]` to reflect official name in the column `label`:
> >   
> >   * From `"cell_census_build_date"` to `"census_build_date"`.
> >   * From `"cell_census_schema_version"` to `"census_schema_version"`.
> > * Adds the following row to `census["census_info"]["summary"]`:
> >   
> >   * `"dataset_schema_version"`
> 
> Tracking ticket #390

Thanks, Pablo!

Also including my originally private comments for posterity, since it includes rationales for these changes:

* Revert to having just raw git SHA1 in the metadata dictionary (e.g. `g92764bb`, and not values like `v0.13.0-20-g92764bb`, which show the last tag for either [but only one] language API). This SHA1 is for engineers/troubleshooting, so it needn't be shown in the user-oriented summary table.
* Punt on having the R and Py API versions in the census summary table, as Pablo pointed out the chicken-and-egg problem of trying to use a potentially incompatible Census API version to figure out which version of the API is compatible for a given Census build.  We could instead put these API versions in the release.json file. That allows them to be somewhat more accessible, since reading this file isn't going to break (as readily). But I think we can do this later, after v1. We only need to add these fields when we know we're about to break backward compatibility from the API(s) to older census releases. And we can add them to release.json later, w/o compatibility concerns.
* Move the `cxg_schema_version` and `census_schema_version` to just the summary info table, since they're duplicated in the `metadata` dictionary of the root SOMA collection.

atolopko-czi (approved):

PR #381, opened 2023-04-13
refractor get_seurat, update docs and tests

### LINKED ISSUE #380: `get_seurat` should not have `obs_query` argument, but instead `obs_coords` and `obs_value_filter`. Same goes for `var` equivalent
## Describe the issue

The main goal for the `get_seurat` (R) and `get_anndata` (Python) is to have a easy-to-use convenience wrappers around `to_seurat` and `to_anndata` from TileDB-SOMA. To this goal we decided to hide the SOMAAxisQuery interphase.

## Expected behavior

The R `get_seurat` should mimic the design of `get_anndata` and thus should be defined as:

```r
get_seurat <- function(
    census,
    organism,
    measurement_name = "RNA",
    X_name = "raw",
    obs_value_filter = NULL, ## <---- 
    obs_coords = NULL, ## <---- 
    obs_column_names = NULL,
    var_value_filter = NULL, ## <---- 
    var_coords = NULL, ## <---- 
    var_column_names = NULL
) {
    ....
}

pablo-gar:
Fixes #380

### DIFF
--- a/.github/workflows/r-check.yml
+++ b/.github/workflows/r-check.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
           working-directory: ./api/r/cellxgene.census
-          extra-packages: any::rcmdcheck, any::styler
+          extra-packages: any::rcmdcheck, any::styler, any::roxygen2
           needs: check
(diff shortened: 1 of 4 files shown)


mlin (approved):
good catch @pablo-gar , thanks!

PR #388, opened 2023-04-18
chore: pin uberon ontology version

### LINKED ISSUE #355: TissueMapper ontology should match Dataset schema ontology version
## Motivation

The builder's UBERON (tissue) ontology obtains the ontology file directly from the 3rd party source, but is always the latest version available. This may be ahead of the version referenced in the Dataset Schema version.  The ontology file should be obtained from the Dataset Schema's "latest" UBERON version listed in https://raw.githubusercontent.com/chanzuckerberg/single-cell-curation/main/cellxgene_schema_cli/cellxgene_schema/ontology_files/owl_info.yml.

## Definition of Done

TissueMapper uses the same uberon ontology owl file as is used by the validation tool for a given Dataset Schema major version

## Tasks

- [x] Change the link at https://github.com/chanzuckerberg/cellxgene-census/blob/main/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/tissue_mapper.py#L123 
- [x] Update Schema Upgrade Process doc

atolopko-czi:
Fixes #355 

Testing is just unit tests, but these load the real ontology file, so should be sufficient.


brianraymor (approved):
The uberon release version is correct.

ebezzi (approved):

Issue #390, opened 2023-04-18
Make census builder adhere to census schema 1.0


pablo-gar:
## Description

To adhere with Census Schema version 1.0

## Tasks

In `census["census_info"]["summary"]`:

- [ ] Update value in column `label` from `"cell_census_build_date"` to `"census_build_date"`.
- [ ] Update value in column `label` from `"cell_census_schema_version"` to `"census_schema_version"` and update `value` to `"1.0.0"`
- [ ] Add  value in column `label`  `"dataset_schema_version"` and `value` `"3.0.0"`

Issue #391, opened 2023-04-18
Finalize contents for `census.metadata`


pablo-gar:
The `census.metadata` needs to be corrected in the following ways:

- [ ] key `git_commit_sha`, its value must only include the SHA, currently includes a semver prefix. For example the current one `"v0.13.0-20-g92764bb"` should be instead `"g92764bb"`
- [ ] key:value pair for the key `cxg_schema_version` must be removed as it is now included in census["census_info"]["summary"].
- [ ] key:value pair for the key `census_schema_version` must be removed as it is included in census["census_info"]["summary"].

PR #394, opened 2023-04-20
Make census builder adhere to census schema 1.0

ebezzi:
#390

### DIFF
--- a/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/census_summary.py
+++ b/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/census_summary.py
@@ -6,7 +6,7 @@ import pyarrow as pa
 import tiledbsoma as soma
 
 from .experiment_builder import ExperimentBuilder, get_summary_stats
-from .globals import CENSUS_SCHEMA_VERSION, CENSUS_SUMMARY_NAME
+from .globals import CENSUS_SCHEMA_VERSION, CENSUS_SUMMARY_NAME, CXG_SCHEMA_VERSION
 
 
 def create_census_summary(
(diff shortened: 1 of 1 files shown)


tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/census_summary.py
```diff
@@ -16,8 +16,9 @@ def create_census_summary(
 
     summary_stats = get_summary_stats(experiment_builders)
     data = [
-        ("cell_census_schema_version", CENSUS_SCHEMA_VERSION),
-        ("cell_census_build_date", build_tag),
+        ("census_schema_version", CENSUS_SCHEMA_VERSION),
```
atolopko-czi, line 19: constant also needs to be updated to `1.0.0`

atolopko-czi (changes requested):
There were a couple more changes that @pablo-gar and I agreed upon (
https://github.com/chanzuckerberg/cellxgene-census/pull/374#issuecomment-1514733540), but which I now realize did not get reflected in the issue:

* Revert to having just raw git SHA1 in the metadata dictionary (e.g. g92764bb, and not values like v0.13.0-20-g92764bb, which show the last tag for either (but only one) language API). This SHA1 is for engineers/troubleshooting, so it needn't be shown in the user-oriented summary table.
* Move the cxg_schema_version and census_schema_version to just the summary info table, since they're duplicated in the metadata dictionary of the root SOMA collection

pablo-gar:
The changes @atolopko-czi is referring to are in the this ticket https://github.com/chanzuckerberg/cellxgene-census/issues/391 which is a P1 since they are not strictly part of the schema.

If you include them, the extra points.

ebezzi:
> The changes @atolopko-czi is referring to are in the this ticket #391 which is a P1 since they are not strictly part of the schema.
> 
> If you include them, the extra points.

They are already included.

atolopko-czi (approved):
LGTM

Issue #248, opened 2023-03-09
R minimal set of vignettes are complete


pablo-gar:
The following is considered a minimal set of vignettes, mirroring the following python notebooks

- [ ][census_axis_query.ipynb](https://github.com/chanzuckerberg/cell-census/blob/main/api/python/notebooks/api_demo/census_axis_query.ipynb)
- [ ] [census_dataset_presence.ipynb](https://github.com/chanzuckerberg/cell-census/blob/main/api/python/notebooks/api_demo/census_dataset_presence.ipynb)
- [ ] [census_dataset_presence.ipynb](https://github.com/chanzuckerberg/cell-census/blob/main/api/python/notebooks/api_demo/census_dataset_presence.ipynb)
- [ ] [census_query_extract.ipynb](https://github.com/chanzuckerberg/cell-census/blob/main/api/python/notebooks/api_demo/census_query_extract.ipynb)
- [ ] [comp_bio_census_info.ipynb](https://github.com/chanzuckerberg/cell-census/blob/main/api/python/notebooks/analysis_demo/comp_bio_census_info.ipynb)
- [ ] [comp_bio_query_data_and_metadata.ipynb](https://github.com/chanzuckerberg/cell-census/blob/main/api/python/notebooks/analysis_demo/comp_bio_query_data_and_metadata.ipynb)

mlin:
Apr 24 status: 

* comp_bio_census_info is written, but blocked on a [tiledbsoma + TileDB Core issue](https://github.com/single-cell-data/TileDB-SOMA/issues/1262)
* census_data_integration (nee comp_bio_data_integration_scvi) [is also written](https://github.com/chanzuckerberg/cellxgene-census/pull/406) but not runnable in the limited RAM of GitHub Actions workers; @pablo-gar and I need to decide what to do about that
* the remaining two (census_compute_over_X and comp_bio_normalizing_full_gene_sequencing) are nice-to-haves IMO; they weren't originally on the list, and added after I suggested we do some of the fancier ones to make sure we touch all major parts of the tiledbsoma API, but I think the census_data_integration accomplishes that.

PR #406, opened 2023-04-21
Add integration vignette

pablo-gar:
partially addresses #248 

preview: https://rpubs.com/pablo-gar/1060914


mlin:
@pablo-gar Hmm I swear I left a comment here earlier this week, evidently not sorry! I suspect the new vignette is using too much memory for the GitHub Actions worker (which has only 7GB RAM). Do you have any feel for what it uses locally?

We can either scale down the problem size in the vignette to use less memory, or forego building this one as part of the CI workflow. (I don't think we build the Python notebooks in CI, do we?) WDYT?

mlin:
@pablo-gar just following up, is https://github.com/single-cell-data/TileDB-SOMA/pull/1274/files maybe going to be helpful for this?

mlin (approved):
I ran it locally, looks great! We'll need to move the file around depending which way we decide to go on #583, but we can handle that in either order easily.

pablo-gar:
Merged via  #583

PR #422, opened 2023-04-27
Docsite GHA workflow

ebezzi:
Creates a GHA workflow for building and releasing the docsite automatically.


atolopko-czi (approved):
👍 CR'd live w/ @ebezzi

PR #423, opened 2023-04-27
Modify README.md to remove pre-release status

ebezzi:
In preparation of Python v1.0 release. This text is shown on pypi.org cellxgene-census package page.

PR #424, opened 2023-04-27
chore: remove prerelease text from readme

atolopko-czi:
In preparation of Python v1.0 release. This text is shown on pypi.org cellxgene-census package page.


atolopko-czi:
dup PR w/ #423

PR #426, opened 2023-04-28
[WIP] builder experiment: make all soma_joinid start from 1

mlin:


mlin:
Closing this since it's fortunately not needed, but, the code diff may be of future interest

PR #451, opened 2023-05-02
A few final edits to gget tutorial

lauraluebbert:
Hi all, sorry for the second PR. 

I noticed that the Google Colab link was not working anymore since the name and location of the notebook changed, so I edited the following:

- Fixed (hopefully) Google Colab link at the top of notebook
- Noted the gget version when the cellxgene module was first added (under ## Install)
- Added citation guidelines to the first markdown cell
- The text still mentioned an argument that was deprecated (`anndata=False`), so I corrected this to `meta_only=True`


pablo-gar (approved):
LGTM

PR #472, opened 2023-05-10
Bump static census version in R tests

ebezzi:
The `2023-04-04` version that was pinned in the tests is expired, so this test is now failing. When an LTS version is available, we should switch to that.


atolopko-czi (approved):
LGTM & ty for the comment!

PR #481, opened 2023-05-12
Add docsite version number from the library version

### LINKED ISSUE #461: Sphinx should get the sidebar version from the package
The current docsite sidebar pulls the version from a variable hardcoded in conf.py. Instead, it should infer it from the repo. [This SO thread](https://stackoverflow.com/questions/26141851/let-sphinx-use-version-from-setup-py) has some good ideas.

ebezzi:
Fixes #461 

Note that this will only work for tagged releases, otherwise it will add a development version number.


api/python/notebooks/api_demo/census_gget_demo.ipynb
atolopko-czi, line 1: Is this file intended for the PR?

ebezzi: No, thanks for the catch. Removing this.

atolopko-czi (approved):
LGTM

PR #483, opened 2023-05-16
Updated gget cellxgene tutorial to reflect workflow updates

lauraluebbert:
Hi all, 

There were some requirement clashes between `cellxgene-census` and other `gget` dependencies (especially tiledbsoma was causing some problems depending on the environment). Hence, I removed the `cellxgene-census` package from the standard `gget` requirements and added it as an optional installation to `gget setup` (similar to the `gget alphafold` workflow). 

In this PR, I adjusted the gget tutorial as follows:
- Add `gget.setup("cellxgene")` command to install optional dependencies
- Changed named version of gget to 0.25.7 (since 0.25.6 was yanked)
- Moved Google Colab link into text since the button does not work on the Census website

FYI: The new version of `gget cellxgene` now also includes the `census_version` argument.

Best,

Laura


pablo-gar (changes requested):
@lauraluebbert Thanks for the changes! This looks great to me.

Would you mind re-running the notebook? we just released the "stable" data release and therefore this warning will look different moving forward:

image

pablo-gar (approved):
Approved! Note that the change will be reflected in next pypi release. We will likely cut it the coming week to incorporate this change

pablo-gar:
Will be merged in #490 to pass `black` tests.

PR #488, opened 2023-05-18
Add Databricks install instructions to FAQ

### LINKED ISSUE #389: Investigate installing Python API on Databricks cluster with ML runtime
Installing `pip install -U cellxgene-census=='0.13.0'` into a Databricks environment _sometimes_ requires installing `pyarrow>=9.0.0`. Databricks ML runtime comes with `pyarrow==7.0.0` installed.

`%pip install pyarrow>=9.0.0` must be run to correct the problem. Note the leading `%`, which "install the package in the virtual environment where the current notebook kernel is running". Using `!pip` does not work.

We need to:
* Understand the root cause of the pip installation problem, and why pyarrow 9.0.0 isn't being installed or found.
* Provide clear instructions on how to install `cellxgene-census` in a Databricks environment, if it requires anything out of the ordinary.

atolopko-czi:
Fixes #389

### DIFF
--- a/docs/cellxgene_census_docsite_FAQ.md
+++ b/docs/cellxgene_census_docsite_FAQ.md
@@ -152,6 +152,3 @@ See also:
 * https://community.databricks.com/s/question/0D53f00001GHVP3CAP/whats-the-difference-between-magic-commands-pip-and-sh-pip
 
 Alternately, you can configure your cluster to install the `cellxgene-census` package each time it is started by adding this package to the "Libraries" tab on the cluster configuration page per these [instructions](https://docs.databricks.com/libraries/cluster-libraries.html).
-
-
-
--- a/docs/cellxgene_census_docsite_FAQ.md
+++ b/docs/cellxgene_census_docsite_FAQ.md
@@ -16,7 +16,7 @@ Last updated: May, 2023.
 - [How can I ask for new features?](#how-can-i-ask-for-new-features)
 - [How can I contribute my data to the Census?](#how-can-i-contribute-my-data-to-the-census)
 - [Why do I get an `ArraySchema` error when opening the Census?](#why-do-i-get-an-arrayschema-error-when-opening-the-census)
-- [Why do I get an error when running `import cellxgene_census` on Databricks?](#why-do-i-get-an-error-when-running-import-cellxgene_census-on-databricks)
+- [Why do I get an error when running `import cellxgene_census` on Databricks?](#why-do-i-get-an-error-when-running-import-cellxgene-census-on-databricks)
 
 ## Why should I use the Census?
 
--- a/docs/cellxgene_census_docsite_FAQ.md
+++ b/docs/cellxgene_census_docsite_FAQ.md
@@ -1,6 +1,6 @@
 # FAQ
 
-Last updated: Apr, 2023.
+Last updated: May, 2023.
 
 - [Why should I use the Census?](#why-should-i-use-the-census)
 - [What data is contained in the Census?](#what-data-is-contained-in-the-census)
--- a/docs/cellxgene_census_docsite_FAQ.md
+++ b/docs/cellxgene_census_docsite_FAQ.md
@@ -139,11 +139,11 @@ This can occur if the `cellxgene_census` Python package is installed in a Databr
 
 To fix, simply install using one of the following Databricks notebook "magic" commands:
 ```
-pip install cellxgene-census
+pip install -U cellxgene-census
 ```
 or
 ```
-%pip install cellxgene-census
+%pip install -U cellxgene-census
 ```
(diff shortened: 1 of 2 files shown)


docs/cellxgene_census_docsite_FAQ.md
```diff
@@ -1,12 +1,12 @@
...
 
-Last updated: Apr, 2023.
+Last updated: May, 2023.
 
 - [Why should I use the Census?](#why-should-i-use-the-census)
 - [What data is contained in the Census?](#what-data-is-contained-in-the-census)
 - [How do I cite the use of the Census for a publication?](#how-do-i-cite-the-use-of-the-census-for-a-publication)
 - [Why does the Census not have a normalized layer or embeddings?](#why-does-the-census-not-have-a-normalized-layer-or-embeddings)
-- [How does the Census differentiate from other services?](#how-does-the-census-differentiate-from-other-services)
+- [How does the Census differentiate from other tools?](#how-does-the-census-differentiate-from-other-tools)
```
atolopko-czi, line 9: unrelated fix (hyperlink)

docs/cellxgene_census_docsite_FAQ.md
```diff
@@ -132,3 +133,22 @@ You may get this error if you are trying to open a Census data build with an old
...
+
+This can occur if the `cellxgene_census` Python package is installed in a Databricks notebook using `%sh pip install cellxgene_census`. This command does _not_ restart the Python process after installing `cellxgene_census` and any pip package dependencies that were pre-installed by the Databricks Runtime environment but upgraded for `cellxgene_census` will not be reloaded with their new version. You may see `numba` or `pyarrow` related errors, for example.
+
+To fix, simply install using one of the following Databricks notebook "magic" commands:
+```
+pip install cellxgene-census
+```
+or
+```
+%pip install cellxgene-census
```
ainfeld, line 146: I think this version still needs the  `-U` so I would adjust to be `%pip install -U cellxgene-census`

ainfeld (changes requested):
After that one small change, everything else LGTM!

docs/cellxgene_census_docsite_installation.md
```diff
@@ -25,6 +25,8 @@ Install the `cellxgene-census` package via pip:
 pip install -U cellxgene-census
 ```
 
+If installing in a Databricks notebook environment, use `%pip install`. Do not use `%sh pip install`. See the [FAQ](cellxgene_census_docsite_FAQ.md#why-do-i-get-an-error-when-running-import-cellxgene_census-on-databricks).
```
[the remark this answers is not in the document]
ebezzi: I can test it via a local build.

docs/cellxgene_census_docsite_FAQ.md
```diff
@@ -16,6 +16,7 @@ Last updated: Apr, 2023.
 - [How can I ask for new features?](#how-can-i-ask-for-new-features)
 - [How can I contribute my data to the Census?](#how-can-i-contribute-my-data-to-the-census)
 - [Why do I get an `ArraySchema` error when opening the Census?](#why-do-i-get-an-arrayschema-error-when-opening-the-census)
+- [Why do I get an error when running `import cellxgene_census` on Databricks?](#why-do-i-get-an-error-when-running-import-cellxgene_census-on-databricks)
```
ebezzi, line 19:
```suggestion
- [Why do I get an error when running `import cellxgene_census` on Databricks?](#why-do-i-get-an-error-when-running-import-cellxgene-census-on-databricks)
```

ebezzi (changes requested):

docs/cellxgene_census_docsite_FAQ.md
atolopko-czi: My (questionable) understanding is that `-U` only upgrades the explicitly listed packages. Any transitive dependencies (e.g. numba) _will_ be upgraded _if_ the explicit packages are upgraded _and_ their dependencies require newer versions. So the transitive dependencies do get updated, as needed, even without `-U`.

I think the `-U` is only important if you're in the same notebook, a new cellxgene-census is released while it's active, and then you want to upgrade the cellxgene_census.

All that said, I'm happy to just include `-U` since my understanding may be wrong, and it's a good habit to include! 😄

ainfeld: Oh interesting, when I tried `%pip install cellxgene-census` without the -U in my current notebook, it gave me an error

docs/cellxgene_census_docsite_installation.md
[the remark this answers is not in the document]
atolopko-czi: ty! it's tough being a human.

docs/cellxgene_census_docsite_FAQ.md
atolopko-czi: Curious, what's the error? same one? I may learn something...

ainfeld: Just tried re-running it and this time I didn't get an error, so I'm not sure now  😄

ainfeld (approved):
LGTM

PR #501, opened 2023-05-24
run experimental tests conditionally

atolopko-czi:
This is a sub-branch on `atol/pytorch-dataloader`, to segregate these changes in a separate PR for easier review.

Changes:
- pytest can now also be run with `--experimental` option to explicitly enable tests on the `experimental` sub-package (ie in addition to the `--expensive` option)
- GHA build only runs experimental tests for Python 3.10 to avoid some missing experimental code pkg dependencies for lower Python versions (torch==2.0.1 not avail for Python 3.7)
- fix bug where `pytest -m 'not live_corpus' --experimental` not working

### DIFF
--- a/.github/workflows/py-unittests.yml
+++ b/.github/workflows/py-unittests.yml
@@ -37,11 +37,11 @@ jobs:
       - name: Test with pytest (API)
         if: matrix.python-version != '3.10'
         run: |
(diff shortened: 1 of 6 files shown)


ebezzi (approved):

bkmartinjr (approved):
Brilliant!

Issue #482, opened 2023-05-15
[r] Replicate open_soma() user warnings in R


atolopko-czi:
See Python changes to add have open_soma() default to `stable` and added user warnings: #435

ebezzi:
Make sure that the vignette api/r/cellxgene.census/vignettes/comp_bio_census_info.Rmd is also updated to reflect this change. At line 31, 

>The `cellxgene.census` R package  contains a convenient `open_soma()` API to open any version of the Census (the latest by default).

should be (roughly)

> The `cellxgene.census` R package  contains a convenient `open_soma()` API to open any version of the Census (the stable by default).

mlin:
I'll take care of this once the `stable` key exists in the `release.json`, if that's alright with everyone

pablo-gar:
That sounds good!

PR #435, opened 2023-04-28
open_soma() alias handling changes

atolopko-czi:
Resolves #433 

- default to stable release
- notify user when an alias is used
- notify user if stable alias is not defined

New output looks like:

```
import cellxgene_census as cc
c = cc.open_soma()
The "stable" Census version is not yet available. Using "latest" Census version instead.
The "latest" release is currently 2023-04-25. Specify 'census_version="2023-04-25"' in future calls to open_soma() to ensure data consistency.

c = cc.open_soma(census_version="stable")
The "stable" Census version is not yet available. Using "latest" Census version instead.
The "latest" release is currently 2023-04-25. Specify 'census_version="2023-04-25"' in future calls to open_soma() to ensure data consistency.

c = cc.open_soma(census_version="latest")
The "latest" release is currently 2023-04-25. Specify 'census_version="2023-04-25"' in future calls to open_soma() to ensure data consistency.

c = cc.open_soma(census_version="2023-04-25")
# silence is bliss

c = cc.open_soma(census_version="bad")
Traceback (most recent call last):
  File "/Users/atolopko/venvs/cellxgene-census/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-7-16de8ccb18df>", line 1, in <module>
    c = cc.open_soma(census_version="bad")
  File "/Users/atolopko/dev/cxg/cell-census/cellxgene-census/api/python/cellxgene_census/src/cellxgene_census/_open.py", line 129, in open_soma
    raise ValueError(
ValueError: The "bad" Census version is not valid. Use get_census_version_directory() to retrieve available versions.

c = cc.open_soma(census_version=None)
Traceback (most recent call last):
  File "/Users/atolopko/venvs/cellxgene-census/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-8-c137ddd80f7f>", line 1, in <module>
    c = cc.open_soma(census_version=None)
  File "/Users/atolopko/dev/cxg/cell-census/cellxgene-census/api/python/cellxgene_census/src/cellxgene_census/_open.py", line 116, in open_soma
    raise ValueError("Must specify either a census version or an explicit URI.")
ValueError: Must specify either a census version or an explicit URI.
```

PR #502, opened 2023-05-25
[r] use `stable` by default & add alias resolution message

mlin:
#482

For parity with python #435

Also adapts to `matrixZeroBasedView` refactoring in https://github.com/single-cell-data/TileDB-SOMA/pull/1411


api/r/cellxgene.census/tests/testthat/test-open.R
```diff
@@ -1,5 +1,5 @@
 test_that("open_soma", {
-  coll <- open_soma("latest")
+  coll <- open_soma("stable")
```
ebezzi, line 2: I think outside of the specific `open_soma_stable` test, we should rely on latest, since that would catch potential issues with the built version earlier. It's also how we do it in Python. @atolopko-czi thoughts?

atolopko-czi (approved):
LGTM. Appears to have feature parity with related Python implementation.

PR #505, opened 2023-05-25
Support loading configuration from env vars

ebezzi:
Since AWS Batch is most easily configured via env vars, this PR allows reading the configuration from them. Note that this only checks if a key in the default configuration is present as an env var and, if so, replaces the value with it. Recommendations for alternative solutions are appreciated.

### DIFF
--- a/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
+++ b/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
@@ -108,6 +108,9 @@ class CensusBuildConfig(Namespace):
 
     def __init__(self, **kwargs: Any):
         config = self.defaults.copy()
+        for k in config.keys():
+            if k.upper() in os.environ:
+                config[k] = os.environ[k.upper()]
         config.update(kwargs)
         super().__init__(**config)
 


tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
```diff
@@ -108,6 +108,9 @@ class CensusBuildConfig(Namespace):
 
     def __init__(self, **kwargs: Any):
         config = self.defaults.copy()
+        for k in config.keys():
+            if k.upper() in os.environ:
+                config[k] = os.environ[k.upper()]
```
atolopko-czi, line 113: One consideration is that the values in the config are typed, and are not always strings. So we may need a more formal specification of the config params (names and types) to do conversions. We could do that if we used `attrs` package and made use of its [converters](https://www.attrs.org/en/stable/init.html#converters).

bkmartinjr: Excellent catch.

I think Andrew's suggestion is a good one.  We need type casting for many of the values.  Attrs would be a very clean framework that could easily replace the built-in Namespace and MutableNamespace classes (just use the [`frozen`](https://www.attrs.org/en/stable/how-does-it-work.html#immutability) feature in `attrs` for the immutable namespace).

That would also allow the use of attr [validators](https://www.attrs.org/en/stable/examples.html#validators) for any types where that was useful.

ebezzi: I agree, this needs more work. For running this in prod, we won't need this PR anymore, so we can close it. We can create a ticket to do this in a more engineered way using `attrs`.

ebezzi:
Replaced by https://github.com/chanzuckerberg/cellxgene-census/pull/552

Issue #500, opened 2023-05-23
PyTorch DataLoader documentation


atolopko-czi:
Add package, class, method docs, and README

Ensure that docs are visible in published online docs.

PR #509, opened 2023-05-30
chore: update lifecycle tags

### LINKED ISSUE #508: Update lifecycle tags
- Existing Python API classes/methods updated to "maturing"
- New `experimental` sub-pkg classes/methods set to "experimental".

Do not release to pypi until tiledbsoma version updated to > `1.2.4`, as this will include similar lifecycle tag updates to the SOMA APIs.

atolopko-czi:
* Changes existing Python API methods' lifecycle tags to `maturing`
* Sets `experimental` package methods' lifecycle tags to `experimental`
* Adds some documentation to `experimental` public classes and methods, with proper formatting for github pages output.
* Exports public names for `experimental.ml` package

Resolves #508 

Partially addresses: #500

### DIFF
--- a/api/python/cellxgene_census/src/cellxgene_census/_experiment.py
+++ b/api/python/cellxgene_census/src/cellxgene_census/_experiment.py
@@ -31,7 +31,7 @@ def _get_experiment(census: soma.Collection, organism: str) -> soma.Experiment:
         ValueError: if unable to find the specified organism.
 
     Lifecycle:
-        Maturing.
+        maturing
 
     Examples:
 
--- a/api/python/cellxgene_census/src/cellxgene_census/_experiment.py
+++ b/api/python/cellxgene_census/src/cellxgene_census/_experiment.py
@@ -31,7 +31,7 @@ def _get_experiment(census: soma.Collection, organism: str) -> soma.Experiment:
         ValueError: if unable to find the specified organism.
 
     Lifecycle:
-        Experimental.
+        Maturing.
 
     Examples:
 
--- a/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
+++ b/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
@@ -34,6 +34,13 @@ pytorch_logger.setLevel(logging.INFO)
 
 @attrs
 class Stats:
+    """
+    Statistics about the data retrieved by ExperimentDataPipe from TileDB-SOMA.
+
+    Lifecycle:
+        Maturing.
+    """
+
     n_obs: int = 0
     """The total number of obs rows retrieved"""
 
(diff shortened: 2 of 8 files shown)


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -252,6 +262,20 @@ def __init__(
         num_workers: int = 0,
         soma_buffer_bytes: Optional[int] = None,
     ) -> None:
+        """
+        Construct a new ``ExperimentDataPipe``.
+
+        Args:
```
ebezzi, line 268: Recommend removing the empty annotations, as they will likely mess up the docsite.

ebezzi (approved):
LGTM, suggest removing the empty annotations, although it's more of a nitpick since I don't believe we'll generate docpages for the experimental module yet.

api/python/cellxgene_census/src/cellxgene_census/_experiment.py
```diff
@@ -31,7 +31,7 @@ def _get_experiment(census: soma.Collection, organism: str) -> soma.Experiment:
         ValueError: if unable to find the specified organism.
 
     Lifecycle:
-        Experimental.
+        Maturing.
```
bkmartinjr, line 34: the style used in SOMA is lower case, no punctuation

Eg.,
```
Lifecycle: maturing
```

or

```
Lifecycle:
    maturing
```

bkmartinjr: I prefer the first, as it is much easier to search for.  Not sure if it will mess up online docsite

bkmartinjr (approved):
one style comment but it is a nit....

api/python/cellxgene_census/src/cellxgene_census/_experiment.py
atolopko-czi: @ebezzi do you know how each will render?

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
atolopko-czi: done

api/python/cellxgene_census/src/cellxgene_census/_experiment.py
atolopko-czi: I can't find any cases of same-line header+content in our docstrings, so conservatively went with the multi-line form for now.

ebezzi: We should use the multi-line. Regardless, this renders as a separate paragraph:
image
so I think the capitalization and dot makes it look better.

ebezzi: To add a bit more context, [this](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) is the parser we're using for our Google-style docstrings - I believe that without the multiline, the lifecycle section will render as a normal paragraph and not with a section title+paragraph, which will IMHO make the generated docstring page less readable. See [an example](https://tiledbsoma.readthedocs.io/en/latest/_autosummary/tiledbsoma.AxisQuery.html) here.

atolopko-czi: Useful context! I'm going with a compromise: multi-line, but lowercase and no punc, to make everyone equally slightly unhappy. 😆 I'lllet @pablo-gar be the arbitor of docstring fashion upon his return and we can change on a future PR.

PR #512, opened 2023-05-31
minor clarifications for the pypi.org release process

atolopko-czi:

PR #513, opened 2023-05-31
fix errant option in release_process.md

atolopko-czi:
Follow-up fix for #512


api/python/cellxgene_census/release_process.md
```diff
@@ -102,4 +102,16 @@ To publish built release assets to PyPi (_note_: this will require your pypi/tes
 3. Use twine to upload to PyPi (this assumes the downloaded assets are in ./artifact/), e.g.,
    ```shell
    pipx run twine upload ./artifact/*
+<<<<<<< Updated upstream
```
bkmartinjr, line 105: this looks like it needs to be hand merged :-)

atolopko-czi:
me brain dead

PR #516, opened 2023-06-02
fix pytorch multiprocessing result

### LINKED ISSUE #515: ExperimentDataPipe returns first partition repeatedly in multiprocessing mode
## Describe the bug

`ExperimentDataPipe` returns the first partition of data repeatedly from each worker when used in multiprocessing mode. 

## To Reproduce

Pseudo-code:
```
dp = ExperimentDataPipe(...)
dl = experiment_dataloader(dp, num_workers=2)
list(iter(dp))
```

## Expected behavior

The full result returned by the DataLoader should be the concatenated result of each worker process.

## Environment

N/A

## Additional context

Not user reported (yet).

atolopko-czi:
Fixes #515

### DIFF
--- a/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
+++ b/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
@@ -3,7 +3,7 @@ import os
 import sys
 from datetime import timedelta
 from time import time
-from typing import Any, Dict, Iterator, Optional, Sequence, Tuple
+from typing import Any, Dict, Iterator, List, Optional, Sequence, Tuple
 
 import numpy as np
 import pandas as pd
@@ -24,7 +24,7 @@ from torch.utils.data.dataset import Dataset
 import cellxgene_census
 from cellxgene_census._open import _build_soma_tiledb_context
 
-ObsDatum = Tuple[Tensor, torch.sparse_coo_tensor]
+ObsDatum = Tuple[Tensor, Tensor]
 
 Encoders = Dict[str, LabelEncoder]
 
@@ -96,7 +96,7 @@ class _ObsAndXIterator(Iterator[ObsDatum]):
 
     def __init__(
         self,
-        obs_joinids: pa.Array,
+        obs_joinids: List[int],
         var_joinids: pa.Array,
         exp_uri: str,
         aws_region: Optional[str],
@@ -109,7 +109,6 @@ class _ObsAndXIterator(Iterator[ObsDatum]):
         sparse_X: bool,
         soma_buffer_bytes: Optional[int] = None,
     ) -> None:
-        self.obs_joinids = obs_joinids
         self.var_joinids = var_joinids
         self.batch_size = batch_size
         self.sparse_X = sparse_X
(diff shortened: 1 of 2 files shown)


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -329,15 +325,24 @@ def _init(self) -> None:
             partition_size = len(obs_joinids) // num_partitions
             partition_start = partition_size * partition
             partition_end_excl = min(len(obs_joinids), partition_start + partition_size)
-            self._obs_joinids_partitioned = obs_joinids[partition_start:partition_end_excl]
+            self._obs_joinids_partitioned = obs_joinids[partition_start:partition_end_excl].tolist()
```
atolopko-czi, line 328: This is the fundamental fix. For some reason that I cannot explain, holding on to a PyArrow array causes SOMA Experiment reader to return the wrong result for some workers. This suggests shared memory is being used for this array across worker processes. This array is being provided by SOMA, so we cannot do anything to change this, other than to make a safe copy of the data.

ebezzi: That's interesting. Maybe worth inquiring in the SOMA chat to see if someone knows? I can see this being a cause of other bugs.

ebezzi (approved):

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
atolopko-czi: I've narrowed it down to it being the difference between these two calls in tiledbsoma:
* `tiledbsoma.libtiledbsoma.SOMAReader.set_dim_points_arrow`
* `tiledbsoma.libtiledbsoma.SOMAReader.set_dim_points_int64`
which are called when setting the coords on an (obs) DataFrame in this elided stack trace:
```
apis/python/src/tiledbsoma/_dataframe.py:268
SOMA/apis/python/src/tiledbsoma/_dataframe.py:410
```

The child processes are being `spawn`ed. And even though each child process is opening the Experiment anew, the obs soma_joinids being passed into `obs.read(coords=...)` are not being respected. I have confirmed they are in fact different sets of ids for each child process. But the obs *result* is the same for each child process. This only happens when the `coords` arg is a PyArrow array, not when it is a list. So it seems as if the coords being passed in the arrow case are being stored in shared memory across the processes and clobbering one another.

I'll have to come back to this to pull out a simpler, reproducible case. So recording here where I've left off on this investigation. 

I think it's okay to merge the fix for PyTorch, even though the underlying cause is not fully understood.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -239,9 +238,7 @@ class ExperimentDataPipe(pipes.IterDataPipe[Dataset[ObsDatum]]):  # type: ignore
     """In multi-processing mode (i.e. num_workers > 0), this ``ExperimentAxisQuery`` object will *not* be pickled; 
     each worker will instantiate a new query"""
 
-    _obs_joinids_partitioned: Optional[pa.Array]
-
-    _obs_and_x_iter: Optional[_ObsAndXIterator]
+    _obs_joinids_partitioned: Optional[List[int]] = None
```
bkmartinjr, line 241: type signature needs to change

bkmartinjr (approved):
There is a type signature that needs to change.  Suggests that mypy may not be running on this code?

Other than that, it looks fine.

atolopko-czi:
> There is a type signature that needs to change. Suggests that mypy may not be running on this code?

Good catch! Interesting, since mypy was run on both local pre-commit and in GHA, but not caught. Perhaps pyarrow.Table.to_numpy() does not have typing hints.

PR #407, opened 2023-04-23
[r] Add comp_bio_census_info.Rmd

mlin:
Furthering https://github.com/chanzuckerberg/cellxgene-census/issues/248

This had been blocked for awhile on: https://github.com/single-cell-data/TileDB-SOMA/issues/1262 now fixed

PREVIEW: https://gistpreview.github.io/?85f31f2c190c4fa4e99edbebc0eae15c

PR #519, opened 2023-06-06
Advance r-universe branch to current main

mlin:
With #407 there's a good chance our r-universe build will work.


mlin:
@ebezzi @pablo-gar This PR is trying to merge `main` *into* the `r-universe` branch which is currently very stale, that's why it has a lot of commits. However, I notice that our branch protection is set up such that we can only squash commit, which (while a good idea for the typical merging of a working branch into main) isn't quite what we want for this case of release branch that we want to advance to match main from time to time. Are you able to adjust the branch protection features for this special case release branch specifically?

ebezzi:
> @ebezzi @pablo-gar This PR is trying to merge `main` _into_ the `r-universe` branch which is currently very stale, that's why it has a lot of commits. However, I notice that our branch protection is set up such that we can only squash commit, which (while a good idea for the typical merging of a working branch into main) isn't quite what we want for this case of release branch that we want to advance to match main from time to time. Are you able to adjust the branch protection features for this special case release branch specifically?

Unfortunately GitHub only allows to enable/disable merge mode globally, not per branch. I think what we can do is to create a GHA workflow that will forcibly replace the `r-universe` branch with `main` on demand, and we can trigger this workflow using a workflow_dispatch. Does that sound reasonable?

pablo-gar (approved):

mlin:
Thanks @ebezzi -- I'll merge this for the moment (right now just trying to see whether our r-universe build works) and break out an issue about automating it including your idea.

Issue #89, opened 2023-01-24
Remove explicit storage of source h5ad files as cell census data artifacts


atolopko-czi:
Depends on https://github.com/chanzuckerberg/cell-census/issues/6

Once the Cell Census metadata includes references to persistent H5AD files, maintained by CELLxGENE Data Portal, the builder no longer needs to store the source H5ADs as part of the Cell Census data.

pablo-gar:
Does this have any implications on API functionality that fetches source h5ads? `cell_census.get_source_h5ad_uri`

bkmartinjr:
> Does this have any implications on API functionality that fetches source h5ads? cell_census.get_source_h5ad_uri

Yes, it will require that we revise / release a new cell_census package, as the current API implementation has hard-wired assumptions about the data location.

However, this change should be backward compatible, and not require end-user code changes (just a package version update)

brianraymor:
> it should be deprecated in favor of the Discover API that lets a user fetch an H5AD given a dataset_id/dataset_version_id pair.

Clarified in slack that the Discover API is not needed for this scenario. The request is a simple GET from CloudFront based on the dataset version uuid.

bkmartinjr:
Obsoleted by #935 - will not fix.

PR #552, opened 2023-06-16
Mirrors support

### LINKED ISSUE #546: The APIs can select geographical location of Census Data
blocked by #545 

## Product requirements

- The user can quickly know what geographical locations are available (I defer to eng on the design, e.g. only docs, API functionality, etc)
- Upon opening the Census with the API `open_soma` the user can select any available geographical locations.

## Design

To be defined by eng @metakuni

ebezzi:
Fixes #546


tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -52,17 +52,24 @@ def do_build(args: CensusBuildArgs, skip_completed_steps: bool = False) -> int:
...
+            do_prebuild_set_defaults,
+            do_prebuild_checks,
+            do_build_soma,
+            do_validate_soma,
+            do_create_reports,
+            do_data_copy,
+            do_the_release,
+            do_report_copy,
+            do_old_release_cleanup,
+            do_log_copy,
```
ebezzi, line 71: This needs to move here, since it is now part of the build proper. Sub-steps will not require to copy logs.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -221,6 +225,18 @@ def do_log_copy(args: CensusBuildArgs) -> bool:
     return True
 
 
+def do_sync_to_external_s3_bucket(args: CensusBuildArgs) -> bool:
```
ebezzi, line 228: The naming here isn't great. Will revisit.

tools/cellxgene_census_builder/src/cellxgene_census_builder/host_validation.py
```diff
@@ -70,28 +70,25 @@ def check_host(args: CensusBuildArgs) -> bool:
 
     return (
         check_os()
-        and check_physical_memory(
-            args.config.get("min_physical_memory", args.config.host_validation_min_physical_memory)
```
ebezzi, line 74: Not sure where these args were coming from. I searched and they don't appear anywhere in the project.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
```diff
@@ -20,38 +20,70 @@
...
+    dryrun: bool = field(
+        converter=bool, default=False
+    )  # if True, will disable copy of data/logs/reports/release.json to S3 buckets. Will NOT disable local build, etc.
     #
     # Paths and census version name determined by spec.
-    "cellxgene_census_S3_path": "s3://cellxgene-data-public/cell-census",
-    "logs_S3_path": "s3://cellxgene-data-public-logs/builder",
-    "build_tag": datetime.now().astimezone().date().isoformat(),
+    cellxgene_census_S3_path: str = field(default="s3://cellxgene-data-public/cell-census")
+    cellxgene_census_S3_remote_path: str = field(default=None)
```
ebezzi, line 39: This one could be: `cellxgene_census_S3_replica_path`

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
```diff
@@ -20,38 +20,70 @@
...
-    "verbose": 1,
-    "log_dir": "logs",
-    "log_file": "build.log",
-    "reports_dir": "reports",
-    "consolidate": True,
-    "disable_dirty_git_check": True,
-    "dryrun": False,  # if True, will disable copy of data/logs/reports/release.json to S3 buckets. Will NOT disable local build, etc.
+
+
+@define
```
bkmartinjr, line 25: what is our coding convention around fully naming short / ambiguous names like this?  I personally prefer @attrs.define, but I will happily follow group conventions.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
bkmartinjr: I prefer replica, but not strongly.

bkmartinjr, line 39: Looking forward, we may very well have multiple replicas.  May want to future-proof now if not expensive to build & test.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
ebezzi: The attrs doc used this convention, so I adopted it, but I am not very opinionated.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
```diff
@@ -20,38 +20,70 @@
...
     #
     # Host minimum resource validation
-    "host_validation_disable": False,  # if True, host validation checks will be skipped
-    "host_validation_min_physical_memory": 512 * 1024**3,  # 512GiB
-    "host_validation_min_swap_memory": 2 * 1024**4,  # 2TiB
-    "host_validation_min_free_disk_space": 1 * 1024**4,  # 1 TiB
+    host_validation_disable: int = field(
+        converter=bool, default=False
+    )  # if True, host validation checks will be skipped
+    host_validation_min_physical_memory: int = field(converter=int, default=512 * 1024**3)  # 512Gi
```
bkmartinjr, line 51: typo in comment:  "GiB"  (missing B)

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
ebezzi: I was thinking that the replicas would be managed by separate SFN steps, and each one will set a different value of `cellxgene_census_S3_remote_path`. This way we can easily run them in parallel.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
```diff
@@ -20,38 +20,70 @@
...
+    host_validation_min_swap_memory: int = field(converter=int, default=2 * 1024**4)  # 2Ti
```
bkmartinjr, line 52: typo in comment - missing B.  Or maybe this is intentional given that they all seem to have changed?   Needs units, so B seems important :-)

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
```diff
@@ -20,38 +20,70 @@
...
+    host_validation_min_free_disk_space: int = field(converter=int, default=1 * 1024**4)  # 1 Ti
     #
     # Release clean up
-    "release_cleanup_days": 32,  # Census builds older than this are deleted
+    release_cleanup_days: int = field(converter=int, default=32)  # Census builds older than this are delete)
```
bkmartinjr, line 56: typo:   s/delete/deleted/

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
ebezzi: Nah, I messed up with the editor's multi-cursor. Will fix them accordingly.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
```diff
@@ -20,38 +20,70 @@
...
+        if not isinstance(user_config, dict):
+            raise TypeError("YAML config file malformed - expected top-level dictionary")
+
+        return cls(**user_config)
+
+    @classmethod
+    def load_from_env_vars(cls) -> Self:
+        config = cls()
+        for fld in fields(CensusBuildConfig):
+            if fld.name.upper() in os.environ:
```
bkmartinjr, line 84: Question: do we want to prefix the ENV vars with something so there is less chance for conflict with other packages?   Names like VERBOSE are pretty common.

E.g., CENSUS_BUILD_VERBOSE?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
bkmartinjr: most of the Builder uses the other convention. Given the diversity out there, I feel we should have a convention in our code base.

Happy to do whatever is best, but one convention seems wise.  I didn't see a good default from Google Py conventions.   And I don't see any reason to hold up this PR for the discussion - just wanted to raise the change as notable given the rest of the code base.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
bkmartinjr: gotcha. I now understand!

tools/cellxgene_census_builder/src/cellxgene_census_builder/data_copy.py
```diff
@@ -21,13 +21,30 @@ def sync_to_S3(from_path: Union[str, pathlib.PosixPath], to_path: str, dryrun: b
...
+    cmd = ["aws", "s3", "sync", from_path, to_path, "--no-progress"]
     if dryrun:
         cmd += ["--dryrun"]
 
     returncode = -1
     try:
-        _log_it(f"Starting copy {from_path.as_posix()} -> {to_path}", dryrun)
+        _log_it(f"Starting copy {from_path} -> {to_path}", dryrun)
         with subprocess.Popen(cmd, stderr=subprocess.STDOUT, stdout=subprocess.PIPE, text=True) as proc:
             print(proc.returncode)
```
bkmartinjr, line 49: debugging code?

tools/cellxgene_census_builder/src/cellxgene_census_builder/host_validation.py
bkmartinjr: looks like a latent bug you have caught...nice

tools/cellxgene_census_builder/src/cellxgene_census_builder/data_copy.py
ebezzi: The code was already there 🤷 I think it doesn't hurt.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
ebezzi: Probably a good idea.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
bkmartinjr: do_sync_bucket_to_bucket?

tools/cellxgene_census_builder/src/cellxgene_census_builder/data_copy.py
bkmartinjr: but cleanup is goodness :-)   I'll take the blame for the garbage!

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -52,17 +52,24 @@ def do_build(args: CensusBuildArgs, skip_completed_steps: bool = False) -> int:
...
-        do_build_soma,
-        do_validate_soma,
-        do_create_reports,
-        do_data_copy,
-        do_the_release,
-        do_report_copy,
-        do_old_release_cleanup,
-    ]
+    # Infer steps from environment variables, if provided
+    if os.getenv("CENSUS_BUILD_STEPS"):
```
bkmartinjr, line 56: this model makes me concerned that we could easily get steps out of order (i.e., the external code could get out of sync with the _required_ execution order).

I would be more comfortable if each step could explicitly check that its dependencies were fulfilled.  The old code did this implicitly with the ordered `build_steps` array, which could not be executed out of order.

The new code is (inferring steps from environment) is fragile, as it allows execution in any order, skipping steps, etc.  The entire system assumes this can't happen, so we need to protect against that. In other words, it is a state machine (simple sequence that can pause/resume), and the new code doesn't enforce that.

Concrete proposal:
1. the env variable steps should be sanity checked for order and that they don't skip anything (e.g.,  step A, B, D when it should be A, B, C, D).
2. each individual step does at least a reasonable check that its upstream dependent steps were executed (via the state and possibly more)

You could skip the first item if you simply let the environment say "start at step X", and didn't allow multiple (possibly out of order) execution.

TL;DR - I'd like to see this get substantially hardened before we deploy it.

bkmartinjr (changes requested):
Several minor things.  The major concern is that the state machine in __main__.py is no longer enforced or sanity checked, which will allow all kinds of potential mischief to occur.

I'd like to see a clean sequence of steps, checks that the sequence is followed, and checks that precursor steps have been done before the next one starts.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
ebezzi: I think the best solution here is to use the `build-state.yaml` file to ensure that each step can start only if all the pre-requisite steps have successfully been completed. Due to the nature of the AWS SFN (and potentially other runners like Airflow), the state needs to be saved in a shared location, e.g. S3. This is a medium effort change, so before I implement it I would make sure we all agree on it. Thoughts @bkmartinjr @atolopko-czi?

bkmartinjr: I agree that it should use build-state (the code already records each step as having been completed).  It already records the succexs of each step with `args.state[build_step.__name__] = True`

Assuming all steps are idempotent (which I _think_ is true), you could simply add a check that the immediately preceding step was completed successfully.

I.e., I'm thinking something like:
1. in `do_build`, make sure you set the step completion state to False before each step (it currently assumes steps will never re-run)
2. In each `do_step_B`, which depends on on `do_step_A`, add something like `assert args.state['do_step_A']`

ebezzi: The issue is that the execution will now be split into multiple containers so we can't do this in memory. In particular, the `sync` step (the new one) and the `release` will run in separate batch jobs.

bkmartinjr: It already commits it to the state file (see line 79 in __main__.py), so the "in file" solution is already in place.  I.e., it was built as a "not" memory only solutoin.

We should talk in person if lines 78, 79 are not clear.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
ebezzi: Maybe `do_sync_to_replica_s3_bucket`. I think using the term `replica` makes it more consistent.

tools/cellxgene_census_builder/src/cellxgene_census_builder/data_copy.py
ebezzi: I'll get rid of it then.

bkmartinjr: I like the logging - it was only the print I was maligning :-)

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
```diff
@@ -20,38 +20,70 @@
...
+    host_validation_disable: int = field(
+        converter=bool, default=False
+    )  # if True, host validation checks will be skipped
+    host_validation_min_physical_memory: int = field(converter=int, default=512 * 1024**3)  # 512GiB
+    host_validation_min_swap_memory: int = field(converter=int, default=2 * 1024**4)  # 2TiB
+    host_validation_min_free_disk_space: int = field(converter=int, default=1 * 1024**4)  # 1 TiB
     #
     # Release clean up
-    "release_cleanup_days": 32,  # Census builds older than this are deleted
+    release_cleanup_days: int = field(converter=int, default=32)  # Census builds older than this are deleted)
```
bkmartinjr, line 56: super nit:  remove trailing ')' in comment

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
ebezzi: I see the commit and I think it's pretty much what we need, aside from:
1. Adding the state file to S3 so it can be used by separate containers
2. Adding explicit checks to make sure that steps can only execute in the right state.
Happy to chat in person if you want to speed up the decision.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -38,21 +39,71 @@ def main() -> int:
...
+        return do_build(build_args)
+    elif command == "replicate":
+        return do_replicate(build_args)
+    elif command == "release":
+        return do_release(build_args)
+    elif command == "sync-release":
+        return do_sync_release(build_args)
+    elif command == "pass":
+        return 0  # Used for testing
+    else:
```
bkmartinjr, line 53: This else branch (full build) will execute upon any unrecognized CENSUS_BUILD_COMMAND, including an unset env variable.  I'd rather error check for sanity - i.e., only do the full build when it is unset, and throw an error if it is set to an unknown value.

In other words, I'd like better error checking

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -62,7 +113,22 @@ def do_build(args: CensusBuildArgs, skip_completed_steps: bool = False) -> int:
...
         do_report_copy,
         do_old_release_cleanup,
+        do_log_copy,
     ]
+    rv = _do_build(build_steps, args, skip_completed_steps)
+    logging.info("Census full build: completed")
+    return rv
+
+
+def _do_build(
```
bkmartinjr, line 123: at this point, this function name is confusing and doesn't match what the function does.   Suggest a rename.

Maybe:   _do_steps ?   I"m agnostic, but it is confusing to see do_release call _do_build

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -62,7 +113,22 @@ def do_build(args: CensusBuildArgs, skip_completed_steps: bool = False) -> int:
...
+    build_steps: Sequence[Callable[[CensusBuildArgs], bool]], args: CensusBuildArgs, skip_completed_steps: bool = False
+) -> int:
+    """
+    Top-level build sequence.
```
bkmartinjr, line 127: comment also wrong at this point.  Good to clean all this up as you refactor

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -38,21 +39,71 @@ def main() -> int:
...
+        do_report_copy,
+        do_log_copy,
+    ]
+    rv = _do_build(build_steps, args)
+    logging.info("Census build: completed")
+    return rv
+
 
-    build_steps: List[Callable[[CensusBuildArgs], bool]] = [
+def do_release(args: CensusBuildArgs) -> int:
```
bkmartinjr, line 79: I would prefer that we have a naming convention that differentiates between the single-step "do something" functions (e.g., do_prebuild_checks), and the top-level UX functions which execute _groups_ of sub-steps.

A possible scheme is to call the aggregate UX functions "groups", so rename do_build to do_build_group, etc.

Just trying for clarity / comprehensibility.  So feel free to improve/modify

bkmartinjr: Also, any reason the three top-level UX commands (release, build, replicate) are not just defined by a data structure containing their steps, with a generic driver, e.g.,
```
{
    "release": [do_the_release],
    "replicate": [do_sync_artifact_to_replica_s3_bucket],
    "build": [
        do_prebuild_set_defaults,
        do_prebuild_checks,
        do_build_soma,
        do_validate_soma,
        do_create_reports,
        do_data_copy,
        do_report_copy,
        do_log_copy,
    ]
}
```

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -221,6 +282,33 @@ def do_log_copy(args: CensusBuildArgs) -> bool:
     return True
 
 
+def do_sync_release_file_to_replica_s3_bucket(args: CensusBuildArgs) -> bool:
```
bkmartinjr, line 285: These new operations (as well as do_the_release) need to have checks that the build was done.

For example, `do_the_release` should contain an assertion like:

> assert args.state['do_validate_soma'] is True
> assert args.state['do_data_copy'] is true

I.e., to confirm that steps it depends on were succesful.

Honestly, I'd like to see _all_ do_foo functions add these assertions (now that we allow potential for out-of-order accidental execution)

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
ebezzi: Absolutely agree. `_do_steps` looks great.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
ebezzi: We can revert to the previous convention, I am just a bit worried about the verbosity.

bkmartinjr: OK, I can live with the change - this is not a major deal :-)

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
ebezzi: Totally agree. The full build should be performed only if `CENSUS_BUILD_COMMAND` is explicitly undefined.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
ebezzi: Discussed offline. We'll introduce a shared state on S3 in the next iteration.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -38,31 +38,60 @@ def main() -> int:
     # Process initialization/setup must be done early. NOTE: do NOT log before this line!
     process_init(build_args)
 
-    # Return process exit code (or raise, which exits with a code of `1`)
-    return do_build(build_args, skip_completed_steps=cli_args.test_resume)
+    # List of available commands for the builder
+    commands = {
```
bkmartinjr, line 42: I recommend adding a top-level command for old-release-cleanup. Otherwise, the only way it can occur is if you do a full build.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -133,3 +149,18 @@ def get_census_version_directory() -> Dict[CensusVersionName, CensusVersionDescr
...
     return cast(Dict[CensusVersionName, CensusVersionDescription], directory)
+
+
+def get_census_mirrors() -> CensusMirrors:
+    response = requests.get(CELL_CENSUS_MIRRORS_DIRECTORY_URL)
+    response.raise_for_status()
+    return cast(CensusMirrors, response.json())
+
+
+def _resolve_census_locator(locator: CensusLocator, mirror: CensusMirror) -> CensusLocator:
```
ebezzi, line 160: This function resolves the `relative_uri` field against the mirror, and returns a new `CensusLocator` object with the resolved absolute uri. If this feels hacky, we can introduce a new type.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -38,31 +38,63 @@ def main() -> int:
...
+            do_prebuild_checks,
+            do_build_soma,
+            do_validate_soma,
+            do_create_reports,
+            do_data_copy,
+            do_the_release,
+            do_report_copy,
+            do_old_release_cleanup,
+            do_log_copy,
+        ],
```
atolopko-czi, line 72: nit: can we order these as they're intended to be called?

atolopko-czi (approved):
LGTM. Added a number of minor suggestions and nits, so have a look, but nothing blocking!

I don't think we can do much in terms of adding tests for the new code, since it all surrounds S3 and invoking the full process, but just a callout to consider this in case there some opportunities for testing any critical sections.

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -76,6 +86,8 @@ def open_soma(
     Args:
         census_version:
             The version of the Census, e.g. "latest" or "stable". Defaults to "stable".
+        mirror:
+            The mirror used to retrieve the Census. Defaults to the `us-west-2` mirror.
```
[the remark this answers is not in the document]
ebezzi: True, although this is going to be public-facing docs, and I'm not sure we want to confuse the users with technical details. Maybe we can use a more _diplomatic_ message, like `If not specified, a suitable mirror will be chosen automatically`. @pablo-gar any opinions?

Also, if we don't want to expose this API publicly, we can omit this from the arguments from the time being.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -20,6 +24,7 @@
     "CensusLocator",
     {
         "uri": str,  # resource URI
+        "relative_uri": str,  # resource URI (relative)
         "s3_region": Optional[str],  # if an S3 URI, has optional region
```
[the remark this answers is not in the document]
ebezzi: Not easily. The current `release.json` cannot rename its fields due to backwards compatibility. We could create an adaptor but that will also make the code more complex. WDYT?

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -38,31 +38,63 @@ def main() -> int:
...
+            do_validate_soma,
+            do_create_reports,
+            do_data_copy,
+            do_report_copy,
+            do_log_copy,
+        ],
+        "cleanup": [
+            do_old_release_cleanup,
+        ],
+        "full-build": [
```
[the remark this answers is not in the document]
ebezzi: Sure, I can rename that. This is unused anyway.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -38,31 +38,63 @@ def main() -> int:
...
+    # The build command is set via the CENSUS_BUILD_COMMAND environment variable.
+    # If the variable is not defined, perform a full build by default.
+    command = os.getenv("CENSUS_BUILD_COMMAND", "full-build")
 
-def do_build(args: CensusBuildArgs, skip_completed_steps: bool = False) -> int:
-    """
-    Top-level build sequence.
+    # Used for testing.
+    if command == "pass":
+        return 0
```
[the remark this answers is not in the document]
ebezzi: It's mostly used if you want to test a specific step of the pipeline. Instead of removing all the other steps, you can just specify `command = pass` so that they don't do anything.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -38,31 +38,63 @@ def main() -> int:
...
 
-    Built steps will be executed in order. Build will stop if a build step returns non-zero
-    exit code or raises.
+    build_steps = commands[command]
+    rv = _do_steps(build_steps, build_args, cli_args.test_resume)
+    return rv
+
+
+def _do_steps(
+    build_steps: Sequence[Callable[[CensusBuildArgs], bool]], args: CensusBuildArgs, skip_completed_steps: bool = False
```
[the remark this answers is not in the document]
ebezzi: By running this in an ephemeral environment, this isn't really used anymore.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -79,11 +111,6 @@ def do_build(args: CensusBuildArgs, skip_completed_steps: bool = False) -> int:
             args.state.commit(args.working_dir / CENSUS_BUILD_STATE)
             logging.info(f"{step_n_of}: complete")
 
-        logging.info("Census build: completed")
```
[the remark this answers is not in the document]
ebezzi: We do have the log above (step n of m completed). I think that's enough. This log has been removed since this new method also deals with partial steps and not full builds anymore.

tools/cellxgene_census_builder/src/cellxgene_census_builder/data_copy.py
```diff
@@ -21,13 +21,30 @@ def sync_to_S3(from_path: Union[str, pathlib.PosixPath], to_path: str, dryrun: b
...
+
+def _sync_to_S3(from_path: str, to_path: str, dryrun: bool = False) -> None:
+    cmd = ["aws", "s3", "sync", from_path, to_path, "--no-progress"]
     if dryrun:
         cmd += ["--dryrun"]
 
     returncode = -1
     try:
-        _log_it(f"Starting copy {from_path.as_posix()} -> {to_path}", dryrun)
+        _log_it(f"Starting copy {from_path} -> {to_path}", dryrun)
```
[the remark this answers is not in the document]
ebezzi: I know you're sarcastic, but they actually are still supported 😄 it's just that `.as_posix()` is called directly in the function caller. I didn't change any of that logic.

api/python/cellxgene_census/src/cellxgene_census/_open.py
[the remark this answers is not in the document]
atolopko-czi: +1 for diplomatic message

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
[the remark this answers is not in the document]
atolopko-czi: ok, let's chalk it up to acceptable tech debt.  we'll deal w/it if ever we host on other cloud providers (adding a new region field and deprecating s3_region, e.g.)

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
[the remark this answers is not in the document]
atolopko-czi: delete it then?

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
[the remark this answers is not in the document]
atolopko-czi: ah, nice!

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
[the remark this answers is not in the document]
atolopko-czi: ok, yeah, I haven't thought through what ad hoc testing looks like now--maybe we could a few sentences about how to manually test the multi-step process in dev env on an instance w/o StepFn orchestration.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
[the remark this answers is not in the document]
ebezzi: It is used by the default behavior. This was Bruce's request in case you want to build manually. I agree with you that it would make more sense to only do a "local" build, so I'll comply.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -35,9 +40,31 @@
 )
 CensusDirectory = Dict[CensusVersionName, Union[CensusVersionName, CensusVersionDescription]]
 
+CensusMirrorName = str  # name of the mirror
+CensusMirror = TypedDict(
```
bkmartinjr, line 44: The fields need better comments and typing.  Reading the code, I can't quite tell what the legal values are, or how they will be used.  I recommend:
* Use `Literal` for `provider` (i.e., for any enumeration)
* Document the semantics / constraints of each field, if any.  E.g., `base_uri` must be absolute, no?

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -35,9 +40,31 @@
...
+    "CensusMirror",
+    {
+        "provider": str,  # provider of the mirror. Only S3 is supported in this version.
+        "base_uri": str,  # name of the bucket or resource
```
bkmartinjr, line 48: comment is misleading.  This is the "Base (absolete) URI of bucket or resource containing the Census mirror", not a "name"

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -11,6 +11,10 @@
 import requests
 from typing_extensions import TypedDict
 
+from ._util import _uri_join
+
+SUPPORTED_PROVIDERS = ["S3", "file"]
+
 """
 The following types describe the expected directory of Census builds, used
 to bootstrap all data location requests.
```
bkmartinjr, line 20: Please add a trivial example to the comment that shows both mirrors.json and release.json (just for ease of reading the code).  I'd also add a sentence or two about how the two are used to resolve the actual Census location.

It is also worth noting that the JSON data MAY contain other fields, which should be ignored, ie.., are private or undefined in this interface.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -35,9 +40,31 @@
 )
 CensusDirectory = Dict[CensusVersionName, Union[CensusVersionName, CensusVersionDescription]]
 
+CensusMirrorName = str  # name of the mirror
+CensusMirror = TypedDict(
+    "CensusMirror",
+    {
+        "provider": str,  # provider of the mirror. Only S3 is supported in this version.
+        "base_uri": str,  # name of the bucket or resource
+        "region": Optional[str],  # region of the bucket or resource
```
bkmartinjr, line 49: This comment is different than the comment for the same field in the Locator. If it has different semantics, I would document them, or rationalize the comments.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
bkmartinjr: Also, please document which fields are going to be deprecated/removed in a future version.  Ideally document both the old and new API usage pattern (or add a unit test to the same effect)

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -133,3 +160,31 @@ def get_census_version_directory() -> Dict[CensusVersionName, CensusVersionDescr
 
     # Cast is safe, as we have removed all aliases
     return cast(Dict[CensusVersionName, CensusVersionDescription], directory)
+
+
+def get_census_mirrors() -> CensusMirrors:
```
bkmartinjr, line 165: if this is a public method, it needs a docstring

api/python/cellxgene_census/src/cellxgene_census/_open.py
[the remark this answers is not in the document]
bkmartinjr: Fully agree with @ebezzi that the top-level docstring needs to describe the defaulting behavior, even if our implementation buries the code in private (undocumented) methods

bkmartinjr, line 90: Defaults to _AWS S3 us-west-2_  (i.e., need to document the provider, not just the region)

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -76,6 +86,8 @@ def open_soma(
...
         uri:
             The URI containing the Census SOMA objects. If specified, will take precedence
             over ``census_version`` parameter.
```
bkmartinjr, line 93: on the next arg, `context`, please add:  "... which will be used to open the SOMA object.  Optional, defaults to None."

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -76,6 +86,8 @@ def open_soma(
     Args:
```
bkmartinjr, line 86: Other changes that need to be added:
* there are now more errors, e.g., mirror unknown, provider unsupported, etc.
* need example using mirror

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -123,11 +135,20 @@ def open_soma(
     """
 
     if uri is not None:
-        return _open_soma({"uri": uri, "s3_region": None}, context)
+        return _open_soma({"uri": uri, "region": None, "provider": "file"}, context)
```
bkmartinjr, line 138: this is confusing to me - the provider is not necessarily a file (see the examples above, where several types of URIs are used -- s3:, file:, etc).

Currently `provider` does nothing - it is there to allow future code to decide how to compose a fully resolved URI and/or which API/protocol to use.  So this works, but will likely fail in the future.

I think we need to do one of two things:
* Have the code guess/determine the provider based on the value of the `uri=` param.
* Or, define an `unknown` provider, which tries its best, but may fail in some cases
* or?

Either way, if the user provides `open_soma(uri="s3://...")`, it seems pretty weird to call this a `file` provider.

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -123,11 +135,20 @@ def open_soma(
...
         raise ValueError("Must specify either a census version or an explicit URI.")
 
+    mirrors = get_census_mirrors()
+    selected_mirror: CensusMirror
+    if mirror is not None:
+        if mirror not in mirrors:
+            raise KeyError("Mirror not found.")
+        selected_mirror = mirrors[mirror]  # type: ignore
+    else:
+        selected_mirror = mirrors[mirrors["default"]]  # type: ignore
```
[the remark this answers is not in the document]
bkmartinjr: this is another great example of why the "resolver" code should be all centralized in this file

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
ebezzi: The semantics are slightly different since `s3_region` is effectively a deprecated field (it will only be used by older versions of the census that do not have mirror resolution). I'll mark it explicitly.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -133,3 +160,31 @@ def get_census_version_directory() -> Dict[CensusVersionName, CensusVersionDescr
...
+    response = requests.get(CELL_CENSUS_MIRRORS_DIRECTORY_URL)
+    response.raise_for_status()
+    return cast(CensusMirrors, response.json())
+
+
+def _assert_mirror_supported(mirror: CensusMirror) -> None:
+    """
+    Verifies if the mirror is supported by this version of the census.
```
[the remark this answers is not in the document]
ebezzi: Modified to `version of the census API` and fixed the error message.

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -151,7 +172,9 @@ def open_soma(
             "consistency."
         )
 
-    return _open_soma(description["soma"], context)
+    locator = _resolve_census_locator(description["soma"], selected_mirror)
+
+    return _open_soma(locator, context)
 
 
 def get_source_h5ad_uri(dataset_id: str, *, census_version: str = DEFAULT_CENSUS_VERSION) -> CensusLocator:
```
bkmartinjr, line 180: will this ever need a mirror name as well?

We are moving to the decoupling of these two data sources, but I don't see why they would not be _both_ potentially mirrored, and for the same reasons.   CC: @pablo-gar @brianraymor

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
ebezzi: I marked it as private. We want a public method to return all the mirrors, but I think we should only return the name and the geographical location, and possibly the provider. @pablo-gar any thoughts? Do we want to expose it immediately?

bkmartinjr: Also, if a public API, naming should be consistent between this and `get_census_version_directory`.  I'd recommend `get_mirror_directory` so that we don't change the aforementioned existing API

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -133,3 +160,31 @@ def get_census_version_directory() -> Dict[CensusVersionName, CensusVersionDescr
...
+    This method provides a proper error message in case an old version of the census
+    tries to connect to an unsupported mirror.
+    """
+    if mirror["provider"] not in SUPPORTED_PROVIDERS:
+        raise ValueError(f"Unsupported mirror provider: {mirror['provider']}. Try upgrading the census package.")
+
+
+def _resolve_census_locator(locator: CensusLocator, mirror: CensusMirror) -> ResolvedCensusLocator:
```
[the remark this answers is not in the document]
ebezzi: I moved them to `_open.py`

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -33,13 +40,15 @@
 api_logger.addHandler(logging.StreamHandler())
```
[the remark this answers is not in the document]
ebezzi: Not sure - I see that Andrew added those. Let's ask him when he's back.

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -179,7 +202,13 @@ def get_source_h5ad_uri(dataset_id: str, *, census_version: str = DEFAULT_CENSUS
         's3_region': 'us-west-2'}
     """
     description = get_census_version_description(census_version)  # raises
-    census = _open_soma(description["soma"])
+
+    # For h5ads, it makes sense to use the default mirror, since the artifacts themselves won't be mirrored
```
bkmartinjr, line 206: I don't think this assumption is true in the longer term.  We have the same motivation to mirror all large datasets

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -35,9 +40,31 @@
...
+    {
+        "provider": str,  # provider of the mirror. Only S3 is supported in this version.
+        "base_uri": str,  # name of the bucket or resource
+        "region": Optional[str],  # region of the bucket or resource
+    },
+)
+
+CensusMirrors = Dict[CensusMirrorName, Union[CensusMirrorName, CensusMirror]]
+
+ResolvedCensusLocator = TypedDict(
```
bkmartinjr, line 55: Suggest renaming this `ResourceLocator` or `ResolvedResourceLocator`, as it is used for both Census and H5AD resources, not just the Census.

api/python/cellxgene_census/src/cellxgene_census/_open.py
bkmartinjr, line 180: The return type for this function is now wrong, or at least somewhat wrong.  It should return something more akin to the `ResolvedCensusLocator` contents, even if it isn't really locating a Census.  I'm not entirely sure what the right approach is here, given that we are soon moving to a model where the H5ADs are only available via HTTPS, and don't need a full locator (i.e., no knowledge of protocol, provider or region is required), and at that point, we could drastically simplify this to just return a naked URI....

api/python/cellxgene_census/tests/test_open.py
```diff
@@ -13,11 +13,21 @@
...
+    "private": {"provider": "S3", "base_uri": "s3://cellxgene-data-public/", "region": "us-west-2"},
+}
 
 
 @pytest.mark.live_corpus
-def test_open_soma_stable() -> None:
+def test_open_soma_stable(requests_mock: rm.Mocker) -> None:
+    requests_mock.real_http = True
+    requests_mock.get(CELL_CENSUS_MIRRORS_DIRECTORY_URL, json=mirrors_json)
+
```
bkmartinjr, line 30: nit: can we clean up the detritus in lines 42-47 (it appears identical to that on lines 36-40, albeit commented out)

api/python/cellxgene_census/src/cellxgene_census/_open.py
ebezzi: My assumption here is that `uri` was used to specify local files, but I now realize that this is more generic and can also be used to point to a file hosted in S3 (not by us). I'll go with `unknown` for now.

api/python/cellxgene_census/tests/test_open.py
```diff
@@ -38,15 +48,21 @@ def test_open_soma_stable() -> None:
...
     # There should _always_ be a 'latest'
     with cellxgene_census.open_soma(census_version="latest") as census:
         assert census is not None
         assert isinstance(census, soma.Collection)
 
 
 @pytest.mark.live_corpus
-def test_open_soma_with_context() -> None:
+def test_open_soma_with_context(requests_mock: rm.Mocker) -> None:
```
bkmartinjr, line 62: this is using the old (deprecated) interface (e.g., `s3_region`).  We should test the new interface (or both)

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -123,11 +135,20 @@ def open_soma(
...
+        return _open_soma({"uri": uri, "region": None, "provider": "file"}, context)
 
     if census_version is None:
         raise ValueError("Must specify either a census version or an explicit URI.")
 
+    mirrors = get_census_mirrors()
+    selected_mirror: CensusMirror
+    if mirror is not None:
+        if mirror not in mirrors:
+            raise KeyError("Mirror not found.")
```
bkmartinjr, line 147: wrong error - should be a ValueError.  KeyError is explicitly restricted to a mapping type lookup failure.

bkmartinjr: and yes, this points out that I used the wrong error in `get_census_version_description`.  Sigh!

api/python/cellxgene_census/tests/test_open.py
```diff
@@ -92,18 +110,123 @@ def test_open_soma_errors(requests_mock: rm.Mocker) -> None:
...
+            KeyError,
+            match=re.escape("Mirror not found."),
+        ):
+            cellxgene_census.open_soma(mirror="bogus-mirror")
+
+
+def test_open_soma_rejects_non_s3_mirror(requests_mock: rm.Mocker) -> None:
+    mock_mirrors = {
+        "default": "test-mirror",
+        "test-mirror": {"provider": "GCS", "bucket": "gcs://mirror-bucket-1/"},
```
bkmartinjr, line 167: this is outdated code - ie., the mock data structure does not match the actual datastructure ("bucket"?)

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
bkmartinjr: Any update on this?

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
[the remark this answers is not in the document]
bkmartinjr: I'm fine changing it!

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -38,31 +38,63 @@ def main() -> int:
...
+            do_the_release,
+            do_report_copy,
+            do_old_release_cleanup,
+            do_log_copy,
+        ],
+    }
 
+    # The build command is set via the CENSUS_BUILD_COMMAND environment variable.
+    # If the variable is not defined, perform a full build by default.
+    command = os.getenv("CENSUS_BUILD_COMMAND", "full-build")
```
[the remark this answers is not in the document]
bkmartinjr: Or even have no default - ie. make it a required param?

api/python/cellxgene_census/tests/test_open.py
ebezzi: Removed.

api/python/cellxgene_census/tests/test_open.py
ebezzi: Since this is a `live_corpus` test, we can update this after the changes to release.json, otherwise it will fail. Agree we should switch to the new interface though.

api/python/cellxgene_census/src/cellxgene_census/_open.py
ebezzi: I will fix them both. Thanks for clarifying the extent of `KeyError`.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -223,6 +275,33 @@ def do_log_copy(args: CensusBuildArgs) -> bool:
...
+
+    sync_to_S3_remote(
+        source_key,
+        dest_key,
+        dryrun=args.config.dryrun,
+    )
+    return True
+
+
+def do_sync_artifact_to_replica_s3_bucket(args: CensusBuildArgs) -> bool:
```
bkmartinjr, line 293: I realize we have previously discussed this, but it would be MUCH better, if these `sync` operations had some sort of error checking that they are copying a clean, built, released Census, and not some random directory and/or failed build.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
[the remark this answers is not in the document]
bkmartinjr: It is still an issue when there is a failure of some sort, not caught correctly by the code.  I am a big fan of belt-and-suspenders at this level.

In other words, perhaps the right behavior is to _error_ out, rather than skipping.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -38,31 +39,67 @@ def main() -> int:
     # Process initialization/setup must be done early. NOTE: do NOT log before this line!
     process_init(build_args)
 
-    # Return process exit code (or raise, which exits with a code of `1`)
-    return do_build(build_args, skip_completed_steps=cli_args.test_resume)
+    # List of available commands for the builder
+    commands = {
+        "release": [do_the_release],
+        "replicate": [do_sync_artifact_to_replica_s3_bucket],
```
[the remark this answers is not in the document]
ebezzi: I introduced `replica` in this context to differentiate from the `primary` bucket. It is not used otherwise.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
ebezzi: `test_open_soma_works_if_no_relative_uri_specified` is the test for the old behavior. I will add some docs as requested.

api/python/cellxgene_census/src/cellxgene_census/_open.py
ebezzi: @pablo-gar do we want to introduce a naming convention for the mirrors?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/__main__.py
```diff
@@ -32,6 +32,7 @@ def main() -> int:
 
 
 def create_args_parser() -> argparse.ArgumentParser:
+    config = CensusBuildConfig()
```
bkmartinjr, line 35: for clarity, I would rename this variable `default_config` or some such....

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
```diff
@@ -20,27 +20,34 @@
...
+    )  # if True, will disable copy of data/logs/reports/release.json to S3 buckets. Will NOT disable local build, etc.
+    #
+    # Primary bucket location
+    cellxgene_census_S3_path: str = field(default="s3://cellxgene-data-public/cell-census")
     #
-    # Paths and census version name determined by spec.
-    "cellxgene_census_S3_path": "s3://cellxgene-data-public/cell-census",
-    "logs_S3_path": "s3://cellxgene-data-public-logs/builder",
-    "build_tag": datetime.now().astimezone().date().isoformat(),
+    # Replica bucket location
```
bkmartinjr, line 40: replica?  mirror?

tools/cellxgene_census_builder/src/cellxgene_census_builder/release_manifest.py
```diff
@@ -18,6 +18,7 @@
     "CensusLocator",
     {
         "uri": str,  # resource URI
+        "relative_uri": str,  # relative URI
```
bkmartinjr, line 21: good to synchronize these comments with the public API (e.g., which are obsolete, etc)

tools/cellxgene_census_builder/src/cellxgene_census_builder/release_manifest.py
```diff
@@ -117,9 +118,27 @@ def _validate_release_info(
...
+        "relative_uri": urlcat(prefix, rls_tag, "soma/"),
+        "s3_region": CENSUS_AWS_REGION,
+    }
+    expected_h5ads_locator = {
+        "uri": urlcat(census_base_url, rls_tag, "h5ads/"),
+        "relative_uri": urlcat(prefix, rls_tag, "h5ads/"),
+        "s3_region": CENSUS_AWS_REGION,
+    }
+
+    print(rls_info["soma"], expected_soma_locator)
```
bkmartinjr, line 137: debugging code?

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
bkmartinjr: follow-up idea: maybe it is time for a short design doc on release.json/mirrors.json to live in the docs directory?  That would be a good place to document the data model assumptions, etc.

tools/cellxgene_census_builder/tests/test_workflow_steps.py
```diff
@@ -8,6 +8,7 @@
 from cellxgene_census_builder.release_manifest import CensusVersionDescription
 
 TEST_BUCKET_PATH = "s3://bucket/path"
+TEST_BUCKET_PREFIX = "/path"
```
bkmartinjr, line 11: as these two constants must be in sync, might be better to do this:
```
TEST_PATH = "/path"
TEST_BUCKET_URI = f"s3://bucket/{TEST_PATH}"
```

Note that I also renamed them to match their actual contents.

tools/cellxgene_census_builder/tests/test_release_manifest.py
```diff
@@ -3,7 +3,7 @@
 from unittest import mock
 
 import pytest
-from cellxgene_census_builder.build_state import CENSUS_CONFIG_DEFAULTS
+from cellxgene_census_builder.build_state import CensusBuildConfig
 from cellxgene_census_builder.release_manifest import (
     CENSUS_AWS_REGION,
```
bkmartinjr, line 8: suggest moving to CensusBuildConfig, rather than harded as a global

tools/cellxgene_census_builder/tests/test_workflow_steps.py
bkmartinjr: I see same pattern in other test files - might want to fix all of them

Also, this is a _nit_

api/python/cellxgene_census/tests/test_open.py
ebezzi: This wasn't detected due to `provider` being the only field used for the test here. Thanks for the catch!

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
bkmartinjr: mentioned elsewhere: we should standardize on one name (e.g., mirror, replica, ....)

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
ebezzi: I agree. I'll see what can be done, if it's too complex I'll file an issue to do this later, likely with the shared S3 build state.

tools/cellxgene_census_builder/src/cellxgene_census_builder/release_manifest.py
ebezzi: Why do we even have two copies of those objects? Since the builder has the census as a dependency, maybe we could be more DRY here...

tools/cellxgene_census_builder/src/cellxgene_census_builder/release_manifest.py
```diff
@@ -117,9 +118,27 @@ def _validate_release_info(
...
+
+    expected_soma_locator = {
+        "uri": urlcat(census_base_url, rls_tag, "soma/"),
+        "relative_uri": urlcat(prefix, rls_tag, "soma/"),
+        "s3_region": CENSUS_AWS_REGION,
+    }
+    expected_h5ads_locator = {
+        "uri": urlcat(census_base_url, rls_tag, "h5ads/"),
+        "relative_uri": urlcat(prefix, rls_tag, "h5ads/"),
+        "s3_region": CENSUS_AWS_REGION,
```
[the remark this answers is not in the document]
ebezzi: `s3_region` is effectively a deprecated field, so I think hardcoding it makes sense. It will always point to the default bucket, which I don't think will ever be moved from `us-west-2`.

api/python/cellxgene_census/src/cellxgene_census/_open.py
bkmartinjr: Discussing with @pablo-gar and @brianraymor  -  it seems possible that we may entirely remove this function in an upcoming release (when we resolve issue #89).

If that happens, all of my comments are moot.  

Given this, I suggest the following approach (@pablo-gar - can you comment on this?):
1. Now:  Mark this method (in the docstring) as to be deprecated in a near-future release.  Otherwise, leave it alone
2. When we fix #89, remove it entirely.  

Assumption: the Census `datasets` table has sufficient information for a user to find the H5AD used to build the Census given access to the Discover REST API.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -133,3 +160,31 @@ def get_census_version_directory() -> Dict[CensusVersionName, CensusVersionDescr
...
+
+def _resolve_census_locator(locator: CensusLocator, mirror: CensusMirror) -> ResolvedCensusLocator:
+    _assert_mirror_supported(mirror)
+
+    if locator.get("relative_uri"):
+        uri = _uri_join(mirror["base_uri"], locator["relative_uri"])
+        region = mirror["region"]
+    else:
+        uri = locator["uri"]
+        region = locator.get("s3_region")
```
mlin, line 189: @ebezzi is it correct for this else branch to ignore `mirror`?

api/python/cellxgene_census/src/cellxgene_census/_open.py
bkmartinjr: ps. if we just mark as deprecated until #89 is resolved, we can hold off any other changes including this one

api/python/cellxgene_census/src/cellxgene_census/_open.py
bkmartinjr: A great question.

That said, I don't think you need a naming scheme to fix this docstring issue.

api/python/cellxgene_census/src/cellxgene_census/_open.py
bkmartinjr: Sounds good.  This further motivates the design doc request, indicating how mirrors.json and release.json fields should be combined to resolve the actual data....

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
ebezzi: Correct. The `else` branch can be removed as soon as we'll patch the `release.json` file with the new field though, so IMHO don't include that in the R API.

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -31,13 +38,15 @@
 api_logger.addHandler(logging.StreamHandler())
 
 
-def _open_soma(locator: CensusLocator, context: Optional[soma.options.SOMATileDBContext] = None) -> soma.Collection:
+def _open_soma(
+    locator: ResolvedCensusLocator, context: Optional[soma.options.SOMATileDBContext] = None
+) -> soma.Collection:
     """Private. Merge config defaults and return open census as a soma Collection/context."""
-    context = _build_soma_tiledb_context(locator.get("s3_region"), context)
+    context = _build_soma_tiledb_context_s3(locator.get("region"), context)
```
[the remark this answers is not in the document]
bkmartinjr: I suggest you go ahead and scaffold out the provider handling code, even if it boils down to:
```python
if provider in ['s3', 'unknown']:
    ... set region
else:
    ... don't set region
```

bkmartinjr (changes requested):
this is getting close.  Lots of nits, plus the following:

1. PR title - it isn't just a builder PR, so I'd remove that bit
2. Need docs file that describe the contents/semantics of release/mirrors.json, and the process for resolving the Census location for all supported providers (including `unknown`).  This is also the right place to declare what is deprecated, what isn't, etc.
3. I'd like to see the builder workflow steps (eg., do_...), especially those that push / sync data in the public buckets, do some error checking to confirm they are working with a complete, successful build.
4. Provider handling needs to be scaffolded in the client API even if it does't do much yet.  E.g., don't set `region` if the provider is `file`, etc.
5. Need decision on whether the mirrors list is public API or not.  @pablo-gar

api/python/cellxgene_census/src/cellxgene_census/_open.py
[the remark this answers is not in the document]
ebezzi: I'll take a stab at a refactor and ping you in a comment.

api/python/cellxgene_census/tests/test_open.py
```diff
@@ -13,11 +13,21 @@
...
+    "default": "private",
+    "private": {"provider": "S3", "base_uri": "s3://cellxgene-data-public/", "region": "us-west-2"},
+}
 
 
 @pytest.mark.live_corpus
-def test_open_soma_stable() -> None:
+def test_open_soma_stable(requests_mock: rm.Mocker) -> None:
+    requests_mock.real_http = True
+    requests_mock.get(CELL_CENSUS_MIRRORS_DIRECTORY_URL, json=mirrors_json)
```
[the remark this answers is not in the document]
ebezzi: These are temporary mocks - they can be removed once the mirrors.json file is deployed in production.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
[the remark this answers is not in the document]
ebezzi: Changed to required.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
bkmartinjr: I think there are simpler checks that would be 80% of the value, for about 1% of the work (and can be replaced later if the better solution arrives).

Example:  just create a function that checks to see if the path "looks like" a Census build, e.g., contains well-known sub-paths, of nonzero length, etc.

No need to let this hold up the PR.

tools/cellxgene_census_builder/src/cellxgene_census_builder/release_manifest.py
bkmartinjr: True although as currently defined they are different (in absolute/relative URL fields).

But DRY is good, as long as we can keep our older package releases running.

tools/cellxgene_census_builder/src/cellxgene_census_builder/release_manifest.py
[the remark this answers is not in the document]
bkmartinjr: OK, not a big deal.  But given that the _bucket_ is defined in the config, I'm not sure why the region isn't also defined in the config.  They are tightly linked configuration values.....

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
bkmartinjr: I don't think you can assume that the docker image will only ever have just the Builder in it.  There are both dependent packages which use environment variables, and other code we may add to be part of the build process.  I think for best practices, I would clean up this namespace.

api/python/cellxgene_census/src/cellxgene_census/_open.py
ebezzi: I think this makes sense, but we should confirm with Pablo.

pablo-gar: Ok so having a conversation with Bruce and Brian we should move forward with the following.

1. Do not mirror h5ads.
2. `download_source_h5ad` and `get_source_h5ad_uri` must still work and always fetch from `us-west-2` since it will the only region available for them.
3. Bruce to to address #89 in the near future.
4. After,`download_source_h5ad` and `get_source_h5ad_uri` are **either** kept but their underlying implementation produces and utilizes Discover's URLs  **OR**  they are deprecated. For now I'm inclined to go with the former, but I'd like to discuss this further.

api/python/cellxgene_census/src/cellxgene_census/_open.py
pablo-gar: For now this suffices: "[provider] [region]" e.g. "AWS S3 us-west-2"

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
pablo-gar: We do need a public API whose output produces a list of values that can be directly used as an argument for `open_soma(mirror = [VALUE])`

I recommend a function `get_mirror_names() -> List[str]`

@ebezzi can you add that please

pablo-gar (changes requested):
See https://github.com/chanzuckerberg/cellxgene-census/pull/552#discussion_r1295187528

and 

https://github.com/chanzuckerberg/cellxgene-census/pull/552/files#r1295199055

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
ebezzi: Will do

bkmartinjr: I don't see how `get_mirror_names` is useful unless it *also* returns information about the mirror, sufficient for the user to pick one.  So I think:
1. it needs to return the entire mirror record, not just the name
2. we may want (over time, later) enhance the mirror record with some additional (human readable?) info that helps this process

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
ebezzi: Added a `CENSUS_BUILD` prefix. This will require some changes in the SFN, but they should be straightforward.

api/python/cellxgene_census/src/cellxgene_census/_open.py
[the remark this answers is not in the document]
ebezzi: I implemented this logic:
1. if provider == S3, then build the context using the region
2. otherwise (which includes provider == unknown), just use the passed context.

I think this makes sense because:
- If the user passes a file URI to the `uri` param, we won't need any special context to use it
- If the user passes an s3 uri with a private copy of the census (provider == unknown and uri starts with s3://), they will likely need to specify the region in a custom context. According to the [TileDB docs](https://docs.tiledb.com/main/how-to/backends/s3), a region is mandatory for non us-east-1 buckets, so in theory we could set a specific if for this condition, but it would only work if the custom bucket is in us-east-1. Not worth imho.
- If the user wants to use any other storage medium, they will also need to provide the configuration in the custom context.

Let me know if this looks good.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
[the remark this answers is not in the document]
ebezzi: Will add the checks.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
ebezzi: Agree, it needs to return the region and  the provider at minimum. I was thinking to omit the `base_uri`, but that could actually be useful if the user wants to explore the mirror via other means. I'll just return a list of the whole records.

api/python/cellxgene_census/src/cellxgene_census/_open.py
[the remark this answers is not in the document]
bkmartinjr: One clarification - we also need to ensure that the default configs are set (e.g., soma.init_buffer_bytes) when the user does *not* specify a context.  So I think the logic is:

if user-provided context, use it.  Else create one with default config.
if provider == S3, set region & anon access flag

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -223,6 +259,33 @@ def do_log_copy(args: CensusBuildArgs) -> bool:
     return True
 
 
+def do_sync_release_file_to_replica_s3_bucket(args: CensusBuildArgs) -> bool:
+    """Copy release.json to replica S3 bucket"""
```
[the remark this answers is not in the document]
bkmartinjr: Yeah, my thinking is mirrors should be _actual_ mirrors.  Contain the same bits.  So they would not have a revised release.json.  I think this makes sense, as "primary" bucket is still the same, no matter the number of mirrors.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
bkmartinjr: sounds good - and that makes it match what we do with `get_census_version_directory`.  Suggest renaming it as well, e.g., `get_mirror_directory`.  That will distinguish it from other methods we might add in the future...  (eg.., get_current_mirror?)

ebezzi: I added the method.

api/python/cellxgene_census/src/cellxgene_census/_open.py
ebezzi: As per Pablo's update, I will leave this method as is and read the h5ads from the absolute URI.

api/python/cellxgene_census/src/cellxgene_census/_open.py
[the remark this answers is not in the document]
atolopko-czi: Not sure if it's good hygiene to put that in __init__.py directly, but maybe into a new logging.py module that is _imported_ by __init__.py.  Either way, out-of-scope change for this PR, but worth considering (I think we have a pending "api logging" issue where it could be addressed).

pablo-gar (approved):

PR #555, opened 2023-06-19
fix incorrect pytorch obs soma_joinids

### LINKED ISSUE #554: PyTorch obs soma_joinids are incorrect for large value
The soma_joinids returned in the `obs` Tensors are incorrect for values exceeding 2^24. 

This is caused by an unintended casting of the (encoded) obs to float32 before being explicitly case to int32. As float32 only has precision of 2^24, any values greater than that will lose their least significant digits (they will be zeroed out).

atolopko-czi:
Fixes #554 

- Use `torch.from_numpy()` instead of `Torch.Tensor()` to construct tensors. The latter created with `dtype=float32`, which is incorrect and expensive.  In addition to fixing the typing and precision-truncation problem, we now also avoid a data copy operation by using this new method.
- Rework the test fixture generation methods to allow for ranges of obs and var soma_joinids that may start at any arbitrary value. Necessary for testing the case that produced this bug.
- Add asserts to ease paranoia.


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -131,34 +131,33 @@ def __next__(self) -> ObsDatum:
...
                 obs_partial, X_partial = self._read_partial_torch_batch(self.batch_size)
-                if X is None:
-                    obs = obs_partial
-                    X = X_partial
-                else:
-                    obs = pd.concat([obs, obs_partial], axis=0)
-                    X = sparse.vstack([X, X_partial])
-
+                obs = pd.concat([obs, obs_partial], axis=0)
+                X = sparse.vstack([X, X_partial])
```
atolopko-czi, line 135: Unrelated to PR. During debugging, realized `X is None` was always `False` and that the "first iteration" case was not in fact necessary.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -131,34 +131,33 @@ def __next__(self) -> ObsDatum:
...
+
         obs_encoded = pd.DataFrame(
             data={"soma_joinid": obs.soma_joinid}, columns=obs.columns, dtype=np.int32, index=obs.index
         )
         for col, enc in self.encoders.items():
             obs_encoded[col] = enc.transform(obs[col])
 
-        obs_tensor = torch.Tensor(obs_encoded.to_numpy()).int()
+        # `as_tensor()` avoids copying the numpy array data
+        obs_tensor = torch.as_tensor(obs_encoded.to_numpy(dtype=np.int32))
```
atolopko-czi, line 151: This is the bug fix

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -131,34 +131,33 @@ def __next__(self) -> ObsDatum:
...
+
         if not self.sparse_X:
-            X_tensor = torch.Tensor(X.todense())
+            X_tensor = torch.as_tensor(X.todense())
```
atolopko-czi, line 154: Not related to bug fix. But use same tensor constructor as above, for memory usage improvement.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -131,34 +131,33 @@ def __next__(self) -> ObsDatum:
...
+
         obs_encoded = pd.DataFrame(
             data={"soma_joinid": obs.soma_joinid}, columns=obs.columns, dtype=np.int32, index=obs.index
         )
         for col, enc in self.encoders.items():
             obs_encoded[col] = enc.transform(obs[col])
 
-        obs_tensor = torch.Tensor(obs_encoded.to_numpy()).int()
+        # `as_tensor()` avoids copying the numpy array data
+        obs_tensor = torch.from_numpy(obs_encoded.to_numpy(dtype=np.int32))
```
atolopko-czi, line 151: This is the bug fix

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -131,34 +131,33 @@ def __next__(self) -> ObsDatum:
...
+
         if not self.sparse_X:
-            X_tensor = torch.Tensor(X.todense())
+            X_tensor = torch.from_numpy(X.todense())
```
atolopko-czi, line 154: Not related to bug fix. But use same tensor constructor as above, for memory usage improvement.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -131,34 +131,33 @@ def __next__(self) -> ObsDatum:
...
         else:
             coo = X.tocoo()
 
             X_tensor = torch.sparse_coo_tensor(
                 # Note: The `np.array` seems unnecessary, but PyTorch warns bare array is "extremely slow"
-                indices=torch.Tensor(np.array([coo.row, coo.col])),
+                indices=torch.from_numpy(np.array([coo.row, coo.col])),
```
atolopko-czi, line 160: Not related to bug fix. But use same tensor constructor as above, though memory usage improvement not necessary here

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -250,8 +257,8 @@ class ExperimentDataPipe(pipes.IterDataPipe[Dataset[ObsDatum]]):  # type: ignore
     def __init__(
         self,
         experiment: soma.Experiment,
-        measurement_name: str,
-        X_name: str,
+        measurement_name: str = "raw",
+        X_name: str = "X",
```
atolopko-czi, line 261: unrelated to bug fix

api/python/cellxgene_census/tests/experimental/ml/test_pytorch.py
```diff
@@ -152,9 +163,22 @@ def test_batching__all_batches_full_size(soma_experiment: Experiment) -> None:
         next(batch_iter)
 
 
+@pytest.mark.experimental
+# noinspection PyTestParametrized,DuplicatedCode
+@pytest.mark.parametrize("obs_range,var_range,X_value_gen", [(range(100_000_000, 100_000_003), 3, pytorch_x_value_gen)])
+def test_unique_soma_joinids(soma_experiment: Experiment) -> None:
```
atolopko-czi, line 169: This is a new regression test for the bug. Was red/green before/after the fix.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
bkmartinjr: I am concerned this may have another bug - it casts the soma_joinid int64 to int32 on line 147, which will lose part of the joinid value.  If these are _really_ soma_joinid values, they they can have any value in the int64 domain including the high 4 bytes.  The census builder currently assigns them in the range [0, n_cells), so it is not yet likely to fail as our collections are < 2^31-1 in size, but it may/will at some point.  And other datasets are free to assign joinids as any legal int64.

If the values are not soma_joinids, but are reindexed to [0, n), then the code is confusing as it calls them "joinids" when they are something different.  And even in this case, we will have collections with > 2^31-1 cells at some point - might as well handle them.

So, full suggestion:  either:
* obs_encoded should not drop top 4 bytes of the joinid, and should create int64 dataframe/tensor
* or, needs to error out when there are more than 2^31-1 cells, AND reindex the joinids down to int32

Unrelated side note: as I read the pytorch docs, they imply that constructors are never the preferred way to create a Tensor, and it should be done with `torch.tensor()`, `torch.from_numpy()`, etc.  Never explicitly stated though, so this may be incorrect.

bkmartinjr (changes requested):
Switch to `from_numpy` looks very good - the docs _imply_ `torch.Tensor` is not the best practice anyway :-)

one concern about the forced down-cast of all joinids to int32, which will fail at some point (without reindexing and/or further error checks).  Would like to sort that out.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -131,34 +131,33 @@ def __next__(self) -> ObsDatum:
...
-
+                obs = pd.concat([obs, obs_partial], axis=0)
+                X = sparse.vstack([X, X_partial])
             except StopIteration:
                 break
 
         if len(obs) == 0:
             raise StopIteration
 
+        assert obs["soma_joinid"].is_unique
```
ebezzi, line 142: Mostly a nitpick, but I think this operation incurs into additional cost since determining uniqueness on a non-sorted array can't be done in linear time. Unless this is part of the operations flow, it might be wise to remove this.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ebezzi: I think this is a valid point but since the census is strongly biased towards cell data, it is also unrealistic that we'll ever have more than `2^31-1` cells within the current iteration. This is roughly 50 times what we have now and might exceed the total number of cells ever sequenced in single cell experiments (although this is purely a guess). That said, this is the same reasoning that led to the Y2K problem (or even the R limitation for sparse matrices), so if using int64 is feasible/doesn't add a lot of work, it might be worth it...

bkmartinjr: I can think of several reasons we might have joinids > 2^31-1 (e.g., we start a policy of not reusing them, we have a sparse range for some other reason, etc).

Bottom line: the code looks wrong or unsafe at best.  Minimally, I would add explicit assertions/error checks, but as joinids are not offsets, it is a bit of a cringe to say they are guaranteed to remain within a range defined by the number of cells.

atolopko-czi: I'll bring the precision in line with the SOMA spec (int64).  In recent TileDB-SOMA discussions, I thought we were moving to int32 for max shape of experiments, but may have misunderstood. In any case int64 is safest for now, agreed.

atolopko-czi: I also found that `Torch()` doesn't seem to be the Right Way to create tensors, though the library didn't complainat me unfortunately. As class constructor doesn't even take `dtype` param, I was properly coerced into using the proper factory methods.

bkmartinjr: No, we are definitely _not_ moving to int32 joinids - was never on the table.  I think you may be conflating several issues (e.g., capacity vs utilized ids).

There is a proposal to:
* set capacity by default to a small number in `soma.io`.  This does not affect the Census as it doesn't use that code
* introduce a means to ask "how many coordiates are used", but this also has no effect on the domain from which legal values can be chosen.

These are completely separate issues in reality - joindids are not offset indices

atolopko-czi: fixed

bkmartinjr (approved):

PR #557, opened 2023-06-20
Fix MacOS failing tests

ebezzi:
Temporarily remove Python 3.7 on MacOS testing due to a bug. Revert when https://github.com/actions/setup-python/issues/682 is fixed.


atolopko-czi:
We don't want to merge this unless we decide to drop support for 3.7, right? Assuming this is just serving as a test PR, perhaps make it Draft for now?

bkmartinjr (approved):

PR #566, opened 2023-06-22
Avoid unnecessary re-initialization of ExperimentDataPipe in mp mode

### LINKED ISSUE #524: Final PyTorch MVP improvements
Laundry list of items that should be addressed before we calling the PyTorch MVP implementation complete:

- [x] Performance: Reuse obs partitioned soma_joinids for each epoch iteration (currently, obs is required and recomputed each epoch).
- [ ] Performance: Pre-fetch next SOMA batch, asynchronously
- [x] Reliability: Close SOMA Experiments after iteration

atolopko-czi:
Partially resolves #524 

- In multiprocessing mode, if `ExperimentDataPipe` has been initialized prior to creating the child processes, each child process will receive a fully initialized object and the ExperimentAxisQuery and encoders do not need to be re-initialized.
- In particular, the full (unpartitioned) `_obs_joinids` array is cached and now partitioned for each child process at the time of `_ObsAndXIterator` instantiation.  In multiprocessing this should allow the PyArrow-backed soma_joinids to be memory-shared across child processes, without having to serialize the array.
- These changes also result in some code simplifications. Since the obs soma_joinids, var soma_joinds from ExperimentAxisQuery are directly cached, rather than the entire ExperimentAxisQuery object itself, the object pickling hacks could be removed (ExperimentAxisQuery cannot be pickled). Also, the lazy initialization behavior is simplified, as the obs_encoders are computed at the same time in the `ExperimentDataPipe._init()` method.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
         args: ["--config", "./tools/pyproject.toml"]
 
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.272
+    rev: v0.0.275
     hooks:
       - id: ruff
         name: ruff-cellxgene-census
@@ -30,7 +30,7 @@ repos:
         args: [ "--config=./tools/pyproject.toml", "--fix" ]
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.3.0
+    rev: v1.4.0
     hooks:
       - id: mypy
         name: mypy-cellxgene-census
--- a/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
+++ b/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
@@ -3,7 +3,7 @@ import os
 import sys
 from datetime import timedelta
 from time import time
-from typing import Any, Dict, Iterator, List, Optional, Sequence, Tuple
+from typing import Any, Dict, Iterator, Optional, Sequence, Tuple
 
 import numpy as np
 import numpy.typing as npt
@@ -97,8 +97,8 @@ class _ObsAndXIterator(Iterator[ObsDatum]):
 
     def __init__(
         self,
-        obs_joinids: List[int],
-        var_joinids: pa.Array,
+        obs_joinids: npt.NDArray[np.int64],
+        var_joinids: npt.NDArray[np.int64],
         exp_uri: str,
         aws_region: Optional[str],
         measurement_name: str,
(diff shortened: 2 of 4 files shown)


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -286,87 +287,92 @@ def __init__(
...
-                else:
-                    joinids_start = joinids_end = 0
-
-                pytorch_logger.debug(
-                    f"Process {os.getpid()} handling partition {partition + 1} of {num_partitions}, "
-                    f"index range={partition_start}:{partition_end_excl}, "
-                    f"soma_joinid range={joinids_start}:{joinids_end}, "
-                    f"{partition_size:}"
-                )
+        partition_size = len(ids) // num_partitions
```
bkmartinjr, line 335: maybe just use numpy.array_split?  (up to you - only saves 4 lines of code)

bkmartinjr (approved):
I did a cursory review, but this looks like a big improvement. nice!

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
atolopko-czi: No maybes about it--done!

api/python/cellxgene_census/tests/experimental/pp/test_hvg.py
```diff
@@ -15,7 +15,7 @@ def small_mem_context() -> soma.SOMATileDBContext:
     """used to keep memory usage smaller for GHA runners."""
     cfg = {
         "tiledb_config": {
-            "soma.init_buffer_bytes": 64 * 1024**2,
+            "soma.init_buffer_bytes": 32 * 1024**2,
```
atolopko-czi, line 18: @bkmartinjr With pytorch unit test OOM failure seemingly fixed, I was next seeing mysterious "workflow cancelled" failures, which I'm assuming meant OOM (no other errors available). This change seems to have helped. OK with lower memory usage? Experimental runs times may be a couple minutes slower, but hard to determine when comparing to past runs due to high variability in run times.

.pre-commit-config.yaml
```diff
@@ -30,7 +30,7 @@ repos:
         args: [ "--config=./tools/pyproject.toml", "--fix" ]
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.3.0
+    rev: v1.4.0
```
atolopko-czi, line 33: Was trying to address local-vs-GHA mypy error diffs.  While this didn't help, seems reasonable to keep the updated versions.

bkmartinjr: You (all) may want to periodically run `pre-commit autoupdate`.  Thanks for doing this manually

PR #567, opened 2023-06-22
fix batch size bug

atolopko-czi:
Uncovered a bug during work on #524. When a PyTorch batch spans multiple SOMA batches, the PyTorch batch size was too large.  

- Fixes bug, adds test
- Refactors `obs_batch` property to `next_soma_batch` method, removing its object side-effects. This makes testing easier and simplifies design.

Branched off of `524-pytorch-reuse-obs-joinids`.


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -130,7 +130,7 @@ def __next__(self) -> ObsDatum:
 
         while len(obs) < self.batch_size:
             try:
-                obs_partial, X_partial = self._read_partial_torch_batch(self.batch_size)
+                obs_partial, X_partial = self._read_partial_torch_batch(self.batch_size - len(obs))
```
atolopko-czi, line 133: This is the fix

bkmartinjr (approved):

PR #570, opened 2023-06-22
Correct pytorch notebook header

pablo-gar:
Fixes top-level header which will correct indentation in doc-site:

image


atolopko-czi (approved):
Thanks for catching that!

PR #575, opened 2023-06-23
Add eager fetching of SOMA batches to ExperimentDataPipe 

### LINKED ISSUE #524: Final PyTorch MVP improvements

atolopko-czi:
Closes #524

- Add `is_eager_fetch` option for conditionally enabling/disabling behavior.
- Refactor inner iteration of SOMA batches into its own iterator class, `_ObsAndXSOMAIterator`. This allows us to wrap it with the `EagerIterator`.
- Stats reporting improvements
- Add debug logging to `Eager{Batch,}Iterator`

### DIFF
--- a/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
+++ b/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
@@ -15,7 +15,7 @@ import somacore
 import tiledbsoma as soma
 import torch
 import torchdata.datapipes.iter as pipes
-from attr import attrs
+from attr import define
 from scipy import sparse
 from sklearn.preprocessing import LabelEncoder
 from somacore.query import _fast_csr
@@ -24,17 +24,29 @@ from torch.utils.data import DataLoader
 from torch.utils.data.dataset import Dataset
 
 import cellxgene_census
-from cellxgene_census._open import _build_soma_tiledb_context
+
+from ..._open import _build_soma_tiledb_context
+from ..util import EagerIterator
 
 ObsAndXDatum = Tuple[Tensor, Tensor]
+
+
+@define
+class ObsAndXSOMABatch:
+    obs: pd.DataFrame
+    X: scipy.matrix
+    stats: "Stats"
+
+    def __len__(self) -> int:
+        return len(self.obs)
+
 
 Encoders = Dict[str, LabelEncoder]
 
 pytorch_logger = logging.getLogger("cellxgene_census.experimental.pytorch")
-pytorch_logger.setLevel(logging.INFO)
 
 
-@attrs
+@define
 class Stats:
     """
     Statistics about the data retrieved by ``ExperimentDataPipe`` via SOMA API.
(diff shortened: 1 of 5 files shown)


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -256,6 +295,7 @@ def __init__(
         batch_size: int = 1,
         sparse_X: bool = False,
         soma_buffer_bytes: Optional[int] = None,
+        is_eager_fetch: bool = True,
```
bkmartinjr, line 298: Nit: don't love the parameter name given that it is a directive to use eager.  Suggest something like:
* use_eager_fetch
* be_eager
etc.

api/python/cellxgene_census/src/cellxgene_census/experimental/util/_eager_iter.py
```diff
@@ -16,18 +20,25 @@ def __init__(
         self.iterator = iterator
         self._pool = pool or futures.ThreadPoolExecutor()
         self._own_pool = pool is None
+        self._future: Optional[Future[_T]] = None
+        self.fetch_next()
+
+    def fetch_next(self) -> None:
```
bkmartinjr, line 26: this needs to be a private function (underscore) as it should never be called by  outside code.

it also does not do the fetch, but starts the fetch, so I suggest naming it something like _submit_next() or some such.....

bkmartinjr (changes requested):
Great PR - the only concern I have is one (public) API param name which I think we should resolve before landing.  See comment about `is_eager_fetch` in pytorch.py

We will have other API coming soon that has the same param / feature, so I'd like it to be consistent and good.

api/python/cellxgene_census/tests/experimental/ml/test_pytorch.py
```diff
@@ -18,7 +18,12 @@
 try:
     from torch import Tensor
 
-    from cellxgene_census.experimental.ml.pytorch import ExperimentDataPipe, experiment_dataloader
+    from cellxgene_census.experimental.ml.pytorch import (
```
bkmartinjr, line 21: need to add test of eager fetch (on and off)

bkmartinjr:
@atolopko-czi - after discussion with Pablo, I'll make the proposed change to the public API (is_eager_fetch -> use_eager_fetch).  I will leave the other items for you to cover in a subsequent PR, as there is desire to land this soon.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
bkmartinjr: this was resolved after a conversation with Pablo.

bkmartinjr (approved):
approved, but a few follow-up items remain for another post-merge PR

atolopko-czi:
> @atolopko-czi - after discussion with Pablo, I'll make the proposed change to the public API (is_eager_fetch -> use_eager_fetch). I will leave the other items for you to cover in a subsequent PR, as there is desire to land this soon.

Perfect, thank you!

Issue #190, opened 2023-02-16
R API 1.0 acceptance tests are in place and can be executed easily.


pablo-gar:
Stress test should mirror the Python stress tests described in #185. Final list for the R stress tests TBD, they will get added to this issue.

These test should mimic the tests as implemented here for python https://github.com/chanzuckerberg/cellxgene-census/issues/192

pablo-gar:
completed #576

PR #576, opened 2023-06-24
Add R acceptance tests, update Python acceptance tests

pablo-gar:
- Addresses #190 
- Also updates python acceptance tests:
   - Decomposes certain "parametrized" tests into individual test -- this allows for easy 1:1 running time comparisons to R tests.
   - Adds a few tests to increase granularity in reading logs

### DIFF
--- a/api/python/cellxgene_census/tests/test_acceptance.py
+++ b/api/python/cellxgene_census/tests/test_acceptance.py
@@ -93,7 +93,6 @@ def test_incremental_read_obs(organism: str, stop_after: Optional[int], ctx_conf
         
 @pytest.mark.live_corpus
 @pytest.mark.parametrize("organism", ["homo_sapiens", "mus_musculus"])
-@pytest.mark.parametrize("stop_after", [2, pytest.param(None, marks=pytest.mark.expensive)])
 @pytest.mark.parametrize(
     ("stop_after", "ctx_config"),
     [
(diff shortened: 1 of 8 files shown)


api/python/cellxgene_census/tests/test_acceptance.py
```diff
@@ -70,18 +70,68 @@ def table_iter_is_ok(tbl_iter: Iterator[pa.Table], stop_after: Optional[int] = 2
...
+    [
+        pytest.param(2, None),
+        pytest.param(None, DEFAULT_TILEDB_CONFIGURATION, marks=pytest.mark.expensive),
+    ],
+)
+def test_incremental_read_obs(organism: str, stop_after: Optional[int], ctx_config: Optional[Dict[str, Any]]) -> None:
     """Verify that obs, var and X[raw] can be read incrementally, i.e., in chunks"""
 
-    # open census with a small (default) TileDB buffer size, which reduces
+    # ctx_config=None open census with a small (default) TileDB buffer size, which reduces
```
bkmartinjr, line 83: See existing `small_mem_context` in test_hvg.py.  You may just want to promote this to conftest.py and use it.  That has the benefit of explicitly controlling the configuration, whereas the current code is subject to whatever defaults tiledbsoma decides are appropriate.

bkmartinjr (approved):
Overall looks fine.  Nits:
* Is there a way to parameterize the R tests?  The manual coding of all of them seems painful to maintain.
* There is an existing fixture to create an explicit "small memory" TileDB context (code curently in test_hvg.py).  Perhaps re-use that, and not depend on tiledbsoma defaults, for more reliable tests?

pablo-gar:
> Is there a way to parameterize the R tests? The manual coding of all of them seems painful to maintain.

`testthat` is the equivalent to `pytest`, `testthat` does not support parametrized tests. Some folks from google developed a package [Patrick](https://github.com/google/patrick) with one function to wrangle tests around a simple parametrization framework. 

I decided not use it because is not officially supported anywhere. I wonder if @mlin has a different opinion on it.

Issue #578, opened 2023-06-26
CI needs markdown linter


bkmartinjr:
There are a few recurring errors that cause our notebooks to mis-render on the docsite.  It would be useful to have a markdown lint in the GHA CI pipeline.

For example, bullet lists that lack a leading newline render as inline text  on the docsite(rather than a list), but render fine in vscode/github.

PR #579, opened 2023-06-26
Notebook reformatting

ebezzi:
Reformat notebook so that they render properly in the docsite.

Eventually, these rules need to be added to a linter:

1. Preformatting isn't allowed inside of a link
2. Bullet points need to have a newline before them

See #578


bkmartinjr (approved):
thank you!

bkmartinjr:
@ebezzi - can we land this, or does it need more work?

PR #589, opened 2023-06-28
[r] install tiledbsoma from github

mlin:
Install the latest tiledbsoma from github, while we're still finalizing the r-universe release process: https://github.com/single-cell-data/TileDB-SOMA/issues/1427


ebezzi (approved):
LGTM. This will install from `main` so it will be a "nightly" build, but based on my recollection r-universe is also supposed to work the same, correct?

bkmartinjr (approved):
thanks Mike!

mlin:
@ebezzi Yes, the TileDB folks are currently adjusting the r-universe configuration in preparation for release. When there is a release, I plan to add a build matrix here so that we'll test against both the released and latest versions.

PR #591, opened 2023-06-29
add sparse iterator to experimental.utils

bkmartinjr:
Add a new experimental API to iterate over X array returning Scipy sparse CSR or CSC matrix.
Update the finddups notebook to use it

Prototype for single-cell-data/TileDB-SOMA#1503

### DIFF
--- a/api/python/cellxgene_census/src/cellxgene_census/experimental/util/_csr_iter.py
+++ b/api/python/cellxgene_census/src/cellxgene_census/experimental/util/_csr_iter.py
@@ -1,19 +1,13 @@
-from concurrent import futures
-from typing import Generator, Iterator, Literal, Tuple, TypeVar
+from typing import Generator, Iterator, Tuple
 
 import numpy as np
 import numpy.typing as npt
 import pandas as pd
 import scipy.sparse as sparse
 import tiledbsoma as soma
+from typing_extensions import Literal
 
-"""
-This might make a good addition to soma.ExperimentAxisQuery.
-
-TODO: this implementation does not leverage the ExperimentAxisQuery internals
-for fast CSR creation, and instead uses the (slower) scipy.sparse implementation.
-If integrated, it should be migrated to the much faster implementation.
-"""
+from ._eager_iter import EagerIterator
 
 _RT = Tuple[Tuple[npt.NDArray[np.int64], npt.NDArray[np.int64]], sparse.spmatrix]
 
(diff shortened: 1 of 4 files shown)


api/python/cellxgene_census/src/cellxgene_census/experimental/util/_csr_iter.py
```diff
@@ -41,21 +35,35 @@ def X_sparse_iter(
...
     Returns:
         An iterator which iterates over a tuple of:
-            obs_coords
-            var_coords
+            (obs_coords, var_coords)
             SciPy sparse matrix
 
+    Examples:
+        >>> with cellxgene_census.open_soma() as census:
+        ...     exp = census["census_data"][experiment]
```
ebezzi, line 57:
```suggestion
        ...    exp = census["census_data"][experiment]
```

api/python/cellxgene_census/tests/experimental/util/test_csr_iter.py
```diff
@@ -0,0 +1,47 @@
...
+            "vfs.s3.no_sign_request": True,
+        },
+    }
+    return soma.SOMATileDBContext().replace(**cfg)
+
+
+@pytest.mark.experimental
+@pytest.mark.live_corpus
+def test_X_sparse_iter(small_mem_context: soma.SOMATileDBContext) -> None:
+    with cellxgene_census.open_soma(census_version="stable", context=small_mem_context) as census:
```
ebezzi, line 25: I recall we settled for running the tests on `latest`, since that could help catch issues with the builder. Is there any specific reason for running these on `stable`?

ebezzi (approved):

pablo-gar (approved):
LGTM, I wonder if we can make it more general to take also any numerical axis matrices `obsm`, `obsp` etc. We need to prepare to allow this type of iterations for those matrices as well.

If there are not immediate solutions for that, then please land.

api/python/cellxgene_census/tests/experimental/util/test_csr_iter.py
bkmartinjr: just made a random choice.  Switch to latest as it does make sense!

bkmartinjr:
> LGTM, I wonder if we can make it more general to take also any numerical axis matrices obsm, obsp etc. We need to prepare to allow this type of iterations for those matrices as well.

There is no practical use for those given that the census lacks these data.  For a more general implementation in SOMA, I agree with your view that it should work with all of the sparse matrices.

PR #594, opened 2023-06-30
pin builder dependencies

### LINKED ISSUE #418: Bump builder tiledbsoma version
Bump to https://pypi.org/project/tiledbsoma/1.2.2/. 

This must be done _after_ #417 and new Python API version is released.

### LINKED ISSUE #581: [builder] all python dependencies should be pinned
to prevent dependency surprises (e.g., the recent scipy/pyarrow issue) all dependencies for the Builder should be pinned.  This requires a periodic chore to update them, but at least that will be a controlled process, and is already an existing chore given that we _must_ pin at least tiledbsoma.

bkmartinjr:
Fixes #581 
Fixes #418 

Pin all builder dependencies (excluding two system packages where there is little value).

This ensures that we use the same package versions for test (GHA and manual) and the docker image.

### DIFF
--- a/tools/cellxgene_census_builder/pyproject.toml
+++ b/tools/cellxgene_census_builder/pyproject.toml
@@ -27,26 +27,26 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
 ]
 dependencies= [
-    "typing_extensions",
-    "pyarrow",
-    "pandas",
-    "anndata>=0.8",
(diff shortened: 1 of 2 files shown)


pablo-gar:
Also fixes #418

pablo-gar (approved):

atolopko-czi (approved):
LGTM

PR #609, opened 2023-07-07
Expose out-of-core Mean and Variance APIs

### LINKED ISSUE #602: Expose out-of-core Mean and Variance APIs
**Product requirements**

A user can obtain col-based mean and/or variance from an ExperimentAxisQuery with the following API design

`cellxgene_census.experimental.pp.mean_variance`:

```
def mean_variance(
    query: soma.ExperimentAxisQuery,
    layer: str = "raw",
    axis: int = 0,
    calculate_mean: bool = False,
    calculate_variance: bool = False,
    ddof: int = 1,
) -> pandas.DataFrame:

    """
    Calculate  mean and/or variance along the `obs` axis for all features (`var`) from a query result. Calculations
    are done in an accumulative chunked fashion. 

    Args:
        query:
            A SOMA query, specifying the obs/var selection over which mean and variance are calculated.

        layer:
            X layer used, e.g., `raw`

        axis:
           Axis or axes along which the means are computed

        calculate_mean:
            If `True it calculates mean, otherwise skips calculation

        calculate_variance:
            If `True it calculates variance, otherwise skips calculation
      
        ddof:
        “Delta Degrees of Freedom”: the divisor used in the calculation for variance is N - ddof, where N represents the number of elements.

    Returns:
        Pandas DataFrame indexed by the obs `soma_joinid` and with columns 
       `mean` (if `calculate_mean = True`), and `variance` (if `calculate_variance = True`)

    Lifecycle:
        experimental
    """
```

ebezzi:
Fixes #602 

The notebook can be viewed [here](https://github.com/chanzuckerberg/cellxgene-census/blob/834f4bc75629707fb95ee48cc6e7c2c1bfad6ae9/api/python/notebooks/experimental/mean_variance.ipynb).


bkmartinjr (commented):
Please add a notebook demonstrating usage.

bkmartinjr:
@pablo-gar - IMHO, the API signature is missing two important features (which are not much work to add):
* axis: int - which axis to compute over
* ddof: int - very important for our use case
These are not much work, and are well worth the effort

api/python/cellxgene_census/tests/experimental/pp/test_stats.py
```diff
@@ -0,0 +1,144 @@
+import math
```
bkmartinjr, line 1: I noticed that test_online.py still exists, which tests MeanVarianceAccumulator directly.  

Suggest adding a test for MeanAccumulator in that file so you directly test the lower level implementation.

api/python/cellxgene_census/tests/experimental/pp/test_stats.py
```diff
@@ -0,0 +1,144 @@
...
+import pandas as pd
+import pytest
+import tiledbsoma as soma
+
+import cellxgene_census
+from cellxgene_census.experimental import pp
+
+
+@pytest.fixture
+def small_mem_context() -> soma.SOMATileDBContext:
```
bkmartinjr, line 12: should move into conftest to keep it DRY (this is a cut & paste of the code in test_hvg.py)

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if calculate_variance:
+        mvn = MeanVarianceAccumulator(n_batches, n_samples, query.n_obs)
+        for obs_dim, data in iterate():
+            mvn.update_single_batch(obs_dim, data)
+            _, _, all_u, all_var = mvn.finalize()
+            if calculate_mean:
+                result["mean"] = all_u
+            result["variance"] = all_var
+        del mvn
```
bkmartinjr, line 74: unnecessary - the GC will do the same upon return

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -0,0 +1,83 @@
...
+    else:
+        mn = MeanAccumulator(n_batches, n_samples, query.n_obs)
+        for obs_dim, data in iterate():
+            mn.update_single_batch(obs_dim, data)
+            _, all_u = mn.finalize()
+            result["mean"] = all_u
+        del mn
```
bkmartinjr, line 81: unnecessary - the GC will do the same upon return

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -0,0 +1,83 @@
...
+
+    n_batches = 1
+    n_samples = np.array([query.n_vars], dtype=np.int64)
+
+    def iterate() -> Generator[Tuple[npt.NDArray[np.int64], Any], None, None]:
+        max_workers = (os.cpu_count() or 4) + 2
+        with futures.ThreadPoolExecutor(max_workers=max_workers) as pool:
+            for arrow_tbl in _EagerIterator(query.X(layer).tables(), pool=pool):
+                obs_indexer = query.indexer
+                obs_dim = obs_indexer.by_obs(arrow_tbl["soma_dim_0"])
```
bkmartinjr, line 58: performance note:  using the query indexer creates a Pandas.Index object. So you are creating two duplicate Index objects (one here, one on line 63), which are identical, and which are reasonably expensive if it is a large query.

As an alternative, create your own Index (i.e., line 63), and then use it to reindex.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -56,6 +56,33 @@ def finalize(
         return batches_u, batches_var, all_u, all_var
 
 
+class MeanAccumulator:
```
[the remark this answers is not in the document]
ebezzi: I agree. I implemented batches (although only partially) mostly because the code was already there, and this could be useful for future usages. But I agree with your statement. If Pablo doesn't think that batches are useful, it's probably worth simplifying this to use numpy as you suggest.

api/python/cellxgene_census/tests/experimental/pp/test_stats.py
```diff
@@ -0,0 +1,144 @@
...
+            test_soma_joinid = int(mean_variance.index[0])  # type: ignore
+
+            with census["census_data"][experiment_name].axis_query(
+                measurement_name="RNA", obs_query=soma.AxisQuery(coords=([test_soma_joinid],))
+            ) as row_query:
+                sparse_row = row_query.X("raw").coos((1 + test_soma_joinid, row_query.n_vars)).concat()
+                sparse_mat = sparse_row.to_scipy()
+                row = sparse_mat.tocsr().getrow(test_soma_joinid)
+
+                assert math.isclose(
```
bkmartinjr, line 103: I have been using numpy.allclose().  If you use that, you can check _all_ of the results, not just the first one in the dataframe.  That seems like a large win in case only the first item is correct :-)

api/python/cellxgene_census/tests/experimental/pp/test_stats.py
```diff
@@ -0,0 +1,144 @@
...
+    obs_value_filter: str,
+    small_mem_context: soma.SOMATileDBContext,
+) -> None:
+    with cellxgene_census.open_soma(census_version="latest", context=small_mem_context) as census:
+        with census["census_data"][experiment_name].axis_query(
+            measurement_name="RNA", obs_query=soma.AxisQuery(value_filter=obs_value_filter)
+        ) as query:
+            mean_variance = pp.mean_variance(query, calculate_mean=True, calculate_variance=False)
+            assert isinstance(mean_variance, pd.DataFrame)
+            assert "mean" in mean_variance
```
bkmartinjr, line 89: I would also check their dtypes

api/python/cellxgene_census/tests/experimental/pp/test_stats.py
```diff
@@ -0,0 +1,144 @@
...
+            assert "variance" not in mean_variance
+            assert mean_variance.index.name == "soma_joinid"
```
bkmartinjr, line 91: also check that the index values match the query obs_joinids

api/python/cellxgene_census/tests/experimental/pp/test_stats.py
```diff
@@ -0,0 +1,144 @@
...
+
+            # Pick one element from the dataframe
```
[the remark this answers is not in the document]
bkmartinjr: Oops - I put this on the `mean` test.  It was intended for the `variance` test!

api/python/cellxgene_census/tests/experimental/pp/test_stats.py
```diff
@@ -0,0 +1,144 @@
...
+
+            with census["census_data"][experiment_name].axis_query(
+                measurement_name="RNA", obs_query=soma.AxisQuery(coords=([test_soma_joinid],))
+            ) as row_query:
+                sparse_row = row_query.X("raw").coos((1 + test_soma_joinid, row_query.n_vars)).concat()
+                sparse_mat = sparse_row.to_scipy()
+                row = sparse_mat.tocsr().getrow(test_soma_joinid)
+
+                assert math.isclose(
+                    row.todense().mean(), mean_variance.loc[test_soma_joinid, "mean"], rel_tol=0.01  # type: ignore
```
bkmartinjr, line 104: You can compute `mean` for a scipy sparse matrix with its built-in (no need to densify, etc.).

See [docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_array.mean.html#scipy.sparse.csr_array.mean)

api/python/cellxgene_census/tests/experimental/pp/test_stats.py
```diff
@@ -0,0 +1,144 @@
+import math
+
+import pandas as pd
+import pytest
+import tiledbsoma as soma
+
+import cellxgene_census
+from cellxgene_census.experimental import pp
+
```
bkmartinjr, line 9: Test cases are missing corner/error cases, e.g.,
* empty query result
* bad params raises expected errors
etc

bkmartinjr (changes requested):
several concerns to look at:
* API signature:  missing ddof and axis.  CC: @pablo-gar 
* need a demo notebook
* tests need improvement (comments are all inline)
* the MeanAccumulator can be drastically simplified if we are not supporting batches (and I don't see why we need them - we only have them in MeanVarianceAccumulator due to the HVG use case).
* performance issue in mean_variance where it double-allocates an Index

I'd like to do another review once we get through these issues.

ebezzi:
> several concerns to look at:
> 
> * API signature:  missing ddof and axis.  CC: @pablo-gar
> * need a demo notebook
> * tests need improvement (comments are all inline)
> * the MeanAccumulator can be drastically simplified if we are not supporting batches (and I don't see why we need them - we only have them in MeanVarianceAccumulator due to the HVG use case).
> * performance issue in mean_variance where it double-allocates an Index
> 
> I'd like to do another review once we get through these issues.

Thanks Bruce! I agree with all of your points. I will wait for Pablo to chime in and then I'll do another pass.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
ebezzi: I see that the duplicate index is available as `query.indexer._obs_index` [which is private](https://github.com/single-cell-data/SOMA/blob/063aef6a32530c2f2730c73942490d3fc55916d3/python-spec/src/somacore/query/query.py#L525) - are we ok with using it even if private?

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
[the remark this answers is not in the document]
ebezzi: Simplified as requested.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -56,6 +56,22 @@ def finalize(
...
+    def __init__(self, n_samples: int, n_variables: int):
+        if n_samples <= 0:
+            raise ValueError("No samples provided - can't calculate mean.")
+
+        self.u = np.zeros(n_variables, dtype=np.float64)
+        self.n_samples = n_samples
+
+    def update(self, var_vec: npt.NDArray[np.int64], val_vec: npt.NDArray[np.float32]) -> None:
+        for col, val in zip(var_vec, val_vec):
```
ebezzi, line 68: @bkmartinjr To the best of my knowledge, this part of the mean calculation could still benefit from numba. Is there any specific numpy function you had in mind that could avoid the loop?

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -0,0 +1,103 @@
...
+import tiledbsoma as soma
+
+from ..util._eager_iter import _EagerIterator
+from ._online import MeanAccumulator, MeanVarianceAccumulator
+
+
+def mean_variance(
+    query: soma.ExperimentAxisQuery,
+    layer: str = "raw",
+    axis: int = 0,
```
ebezzi, line 19: This follows the numpy convention where `axis=0` is by column. Are we ok with that?

pablo-gar: Yup!

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
ebezzi: I saw that but `by_obs` effectively forces a reindex, while here I need the "raw" `soma_joinid` index.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
bkmartinjr: Yep, as written it will be slow.  You can either:
* put that code in a numba.jitted function and call it.  Numba will make it fast
* or, rewrite to use Numpy advanced indexing
the latter is:

> self.u[var_vec] = self.u[var_vec] + val_vec

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -40,22 +41,38 @@ def update_single_batch(self, var_vec: npt.NDArray[np.int64], val_vec: npt.NDArr
...
+class MeanAccumulator:
+    def __init__(self, n_samples: int, n_variables: int):
+        if n_samples <= 0:
+            raise ValueError("No samples provided - can't calculate mean.")
+
+        self.u = np.zeros(n_variables, dtype=np.float64)
+        self.n_samples = n_samples
+
+    def update(self, var_vec: npt.NDArray[np.int64], val_vec: npt.NDArray[np.float32]) -> None:
+        _reduce_mean_vector(self.u, var_vec, val_vec)
```
bkmartinjr, line 70: lets pick a single naming scheme (I'm fine with `reduce`, but lets make the other reducers - now called `update_*` match)

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -40,22 +41,38 @@ def update_single_batch(self, var_vec: npt.NDArray[np.int64], val_vec: npt.NDArr
...
         # correct each batch to account for sparsity
         _mbomv_sparse_correct_batches(self.n_batches, self.n_samples, self.n, self.u, self.M2)
 
         # compute u, var for each batch
         batches_u = self.u
-        batches_var = (self.M2.T / (self.n_samples - 1)).T
+
+        # Note: if N-ddof is less than or equal to 0, we will return Inf - this is consistent
+        # with the numpy.var behavior.
+        batches_var = (self.M2.T / np.maximum(np.zeros(self.n_samples.size), (self.n_samples - self.ddof))).T
```
bkmartinjr, line 52: why does this allocate an entire array of zeros for the `minimum` operation? Waste of memory.

Just use normal ufunc semantics and do something like:
> np.maximum(0,  (self.n_samples - self.ddof))).T

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -40,22 +41,38 @@ def update_single_batch(self, var_vec: npt.NDArray[np.int64], val_vec: npt.NDArr
...
 
         # accum all batches using Chan's
         all_u, all_M2 = _mbomv_combine_batches(self.n_batches, self.n_samples, self.u, self.M2)
-        all_var = all_M2 / max(1, (self.n_samples.sum() - 1))
+        all_var = all_M2 / max(0, (self.n_samples.sum() - self.ddof))
```
bkmartinjr, line 56: why python `max` and not numpy?

bkmartinjr: Also, given that this will (intentionally) divide by zero, shouldn't this be wrapped in a numpy error context?

ebezzi: `max` was already used, I guess because the denominator is just a scalar here?

For the division by zero - it's a good point, I will add it.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Returns:
+        Pandas DataFrame indexed by the obs `soma_joinid` and with columns
+       `mean` (if `calculate_mean = True`), and `variance` (if `calculate_variance = True`)
+
+    Lifecycle:
+        experimental
+    """
+
+    if axis not in [0, 1]:
```
bkmartinjr, line 55: tiny, tiny nit: use tuple not list if it is immutable

api/python/cellxgene_census/tests/experimental/pp/test_online.py
```diff
@@ -23,7 +23,8 @@ def matrix(m: int, n: int) -> sparse.coo_matrix:
 @pytest.mark.parametrize("stride", [101, 53])
 @pytest.mark.parametrize("n_batches", [1, 3, 11, 101])
 @pytest.mark.parametrize("m,n", [(1200, 511), (100001, 57)])
-def test_meanvar(matrix: sparse.coo_matrix, n_batches: int, stride: int) -> None:
+@pytest.mark.parametrize("ddof", [1, 100, 10000])
```
bkmartinjr, line 26: I would add zero.  if the list gets too long, the very large ddof values are of marginal utility IMHO

api/python/cellxgene_census/tests/experimental/pp/test_online.py
```diff
@@ -3,7 +3,7 @@
 import pytest
```
bkmartinjr, line 3: Can you add some test cases for the expected error conditions?  For example, using MeanAccumulator with n_samples==0 or n_variables==0

api/python/cellxgene_census/tests/experimental/pp/test_stats.py
```diff
@@ -0,0 +1,184 @@
...
+from scipy import sparse
+
+import cellxgene_census
+from cellxgene_census.experimental import pp
+
+
+@pytest.mark.experimental
+@pytest.mark.live_corpus
+@pytest.mark.parametrize("axis", [0, 1])
+@pytest.mark.parametrize(
```
bkmartinjr, line 16: Missing edge case that needs testing:   query that returns nothing

api/python/notebooks/experimental/mean_variance.ipynb
```diff
@@ -0,0 +1,236 @@
+{
```
bkmartinjr, line 1: notebook comments:
* blank cell at the end of the file should be deleted
* suggest adding example cell at end that joins with the axis dataframe (eg. obs) and displays the mean/var in that context. This will be the typical usage so we should highlight it.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
bkmartinjr: I don't have a preference beyond consistency - so you can decide!

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
bkmartinjr: Re: max - you are right, it isn't a vector operation. Sorry - I misread the code.  OK to leave as is - I don't think it matters at all.

bkmartinjr (changes requested):
Second pass review - we are super close.  My only concerns are now small items related to code clarity, etc.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
ebezzi: On a second thought, I believe we don't need to add an additional error context since it will inherit the one from the division by zero (`np.var` has the same behavior - raises a division by zero warning). What we could do is to add a specific warning that states that ddof must be >= N. Is that desirable?

bkmartinjr: Not sure I understand the point about inheriting an existing numpy error context (where is that defined?)   And I don't think we should add warnings as the behavior is well specified.

What I meant was that the code should look like:
```
with np.errstate(divide='ignore'):
    all_var = all_M2 / max(0, (self.n_samples.sum() - self.ddof))
```

api/python/cellxgene_census/tests/experimental/pp/test_stats.py
ebezzi: There is a dedicated test for that: `test_mean_variance_empty_query`

bkmartinjr: so sorry - I missed that!

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_highly_variable_genes.py
```diff
@@ -88,14 +88,13 @@ def _highly_variable_genes_seurat_v3(
...
             if batch_indexer:
                 _batch_take_at_future = pool.submit(batch_indexer, arrow_tbl["soma_dim_0"])
-                var_dim = var_indexer.by_var(arrow_tbl["soma_dim_1"])
                 _batch_vec = batch_codes[_batch_take_at_future.result()]
-                mvn.update_by_batch(_batch_vec, var_dim, data)
             else:
-                var_dim = var_indexer.by_var(arrow_tbl["soma_dim_1"])
-                mvn.update_single_batch(var_dim, data)
+                _batch_vec = None
+            mvn.update(var_dim, data, _batch_vec)
```
ebezzi, line 97: I simplified the code a bit here, so that the `update` method is only called once.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -130,13 +160,22 @@ def _mbomv_update_by_batch(
...
 
 @numba.jit(
-    numba.void(
-        numba.types.Array(numba.int64, 1, "C", readonly=True),
-        numba.types.Array(numba.float32, 1, "C", readonly=True),
-        numba.int32[:, :],
-        numba.float64[:, :],
-        numba.float64[:, :],
-    ),
+    [
```
ebezzi, line 163: This is required because previously `test_meanvar` was hitting the batched case even for n_batches=1. With the changes, it's now hitting this code and the type constraints didn't match the test case where `var_vec` is an int32.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_highly_variable_genes.py
bkmartinjr: this simplification removes the parallelism (by moving the var reindexing up, it completely serializes everything - the thread submit of the row indexer is wasted overhead).

Either one of these is fine (although the second has a performance hit):
1. move the var reindexing so it remains in-parallel with row indexing
2. or, get rid of the thread submit for row indexing

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_highly_variable_genes.py
```diff
@@ -154,14 +153,13 @@ def _highly_variable_genes_seurat_v3(
         acc = CountsAccumulator(n_batches, query.n_vars, clip_val)
         for arrow_tbl in _EagerIterator(query.X(layer).tables(), pool=pool):
             data = arrow_tbl["soma_data"].to_numpy()
+            var_dim = var_indexer.by_var(arrow_tbl["soma_dim_1"])
```
bkmartinjr, line 156: same bug is introduced as before ...

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
bkmartinjr: really good catch!!!  Thank you.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_highly_variable_genes.py
ebezzi: Ah, that's interesting to know. I will fix it. Thanks!

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -0,0 +1,99 @@
...
+    n_dim_1 = query.n_vars if axis == 1 else query.n_obs
+
+    n_batches = 1
+    n_samples = np.array([n_dim_1], dtype=np.int64)
+
+    idx = pd.Index(data=query.obs_joinids() if axis == 1 else query.var_joinids(), name="soma_joinid")
+
+    def iterate() -> Generator[Tuple[npt.NDArray[np.int64], Any], None, None]:
+        max_workers = (os.cpu_count() or 4) + 2
+        with futures.ThreadPoolExecutor(max_workers=max_workers) as pool:
```
bkmartinjr, line 74: small simplification (nit, optional):  EagerIterator can use at most a single thread.  So you could remove the code that computes max_workers (line 73), and just hard-wire a `max_workers=1` into line 74.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
bkmartinjr: actually, don't we have the same issue on all of the routines that accept coordinates - they are all wired to only accept int64 (I did this originally because I assumed they would all be soma_joinids, but you are doing the right thing by assumig they could be int32s in some cases).  so I see the same issue in _accum_clipped_counts and _accum_clipped_counts_by_batch.

api/python/cellxgene_census/tests/experimental/pp/test_online.py
```diff
@@ -3,7 +3,7 @@
...
 from scipy import sparse
 
-from cellxgene_census.experimental.pp._online import MeanVarianceAccumulator
+from cellxgene_census.experimental.pp._online import MeanAccumulator, MeanVarianceAccumulator
```
bkmartinjr, line 6: It is totally *not* your bug, but could you add a simple / quick test of CountsAccumulator?

ebezzi: Sure!

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
ebezzi: I think that's because `CountAccumulator` lacks an explicit test, as you pointed out. I will add it and we'll likely need to extend this change there.

bkmartinjr (approved):
LGTM

api/python/notebooks/experimental/mean_variance.ipynb
```diff
@@ -0,0 +1,726 @@
...
+   "execution_count": 13,
+   "id": "a06549db-60c5-49b6-b57d-29ee6db2c61c",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "# Import packages\n",
+    "import cellxgene_census\n",
+    "from cellxgene_census.experimental.pp import mean_variance"
```
pablo-gar, line 34: `import somacore a soma`

api/python/notebooks/experimental/mean_variance.ipynb
```diff
@@ -0,0 +1,726 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "c4b089f9-5304-4479-b739-20792e12595a",
+   "metadata": {},
+   "source": [
+    "# Out-of-core mean and variance calculation\n",
+    "\n",
+    "This tutorial describes use of the cellxgene_census.experimental.pp API for calculating out-of-core mean and variance in the Census. The variance calculation is performed using [Welford's online algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance).\n",
+    "\n",
+    "There is a single API available: \n",
```
pablo-gar, line 12: Please convert this into a header:

`## The mean and variance API`

api/python/notebooks/experimental/mean_variance.ipynb
```diff
@@ -0,0 +1,726 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "c4b089f9-5304-4479-b739-20792e12595a",
+   "metadata": {},
+   "source": [
+    "# Out-of-core mean and variance calculation\n",
+    "\n",
+    "This tutorial describes use of the cellxgene_census.experimental.pp API for calculating out-of-core mean and variance in the Census. The variance calculation is performed using [Welford's online algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance).\n",
+    "\n",
```
pablo-gar, line 11: See [Census API notebook/vignette editorial guidelines](https://github.com/chanzuckerberg/cellxgene-census/blob/main/docs/census_notebook_guidelines.md)

Please add a table of contents:

```
**Contents**

1. The mean and variance API.
2. Example: calculate mean and variance for a slice of Census.
```

api/python/notebooks/experimental/mean_variance.ipynb
```diff
@@ -0,0 +1,726 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c5385575-d591-4c5c-8bc3-aea4ea02a8da",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "## Calculate mean and variance"
```
pablo-gar, line 44:
```suggestion
    "## Example: calculate mean and variance for a slice of Census"
```

api/python/notebooks/experimental/mean_variance.ipynb
```diff
@@ -0,0 +1,726 @@
...
+    "        measurement_name=\"RNA\", obs_query=soma.AxisQuery(value_filter=obs_value_filter)\n",
+    "    ) as query:\n",
+    "        mv_df = mean_variance(\n",
+    "            query,\n",
+    "            axis=1,\n",
+    "            calculate_mean=True,\n",
+    "            calculate_variance=True,\n",
+    "        )\n",
+    "\n",
+    "    obs_df = census[\"census_data\"][experiment_name].obs.read(coords=(query.obs_joinids(),)).concat().to_pandas()\n",
```
pablo-gar, line 203: since you just did a query you should just get the obs from the query, is cleaner

```
obs_df = query.obs().concat().to_pandas()
```

api/python/notebooks/experimental/mean_variance.ipynb
bkmartinjr: You probably want `import tiledbsoma as soma`.  You almost never want to import somacore directly.

pablo-gar: oh yeah, that was my bad typing ... I had the right version on my own testing code (facepalm)

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -0,0 +1,97 @@
...
+    query: soma.ExperimentAxisQuery,
+    layer: str = "raw",
+    axis: int = 0,
+    calculate_mean: bool = False,
+    calculate_variance: bool = False,
+    ddof: int = 1,
+) -> pd.DataFrame:
+    """
+    Calculate  mean and/or variance along the `obs` axis from query results. Calculations
+    are done in an accumulative chunked fashion.
```
pablo-gar, line 25: Let's clarify the denominator for our calculation, some users may expect the average of non-zero values. Which we should implement shortly after landing this PR.

```suggestion
    are done in an accumulative chunked fashion. For the mean and variance calculations, the total number of elements (N) is the corresponding dimension size: for column-wise calculations (`axis = 0`) N is number of rows, for row-wise calculations (`axis = 1`)  N is number of columns.
```

pablo-gar (approved):
Please incorporate my edits, and run the notebook beginning to end in one run to have contiguous cell numbers.

Then go ahead and merge!

api/python/notebooks/experimental/mean_variance.ipynb
```diff
@@ -0,0 +1,726 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "c4b089f9-5304-4479-b739-20792e12595a",
+   "metadata": {},
+   "source": [
+    "# Out-of-core mean and variance calculation\n",
```
pablo-gar, line 8:
```suggestion
    "# Out-of-core (incremental) mean and variance calculation\n",
```

api/python/notebooks/experimental/mean_variance.ipynb
ebezzi: Thanks, I forgot that `query` has an `obs()` method!

PR #625, opened 2023-07-19
hard error if the builder receives an unsupported CXG schema version

bkmartinjr:
This PR tightens up error detection for a "should never happen" case where we receive an unsupported CxG schema version in an H5AD file.  The previous behavior was to log a warning and drop the dataset.  The new behavior is to fail the build.

There are no cases of such warning in the logs for the past few weekly builds, and our process is such that it should not occur. This is just an additional sanity check.

### DIFF
--- a/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/anndata.py
+++ b/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/anndata.py
@@ -42,8 +42,9 @@ def open_anndata(
         if h5ad.schema_version == "":
             h5ad.schema_version = get_cellxgene_schema_version(ad)
         if h5ad.schema_version not in CXG_SCHEMA_VERSION_IMPORT:
-            logging.warning(f"H5AD has old schema version, skipping {h5ad.dataset_h5ad_path}")
-            continue
+            msg = f"H5AD has unsupported schema version {h5ad.dataset_h5ad_path}, expected {CXG_SCHEMA_VERSION}"
+            logging.error(msg)
+            raise RuntimeError(msg)
 
         # Multi-organism datasets - any dataset with 2+ feature_reference organisms is ignored,
         # exclusive of values in FEATURE_REFERENCE_IGNORE. See also, cell filter for mismatched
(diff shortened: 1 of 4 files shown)


tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/manifest.py
```diff
@@ -62,8 +62,9 @@ def load_manifest_from_CxG() -> List[Dataset]:
         schema_version = dataset["schema_version"]
 
         if schema_version not in CXG_SCHEMA_VERSION_IMPORT:
-            logging.warning(f"Dropping dataset {dataset_id} due to unsupported schema version")
-            continue
+            msg = f"Dropping dataset {dataset_id} due to unsupported schema version"
```
ebezzi, line 65: I think the message should reflect the new behavior. Something like `Aborting build due to {dataset_id} having unsupported version {schema_version}`.

bkmartinjr: doh!   Will fix.  TY

bkmartinjr: this also reminds me that we need to check for this in the H5AD itself.  Will add that as well.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/anndata.py
```diff
@@ -42,8 +42,9 @@ def open_anndata(
         if h5ad.schema_version == "":
             h5ad.schema_version = get_cellxgene_schema_version(ad)
         if h5ad.schema_version not in CXG_SCHEMA_VERSION_IMPORT:
-            logging.warning(f"H5AD has old schema version, skipping {h5ad.dataset_h5ad_path}")
-            continue
+            msg = f"H5AD has unsupported schema version {h5ad.dataset_h5ad_path}, expected {CXG_SCHEMA_VERSION}"
```
ebezzi, line 45:
```suggestion
            msg = f"H5AD has unsupported schema version {h5ad.schema_version}, expected {CXG_SCHEMA_VERSION}"
```

bkmartinjr: I think we want all three (dataset_id, schema_version, expected schema version).  Will push a change.

ebezzi (approved):

atolopko-czi (approved):
LGTM

Issue #242, opened 2023-03-06
Protect long term support census builds


atolopko-czi:
## Motivation

Long Term Support (LTS) Census releases are intended to be maintained for 5 years.  It is critical that we do not accidentally delete these releases.

## Definition of Done

All objects of every LTS release should have AWS S3 "object lock" flag set.

## Tasks

- [ ] Update the builder's s3 upload operation to set the "object lock" flag on all uploaded S3 objects.

atolopko-czi:
Superseded by #241

atolopko-czi:
Reopening this story, since there a simple, additional level of protection we can add to LTS releases: S3 "object lock"

atolopko-czi:
> Yeah, I'm gonna reach out to the AWS support to enable it. Shall we do this directly on the public bucket?

As opposed to testing on a temporary bucket? I think it's okay to do it on the public bucket, since we're not depending upon yet, so if it were to introduce any issues, we can fix.

ebezzi:
> > Yeah, I'm gonna reach out to the AWS support to enable it. Shall we do this directly on the public bucket?
> 
> As opposed to testing on a temporary bucket? I think it's okay to do it on the public bucket, since we're not depending upon it yet, so object lock config were to introduce any issues, we can still fix.

No, as opposed to use the current bucket (the one in single-cell-prod).

pablo-gar:
Please revisit after this is addressed: [Create Census data mirrors in the top 5 most used geographical locations](https://github.com/chanzuckerberg/cellxgene-census/issues/545)

atolopko-czi:
With the upcoming 2nd LTS "emergency" census data release, @bkmartinjr realized an edge case whereby a prior LTS release could be deleted due to the 31 day deletion janitor logic, if it becomes unaliased. That will be addressed in another issue/PR. But more broadly, the concern over LTS data protection has been re-elevated. In light of that, this could be a good story to take on. Minimally, as a spike to determine whether this S3 object locking protection is easy to do. @ebezzi, I assigned to you provisionally, in case you have bandwidth. And @pablo-gar, for awareness or if you have concerns about allocating time to this.

bkmartinjr:
I will drop a PR shortly to deal with the builder GC logic (adding a "do_no_delete" flag to release.json). 

Agree that it would also be nice to have an additional layer of protection described in this issue...

ebezzi:
I'll resume looking into that.

[...]

PR #628, opened 2023-07-19
add do_not_delete flag to release.json format

bkmartinjr:
Related to #242 

Adds a new flag to the release.json format:  `do_not_delete`.  If set to a true-ish value, will prevent the automated build process from garbage collecting a build.  This is intended for use in protecting LTS releases (when the LTS is tagged, this flag should be set on the build).

If the flag is not set, or is false-ish, the build will be subject to the standard test build cleanup.

### DIFF
--- a/tools/cellxgene_census_builder/tests/test_release_cleanup.py
+++ b/tools/cellxgene_census_builder/tests/test_release_cleanup.py
@@ -14,8 +14,8 @@ def tag_days_old(days_old: int) -> str:
 TAG_NOW = tag_days_old(0)
 TAG_10D_OLD = tag_days_old(10)
 TAG_100D_OLD = tag_days_old(100)
-TAG_LTS_A = tag_days_old(1000)
-TAG_LTS_B = tag_days_old(1001)
+TAG_LTS_CURRENT_STABLE = tag_days_old(1000)
+TAG_LTS_PREVIOUS_STABLE = tag_days_old(1001)
 
 S3_PREFIX = "s3://bucket/path/"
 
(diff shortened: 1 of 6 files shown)


ebezzi (approved):

tools/cellxgene_census_builder/tests/test_release_cleanup.py
```diff
@@ -34,6 +36,18 @@ def tag_days_old(days_old: int) -> str:
...
+        "release_build": TAG_LTS_A,
+        "soma": {"uri": f"{S3_PREFIX}{TAG_LTS_A}/soma/", "s3_region": "us-west-2"},
+        "h5ads": {"uri": f"{S3_PREFIX}{TAG_LTS_A}/h5ads/", "s3_region": "us-west-2"},
+    },
+    TAG_LTS_B: {  # mock LTS release lacking an alias, but with a do_not_delete flag
+        "release_build": TAG_LTS_B,
+        "soma": {"uri": f"{S3_PREFIX}{TAG_LTS_B}/soma/", "s3_region": "us-west-2"},
+        "h5ads": {"uri": f"{S3_PREFIX}{TAG_LTS_B}/h5ads/", "s3_region": "us-west-2"},
+        "do_not_delete": True,
+    },
```
atolopko-czi, line 50: Is there a test verifying that `TAG_LTS_B` is not deleted? Hard to tell from the just the PR diff.  Maybe add an explicit `assert` somewhere showing `TAG_LTS_B` is not deleted?

tools/cellxgene_census_builder/tests/test_release_cleanup.py
```diff
@@ -14,6 +14,8 @@ def tag_days_old(days_old: int) -> str:
 TAG_NOW = tag_days_old(0)
 TAG_10D_OLD = tag_days_old(10)
 TAG_100D_OLD = tag_days_old(100)
+TAG_LTS_A = tag_days_old(1000)
+TAG_LTS_B = tag_days_old(1001)
```
atolopko-czi, line 18: Consider names:
* `TAG_LTS_CURRENT_STABLE`
* `TAG_LTS_PREVIOUS_STABLE`

atolopko-czi (approved):
Code LGTM, with a Q about test coverage.

Worth now also adding the setting of this flag to the [manual instructions](https://docs.google.com/document/d/12JwqpLWrm9_JbhEg99r7LB9sscevxL1gtliqu0s80Bo/edit#heading=h.6szkgsosfy8) for LTS releases. Although we really need to move those instructions into the repo, which could be done on this PR or a new one.

Process-wise, we should probably set the do_not_delete flag as soon as the "stable" alias is applied, rather than waiting for an LTS to be superseded by a new one. Note that the `TAG_LTS_A` test fixture suggests that we might only set `do_not_delete` when a new stable LTS is released, which is still a valid test case, but suggests a state we probably do not want in practice.

Another thought about deletion protection: If we had the process of re-aliasing "stable" to "LTS-<some_sequence_or_timestamp>", we'd get similar deletion protection. I know @pablo-gar had suggested doing something along those lines. I still think the explicitness of the `do_not_delete` flag is preferable to relying on aliases for protection. And the flag can be applied to non-LTS releases as well, if ever needed. So I think we should merge this PR, but mentioning the thought nonetheless.

bkmartinjr:
@atolopko-czi 

> Worth now also adding the setting of this flag to the manual instructions for LTS releases

Agree, and done.

> Process-wise, we should probably set the do_not_delete flag as soon as the "stable" alias is applied

Agree, and also done by the above.

> Another thought about deletion protection: If we had the process of re-aliasing "stable" to "LTS-<some_sequence_or_timestamp>", we'd get similar deletion protection

I actually think it is worth doing both.  Double protection on something like this feels complimentary: low overhead / high value.

bkmartinjr:
FYI @pablo-gar - just so you know this is being added.

PR #636, opened 2023-07-21
add a dataset blocklist

bkmartinjr:
Add a block list for the Census builder, allowing control over which dataset_ids are included in the build.

The block list is a text file containing dataset IDs (_not_ collection or dataset version IDs), one per line.  Blank lines and comments are allowed.  Comments are a full line prefixed with a `#`.

Example:
```
# a comment followed by a blank line, both ignored

e40c6272-af77-4a10-9385-62a398884f27
d6dfdef1-406d-4efb-808c-3c5eddbfe0cb
```

The builder will pull this file from the `main` branch at `cellxgene_census_builder/tools/dataset_blocklist.txt`.  If the file is not present, no datasets will be blocked.

This PR also includes an initial block list.  @pablo-gar - please confirm the IDs are correct.


pablo-gar (approved):
Just confirmed the datasets match the unresolved datasets from the unmarked collections here https://github.com/chanzuckerberg/single-cell-curation/issues/528

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/manifest.py
```diff
@@ -96,10 +98,43 @@ def load_manifest_from_CxG() -> List[Dataset]:
...
+
+
+def load_manifest(
+    manifest_fp: str | io.TextIOBase | None = None,
+    dataset_id_blocklist_uri: str | None = None,
+) -> List[Dataset]:
     """
     Load dataset manifest from the file pointer if provided, else bootstrap
-    the load rom the CELLxGENE REST API.
+    the from the CELLxGENE REST API.  Apply the blocklist if provided.
```
ebezzi, line 137:
```suggestion
    from the CELLxGENE REST API.  Apply the blocklist if provided.
```

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
```diff
@@ -48,6 +48,9 @@
     # Release clean up
     "release_cleanup_days": 32,  # Census builds older than this are deleted
     #
+    # Block list of dataset IDs
+    "dataset_id_blocklist_uri": "https://raw.githubusercontent.com/chanzuckerberg/cellxgene-census/main/tools/cellxgene_census_builder/dataset_blocklist.txt",
```
ebezzi, line 52: Small concern: if the GitHub repo structure changes, this will likely break and might go unnoticed. Ideas:
1. Move to S3
2. Add it to the Docker image and read it from a file
3. Add a test that reads from this file specifically, so that in case it gets moved the tests will fail (this has to be done in a subsequent PR since the file isn't in main yet)

ebezzi (approved):

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/manifest.py
```diff
@@ -96,10 +98,43 @@ def load_manifest_from_CxG() -> List[Dataset]:
...
+
+
+def apply_blocklist(datasets: List[Dataset], dataset_id_blocklist_uri: str | None) -> List[Dataset]:
+    try:
+        blocked_dataset_ids = load_blocklist(dataset_id_blocklist_uri)
+        return list(filter(lambda d: d.dataset_id not in blocked_dataset_ids, datasets))
+
+    except FileNotFoundError:
+        # Blocklist may not exist, so just skip the filtering in this case
+        logging.info("No dataset blocklist found - skipping")
```
atolopko-czi, line 127: I think we should error on this instead. If a blocklist is specified and it doesn't exist, that seems like an configuration error. And means datasets are included unexpectedly. The blocklist will be immediately available upon this PR being merged.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_state.py
atolopko-czi: Docker image solution could work. I know Bruce wanted to decouple builder releases from this data-only config: "the lifecycle of the block list is separated from the software release cycle and/or builder docker image build." But updating the blocklist file is going to create a new builder image anyway, right?

I considered the S3 solution as well during earlier Slack discussion, but felt we didn't have a great spot for it. I don't think we want it the Census bucket(s), if it ends up being visible to users, right?

Adding a test will only catch the file error at an arbitrary time, based upon development activity, which may not be in time for the next builder run. So probably not worth doing.

I'm also a tad concerned about Github file read design, but only because it makes our production system (builder) dependent on Github availability. It's not a great risk, but I've seen prod systems go down due to this. But since it's a batch process, I'm not as worried.

Overall, I'm okay with sticking with the design as is, but embedding in the Docker image seems fine to me as well and removes the runtime Github dependency.

One advantage of GitHub is the versioning of the file. S3 would provide versioning too, but not with changelogs and  authorship attribution.

tools/cellxgene_census_builder/tests/test_manifest.py
```diff
@@ -37,6 +37,16 @@ def test_load_manifest_does_dedup(manifest_csv_with_duplicates: str) -> None:
...
 
 
+def test_manifest_dataset_block(tmp_path: pathlib.Path, manifest_csv: str) -> None:
+    blocklist_fname = f"{tmp_path}/blocklist.txt"
+    blocklist_content = """# a comment\n\ndataset_id_1\n\n"""
+    with open(blocklist_fname, "w+") as f:
+        f.writelines(blocklist_content.strip())
+
+    manifest = load_manifest(manifest_csv, blocklist_fname)
+    assert len(manifest) == 1
```
atolopko-czi, line 47: nit: I would also assert explicitly that `dataset_id_1` is not admitted, and maybe generate the blocklist by taking a known dataset id from the manifest itself (in case the manifest csv test fixture is changed in the future).

atolopko-czi (changes requested):

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/manifest.py
bkmartinjr: will add

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/manifest.py
bkmartinjr: will fix

.pre-commit-config.yaml
```diff
@@ -47,7 +47,7 @@ repos:
       - id: mypy
         name: mypy-tools-cellxgene_census_builder
         files: ^tools/cellxgene_census_builder
-        args: ["--config-file=./tools/cellxgene_census_builder/pyproject.toml"]
+        args: ["--config-file=./tools/cellxgene_census_builder/pyproject.toml", "--python-version=3.10"]
```
bkmartinjr, line 50: allows use of Python 3.10 syntax in builder

tools/cellxgene_census_builder/dataset_blocklist.txt
jahilton, line 1: `e40c6272-af77-4a10-9385-62a398884f27` & `d6dfdef1-406d-4efb-808c-3c5eddbfe0cb` are in the same Collection as one with a question, but we have no indication that these 2 are impacted

jahilton: That's my suggestion unless @pablo-gar has any hesitations about those datasets.

atolopko-czi (approved):
LGTM!

tools/cellxgene_census_builder/dataset_blocklist.txt
pablo-gar: my reasoning is to remove all datasets from those collections, as cherry picking the offending datasets may introduce some confusion for anyone trying to access all data for those collections.

ebezzi (approved):

tools/cellxgene_census_builder/dataset_blocklist.txt
jahilton: That reasoning seems at odds with Collections that have some, but not all, datasets that meet Census criteria. i.e. we don't make any promise that Census users will have access to all data from a Collection.
On the other hand, I could blocking all 3 datasets with the justification that we know something is _off_ with some of the data in this study, there's no telling that the issue isn't seeping into the other two datasets and we just don't see the same symptom

PR #637, opened 2023-07-21
dataset blocklist location check

bkmartinjr:
Related to #636 

Enforce location and contents of block list.  This unit test is intended as a warning signal that the builder config and blocklist location are out of sync.

NOTE: this PR will fail unit tests until #636 is merged to main.


tools/cellxgene_census_builder/tests/test_manifest.py
```diff
@@ -183,3 +186,28 @@ def test_load_manifest_from_cxg_excludes_datasets_with_no_assets(empty_blocklist
...
         with pytest.raises(RuntimeError):
             load_manifest(None, empty_blocklist)
+
+
+def test_blocklist_alive_and_well() -> None:
+    """
+    Perform three checks:
+    1. Block list is specified in the default configuration
+    2. The file exists at the specified location
+    3. The file "looks like a block list
```
atolopko-czi, line 196:
```suggestion
    3. The file "looks like" a block list
```

tools/cellxgene_census_builder/tests/test_manifest.py
```diff
@@ -183,3 +186,28 @@ def test_load_manifest_from_cxg_excludes_datasets_with_no_assets(empty_blocklist
...
+    with fsspec.open(dataset_id_blocklist_uri, "rt") as fp:
+        for line in fp:
+            # each line must be a comment, blank or a UUID
+            line = line.strip()
+            if not line or line.startswith("#"):
+                continue
+
+            # UUID() raises ValueError upon malformed UUID
+            # Equality check enforces formatting (i.e., dashes)
+            assert line == str(uuid.UUID(hex=line))
```
atolopko-czi, line 213: If the file location (and config value pointing to it) is moved in a PR, this will fail until the PR is merged, right? Since it's checking the file on GH in main branch. Not sure how to make this work...

atolopko-czi (changes requested):

tools/cellxgene_census_builder/tests/test_manifest.py
ebezzi: I don't think it's a big deal, since the files won't move often. I would rather have a test failure than a silent failure on build day, anyway!

ebezzi (approved):

tools/cellxgene_census_builder/tests/test_manifest.py
atolopko-czi: Should at least note in a comment here, "if file moves, this will fail until merged to main".  Could side-step by temporarily copying the file to new location in a PR, then change the config and delete old file in the old location. 🤯

bkmartinjr: will add

atolopko-czi (approved):

Issue #518, opened 2023-06-05
Integrate schema gated dataset fetch with Census Builder


brianraymor:

pablo-gar:
Dup of #89 maybe?

brianraymor:
@pablo-gar - it's not precisely a dup, but it's definitely related.

#89 is about the census builder making copies now that the Discover API supports permanent dataset version urls. And depending on where we land with the deletion policy, there may be some advantages for census to keep making copies. 

This issue is about **how** the census builder requests the datasets. This issue allows census builder to keep requesting schema 3 versions until it's ready for schema 4 versions.

PR #650, opened 2023-07-26
Update builder to use schema_version param to pin dataset CxG schema versions to 3.0.0

nayib-jose-gloria:
#518

### DIFF
--- a/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/manifest.py
+++ b/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/manifest.py
@@ -7,7 +7,7 @@ from typing import List, Set
 import fsspec
 
 from .datasets import Dataset
-from .globals import CXG_SCHEMA_VERSION_IMPORT
+from .globals import CXG_SCHEMA_VERSION, CXG_SCHEMA_VERSION_IMPORT
 from .util import fetch_json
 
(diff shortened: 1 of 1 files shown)


tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/manifest.py
```diff
@@ -54,7 +54,7 @@ def load_manifest_from_CxG() -> List[Dataset]:
     logging.info("Loading manifest from CELLxGENE data portal...")
 
     # Load all collections and extract dataset_id
-    datasets = fetch_json(f"{CXG_BASE_URI}curation/v1/datasets")
+    datasets = fetch_json(f"{CXG_BASE_URI}curation/v1/datasets?schema_version=3.0.0")
```
ebezzi, line 57: There is a `CXG_SCHEMA_VERSION` global variable in globals.py - we should use that one.

nayib-jose-gloria: There's also CXG_SCHEMA_VERSION_IMPORT. I think it can make sense to have both (one to pin to a major/minor version if you want to do that eventually, one to list supported patch versions) but they should probably have more distinct names. Going to leave as-is for now.

ebezzi (approved):

PR #640, opened 2023-07-25
Add `get_single_cell_experiment`

pablo-gar:
Addresses #641 

- Adds `get_single_cell_experiment`
- Adds tests
- Adds examples to `get_single_cell_experiment` and `get_seurat`
- Updates docs
- Adds `var_index` arg to `get_seurat`

PR #658, opened 2023-07-31
[r] add examples to function docs

mlin:
#648

Add remaining examples (except for `get_seurat()` which is covered in #640)


pablo-gar:
Merged in #583

PR #675, opened 2023-08-01
Fetch docsite version number from git tags

ebezzi:
Instead of getting the version from the (local) census version, just get the most recent tag from git. This will match the latest released version.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -10,9 +10,12 @@ project = 'cellxgene-census'
 copyright = '2022-2023 Chan Zuckerberg Initiative'
 author = 'Chan Zuckerberg Initiative'
 
-import cellxgene_census
+import git
+repo = git.Repo(search_parent_directories=True)
+tags = sorted(repo.tags, key=lambda t: t.tag.tagged_date)
+latest_tag = tags[-1]
 
-version = cellxgene_census.__version__
+version = str(latest_tag)
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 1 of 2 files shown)


bkmartinjr (approved):
Does this assume that the git tag is the correct version for all docs?  I was under the impression @pablo-gar was still trying to decide if he wants one version or multiple versions.

Approving on the assumption that this decision has been taken, as code looks good.

ebezzi:
> Does this assume that the git tag is the correct version for all docs? I was under the impression @pablo-gar was still trying to decide if he wants one version or multiple versions.
> 
> Approving on the assumption that this decision has been taken, as code looks good.

The current docsite setup doesn't allow for multiple version. We'd need to switch back to readthedocs or use a sphinx plugin for that. This PR merely ensures that no "dev" tags will be added to the version number in the docsite when we build it from main.

bkmartinjr:
> The current docsite setup doesn't allow for multiple version

I meant different versions for Python & R packages, and perhaps even another for data docs.   But I take your point that this is a problem for another day!

atolopko-czi (approved):
LGTM. 

Is there an associated issue to link to?

pablo-gar:
> I meant different versions for Python & R packages, and perhaps even another for data docs. But I take your point that this is a problem for another day!

Python and R will have mirrored versions. Data docs are currently not versioned (which we either need to do or make a good effort to create good changelogs)

PR #690, opened 2023-08-05
[r] move SingleCellExperiment from Imports to Suggests

mlin:
The dependency on SingleCellExperiment is tricky because it's sourced from bioconductor, not CRAN nor r-universe. The fresh install instructions fail unless additional steps are taken to install it from bioconductor. 

Move it from our DESCRIPTION's Imports to Suggests, which [follows tiledbsoma](https://github.com/single-cell-data/TileDB-SOMA/blob/a103172aeabb57ae0865d52e6107326e4e5b4878/apis/r/DESCRIPTION#L69).

Right now just seeing if we still pass CI.

### DIFF
--- a/api/r/cellxgene.census/DESCRIPTION
+++ b/api/r/cellxgene.census/DESCRIPTION
@@ -22,7 +22,6 @@ Imports:
   jsonlite,
   methods,
   Seurat (>= 4.1.0),
-  SingleCellExperiment (>= 1.20.0),
   stats,
   tiledbsoma,
   tiledb
@@ -30,6 +29,7 @@ Suggests:
   bit64,
   knitr,
   rmarkdown,
+  SingleCellExperiment (>= 1.20.0),
   testthat (>= 3.0.0),
   withr
 Config/testthat/edition: 3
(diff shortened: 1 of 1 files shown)


mlin:
In tiledbsoma, `SOMAExperimentAxisQuery$to_single_cell_experiment()` has a runtime check for whether SingleCellExperiment is installed: https://github.com/single-cell-data/TileDB-SOMA/blob/a380da222a8b20d48258f2686dd9d058462fd176/apis/r/R/SOMAExperimentAxisQuery.R#L888

albeit with a barebones error message if not, `Error: 'SingleCellExperiment' must be installed`.

So the hope is this change will allow the package to install & load, and only hit a problem if user actually tries to use `get_single_cell_experiment()` without having installed SingleCellExperiment. To test completely, we'd need to cut another RC with this diff so that it goes to r-universe.

mlin:
@pablo-gar Please confirm whether to move Seurat as well. Personally I don't feel it necessary since Seurat installs from CRAN, but I'm not against it either.

mlin:
Superseded in #693

Issue #698, opened 2023-06-14
DistributedDataParallel Support


atolopko-czi:
- [x] ExperimentDataPipe should load only its own shard of data, based upon its worker index. With PyTorch Lightning, this is determined by the NODE_RANK and LOCAL_RANK env vars ([docs](https://pytorch-lightning.readthedocs.io/en/1.4.9/advanced/multi_gpu.html#distributed-data-parallel))
- [ ] ~Add notebook for single-host, multi-GPU example for using `DistributedDataParallel` with `ExperimentDataPipe`.~

atolopko-czi:
On a `g4dn.12xlarge` (4 T4 GPU, 192GB RAM, 48 VCPU), with locally-stored Census on EBS volume (gp3, 3000 IOPS), a single epoch of SCVI training ran in **2h 42m** for all is_primary human data across all 60K genes. 

Of note, max RAM usage was 50GB (~25%). 

Total SOMA data retrieval sizes and times:
```
Retrieved SOMA batch totals: n_soma_batches=32, n_obs=8341060, nnz=9047894500, elapsed=0:07:33
Retrieved SOMA batch totals: n_soma_batches=32, n_obs=8341060, nnz=11406725005, elapsed=0:09:19
Retrieved SOMA batch totals: n_soma_batches=32, n_obs=8341061, nnz=25140416440, elapsed=0:16:28
Retrieved SOMA batch totals: n_soma_batches=32, n_obs=8341061, nnz=19291365281, elapsed=0:14:04
```

Command line and gnu time output:
```
python census_scvi.py --obs-value-filter is_primary_data == True --torch-devices 4 --soma-buffer-bytes 2097152 --census-uri /mnt/census"
	User time (seconds): 30730.48
	System time (seconds): 873.48
	Percent of CPU this job got: 324%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 2:42:33
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 52908476
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 18589
	Minor (reclaiming a frame) page faults: 175233939
	Voluntary context switches: 42812418
	Involuntary context switches: 1238673
	Swaps: 0
	File system inputs: 68215288
	File system outputs: 24
	Socket messages sent: 0
	Socket messages received: 0
[...]

PR #699, opened 2023-08-07
[python] add support for distributed PyTorch

atolopko-czi:
Partially addresses #698.

Reduces memory consumption when used with [PyTorch DDP](https://pytorch.org/tutorials/intermediate/ddp_tutorial.html):
* `ExperimentDataPipe` partitions data by distributed training process and loads only the partition of data assigned to that process. Previously, each process loads the full query result, retrieving`(N-1)/N`of the result unnecessarily. Uses DDP's "rank" and "world size" values to determine the partitions.
* Use float32 for X tensor dtype. Previously, erroneously used float64.


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -21,14 +22,17 @@
...
 from torch.utils.data.dataset import Dataset
 
 from ..._open import _build_soma_tiledb_context
 from ..util._eager_iter import _EagerIterator
 
 pytorch_logger = logging.getLogger("cellxgene_census.experimental.pytorch")
-
+if not pytorch_logger.hasHandlers():
+    pytorch_logger.addHandler(logging.StreamHandler(sys.stderr))
+    pytorch_logger.setLevel(logging.DEBUG)
```
bkmartinjr, line 35: Note that the [logging docs](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) advise against this:
> Note It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers.

bkmartinjr (approved):
one comment inline (re: logging), but otherwise LGTM

Issue #587, opened 2023-06-27
Modify the R API to support mirrors and relative URIs


ebezzi:
See #586, but with R. Since the two APIs are very similar, this should roughly involve the same steps.

PR #719, opened 2023-08-16
[r] support mirrors & relative URI

mlin:
#587

cribbing from #552


api/r/cellxgene.census/R/open.R
```diff
@@ -20,8 +21,28 @@
...
+    uri <- locator$uri
+    if (is.null(tiledbsoma_ctx)) {
+      tiledbsoma_ctx <- new_SOMATileDBContext_for_census(NULL, mirror = locator$mirror_info)
+    }
+  }
+
+  return(tiledbsoma::SOMACollectionOpen(uri, tiledbsoma_ctx = tiledbsoma_ctx))
+}
+
+SUPPORTED_PROVIDERS <- c("S3", "file")
```
ebezzi, line 40: We decided to add `unknown` for when the user opens the census with `uri=...`. This should be treated similarly to `file` (use a default context without adding S3 params like region).

ebezzi (approved):
LGTM, thanks Mike!

api/r/cellxgene.census/R/release_directory.R
```diff
@@ -58,6 +47,36 @@ get_census_version_directory <- function() {
...
 
+#' Get the directory of Census mirrors currently available
+#'
+#' @return Nested list with information about available mirrors
+#' @importFrom jsonlite fromJSON
+#' @export
+#'
+#' @examples
+#' get_census_mirrors()
+get_census_mirrors <- function() {
```
ebezzi, line 58: @mlin I believe this needs to be exported to make it accessible.

ebezzi:
@mlin I pushed those two changes:
1. `get_census_mirrors` -> `get_census_mirror_directory` (consistent with Python's API)
2. Exported it so it's a public function.

PR #728, opened 2023-08-22
get_census_version_directory returns sorted elements

### LINKED ISSUE #439: Order release directory result
## Description

The `get_census_version_directory()` method returns a dictionary of census versions, but it is not ordered. To be friendly to humans, this dict should be an ordered dict, which aliases first, followed by release versions in descending order.

## Context

A very technically savvy user read the output of get_census_version_directory() and missed that a past build still existed, since it appeared lower in the list.

## Impact

Confusion and frustration when a user is unable to find the census version tag they believe should still exist.

## Alternatives you've considered

- Document the non-deterministic ordering of the result.
- New method that just returns the non-aliased release versions as an ordered list.

## Ideal behavior 

This returned dict should be an ordered dict, with aliases first shown first, followed by release versions in descending order.

ebezzi:
Resolves #439

### DIFF
--- a/api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
+++ b/api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
@@ -6,6 +6,7 @@
 
 Methods to retrieve information about versions of the publicly hosted Census object.
 """
+from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


atolopko-czi (approved):
LGTM

PR #731, opened 2023-08-24
Add exclude_zeros to pp.mean_variance

### LINKED ISSUE #635: `cellxgene_census.experimental.pp.mean_variance` should let the user choose exclusion/inclusion of zeroes for its calculations
**Product requirements**

A user can specify whether or not to use zeroes in the calculation of mean and variance. Currently zeroes are included.

An argument `exclude_zeroes`

`cellxgene_census.experimental.pp.mean_variance`:

```
def mean_variance(
   *,
   exclude_zeroes = False,
) -> pandas.DataFrame:

    """
    [...]

    Args:

        exclude_zeroes:
        Whether to exclude zeroes in the mean and variance calculations. If `True` the 
        total number of elements N is the the number of non-zero elements, otherwise it is the corresponding size of the dimension.
   
    [...]
    """
```

ebezzi:
Resolves #635


api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -74,12 +74,20 @@ def __init__(self, n_samples: int, n_variables: int):
 
         self.u = np.zeros(n_variables, dtype=np.float64)
         self.n_samples = n_samples
+        # If we want to exclude zeros, we need to keep track of the denominator
+        self.exclude_zeros = exclude_zeros
+        self.denominator = np.zeros(n_variables)
 
     def update(self, var_vec: npt.NDArray[np.int64], val_vec: npt.NDArray[np.float32]) -> None:
```
bkmartinjr, line 81: This approach requires two passes through the `var_vec` if `exclude_zeros is True`.  I am almost certain a faster approach would be to have two separate `_update_...` functions, one which updates only the `u` accumulator, and the other which updates _both_ `u` and `demoninator`.

In other words, this `update` method would look like:
```
def update(self, var_vec: npt.NDArray[np.int64], val_vec: npt.NDArray[np.float32]) -> None:
    if self.exclude_zeros:
        _update_mean_and_denominator_vectors(self.u, self.denominator, var_vec, val_vec)
    else:
        _update_mean_vector(self.u, var_vec, val_vec)
```

The implementation of `_update_mean_vector` is unchanged.  The `_update_mean_and_denominator_vectors` performs _both_ operations within a single loop, e.g.,
```
def _update_mean_and_denominator_vector(
    u: npt.NDArray[np.float64],
    denominator: npt.NDArray[np.float64],
    var_vec: npt.NDArray[np.int64],
    val_vec: npt.NDArray[np.float32],
) -> None:
    for col, val in zip(var_vec, val_vec):
        u[col] = u[col] + val
        denominator[col] += 1
```

bkmartinjr:
> Notes: implementing the same approach in MeanVarianceAccumulator would require significant work since the method supports batching and the division is implemented down in the numba methods. Since batching isn't required by the mean_variance method, we could just rewrite a MeanVarianceAccumulatorUnbatched that also supports exclude_zeros. Otherwise, this can be implemented even for batching but it will require some refactoring of the code and will take more time. Thoughts?

Actually, I think it is pretty easy, as the current code _already_ counts the number of values per variable (equivalent to what you call `denominator` in this PR -- called `n` in the existing code), _and_ it already corrects for the difference between the non-zero values and the total samples (see the code in `_mbomv_sparse_correct_batches`)

I believe that it may be as simple as not calling `_mbomv_sparse_correct_batches` in the new mode.  `_mbomv_sparse_correct_batches` is simply updating all the batch accumulators for the implicit zero-valued (i.e., sparse or missing) array elements.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -18,17 +18,28 @@ class MeanVarianceAccumulator:
...
 
+        self.exclude_zeros = exclude_zeros
         self.ddof = ddof
         self.n_batches = n_batches
         self.n_samples = n_samples
         self.n = np.zeros((n_batches, n_variables), dtype=np.int32)
         self.u = np.zeros((n_batches, n_variables), dtype=np.float64)
         self.M2 = np.zeros((n_batches, n_variables), dtype=np.float64)
 
+        if self.exclude_zeros and self.n_batches > 1:
```
ebezzi, line 40: Implementing `exclude_zeros` with batching requires some nontrivial changes to `_mbomv_combine_batches`, so since we're not using it, I decided to disable it. We can enable it later if necessary.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -58,14 +71,20 @@ def finalize(
...
         all_u, all_M2 = _mbomv_combine_batches(self.n_batches, self.n_samples, self.u, self.M2)
+
         with np.errstate(divide="ignore"):
-            all_var = all_M2 / max(0, (self.n_samples.sum() - self.ddof))
+            if self.exclude_zeros:
+                all_var = all_M2 / np.maximum(0, self.n - self.ddof)
+                all_var = all_var[0]
+            else:
+                all_var = all_M2 / max(0, (self.n_samples.sum() - self.ddof))
+                print(all_var, all_var.shape)
```
ebezzi, line 81:
```suggestion
```

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -19,6 +19,7 @@ def mean_variance(
     calculate_mean: bool = False,
     calculate_variance: bool = False,
     ddof: int = 1,
+    exclude_zeros: bool = False,
 ) -> pd.DataFrame:
     """
     Calculate  mean and/or variance along the ``obs`` axis from query results. Calculations
```
bkmartinjr, line 25: the docstrings lack documentation of the `exclude_zeros` argument

bkmartinjr (commented):
One pervasive confusion I have with this PR is if we are adding a flag to:
* ignore zeros by value
* or ignore nnz (i.e., ignore implicit zeros, but _include_ explicit zeros)

The latter is coded.  But the argument flag (`exclude_zeros`) is ambiguous.

SOMA defines `nnz` as explicitly written values, even if that value is zero. The Census never intentionally writes a zero, but given that this API operates on SOMA data, the flag should make it clear which is implemented, and make the code / arg names match.

My personal opinion is that the latter (ignore implicit zeros) is what you want, for both API consistency and performance reasons.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
bkmartinjr: Batching seems pretty fundamental to the use case - @pablo-gar what are your thoughts? 
 Without it, any use case that requires batching will have to do batching above this API, requiring multiple passes through the data.

I"m also unclear what complexity is implied by adding multi-batch support to the combine batches step.  I believe that the only change is to not adjust the value of `n` for the total sample count (in `n_samples`), which is a 1-2 line change....

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -58,14 +71,19 @@ def finalize(
...
         # accum all batches using Chan's
         all_u, all_M2 = _mbomv_combine_batches(self.n_batches, self.n_samples, self.u, self.M2)
+
         with np.errstate(divide="ignore"):
-            all_var = all_M2 / max(0, (self.n_samples.sum() - self.ddof))
+            if self.exclude_zeros:
+                all_var = all_M2 / np.maximum(0, self.n - self.ddof)
+                all_var = all_var[0]
+            else:
+                all_var = all_M2 / max(0, (self.n_samples.sum() - self.ddof))
```
bkmartinjr, line 80: i would be consistent on use of max vs np.maximum

I'm not sure which is right, but I assume that np.maximum was introduced for a reason?

api/python/cellxgene_census/tests/experimental/pp/test_online.py
```diff
@@ -37,7 +38,10 @@ def test_meanvar(matrix: sparse.coo_matrix, n_batches: int, stride: int, ddof: i
     assert n_samples.sum() == matrix.shape[0]
     assert len(n_samples) == n_batches
 
-    olmv = MeanVarianceAccumulator(n_batches, n_samples, matrix.shape[1], ddof)
+    # exclude_zeros only if there is a single batch
```
bkmartinjr, line 41: missing test to confirm that the code raises if `batches>1 and exclude_zeros`

api/python/cellxgene_census/tests/experimental/pp/test_online.py
```diff
@@ -54,14 +58,25 @@ def test_meanvar(matrix: sparse.coo_matrix, n_batches: int, stride: int, ddof: i
...
-    assert allclose(all_u, dense.mean(axis=0))
-    assert allclose(all_var, dense.var(axis=0, ddof=ddof, dtype=np.float64))
 
-    matrix = matrix.tocsr()  # COO not conveniently indexable
-    for batch in range(n_batches):
-        dense = matrix[batches == batch, :].toarray()
-        assert allclose(batches_u[batch], dense.mean(axis=0))
-        assert allclose(batches_var[batch], dense.var(axis=0, ddof=ddof, dtype=np.float64))
+    if should_exclude_zeros:
+        # Normalize dense so that 0s are NaNs - allows to use np.nanmean and np.nanvar
```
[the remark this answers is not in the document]
bkmartinjr: Also, if we end up restricting this to `batches==1`, then an even simpler approach is to just compute over matrix.data directly, which is a simple numpy array omitting any implicitly stored value.

bkmartinjr (changes requested):
several thoughts:
1. API needs clarification (and possibly code changes) to define if this "excludes zeros by value" or "excludes implicit zeros".  I am advocating for the latter as it is consistent with SOMA, but this needs clarification, docstrings and unit test changes to match (the unit tests are inconsistent in this regard).
2. Unit tests are using NaN to indicate values to ignore, but Nan is a legal float.  Should pick another approach (suggested several inline)
3. WRT point number 1 above, the name of the param (exclude_zeros) is ambiguous.  Minimally needs a good docstring, but even better, we might refine the name if someone has a good idea :-)
4. I believe implementing this feature for batched meanvar is simple - I suggest we just do it, and expose batching at the top-level.  Ultimately a call for @pablo-gar

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
ebezzi: Oops, forgot to add that. Thanks!

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
ebezzi: `np.maximum` does an element-wise maximum on the array here, while the latter is a scalar.

api/python/cellxgene_census/tests/experimental/pp/test_stats.py
```diff
@@ -88,6 +88,62 @@ def test_mean_variance(
...
+@pytest.mark.parametrize("calc_mean,calc_variance", [(True, True)])
+@pytest.mark.parametrize(
+    "experiment_name,obs_value_filter,obs_coords",
+    [
+        ("mus_musculus", 'tissue_general == "liver" and is_primary_data == True', ()),
+        pytest.param("mus_musculus", "is_primary_data == True", (slice(0, 400_000),), marks=pytest.mark.expensive),
+        pytest.param("homo_sapiens", "is_primary_data == True", (slice(0, 400_000),), marks=pytest.mark.expensive),
+    ],
+)
+def test_mean_variance_exclude_zeros(
```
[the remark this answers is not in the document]
ebezzi: The axis=1 was an omission, thanks for catching that! I also agree it's probably better to merge the tests into one, kinda like it did in `test_online`. It will require some additional branching but it will also make it faster, since it won't have to re-query the census twice. Will go for it.

ebezzi:
> One pervasive confusion I have with this PR is if we are adding a flag to:
> 
> * ignore zeros by value
> * or ignore nnz (i.e., ignore implicit zeros, but _include_ explicit zeros)
> 
> The latter is coded. But the argument flag (`exclude_zeros`) is ambiguous.
> 
> SOMA defines `nnz` as explicitly written values, even if that value is zero. The Census never intentionally writes a zero, but given that this API operates on SOMA data, the flag should make it clear which is implemented, and make the code / arg names match.
> 
> My personal opinion is that the latter (ignore implicit zeros) is what you want, for both API consistency and performance reasons.

spoke with @pablo-gar and we can go with ignoring nnz (which is what the code currently implements), and we'll add a note to the docstrings saying that the census doesn't add explicit zeros. This should hopefully clarify doubts that users may have with respect to this issue

bkmartinjr:
> go with ignoring nnz

great. The additional change required, besides docstrings, is unit tests.  They are a bit muddled on which they test for - there should be an explicit test with and without explicitly stored zeros.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
ebezzi: Will implement as part of https://github.com/chanzuckerberg/cellxgene-census/issues/786

api/python/cellxgene_census/tests/experimental/pp/test_online.py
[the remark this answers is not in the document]
ebezzi: I changed the test to use masked arrays.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
pablo-gar: Some extra context. Batching has a bit of a different connotation here as compared to highly variable genes, in highly variable genes average/variance are calculated per batch and the results are aggregated to retrieve _batch-corrected_ variable genes. In this case the functionality that we want to enable (which I believe would be enabled by this option) is a "group_by" functionality to save user the hassle to call this function in a for loop.

Specifically, as an example a user may want perform an axis query to obtain all cells from a tissue and _then_ perform average gene expression for each cell types a.k.a "group by".

Agreed to add this in a different PR

pablo-gar: I don't see a use case for "batch-corrected" mean and variance calculations.

pablo-gar (approved):
Looks good to me. I did a sanity check with the pre-calculated stats that @bkmartinjr has done already. It's working great!

```python
import cellxgene_census
from cellxgene_census.experimental.pp import mean_variance
import tiledbsoma as soma
import pandas as pd

    with census["census_data"][experiment_name].axis_query(
        measurement_name="RNA", obs_query=soma.AxisQuery(coords=(range(20),))
    ) as query:
        mv_df = mean_variance(
            query,
            axis=1,
            calculate_mean=True,
            calculate_variance=True,
            exclude_zeros=True,
        )

        obs_df = query.obs(column_names = ["soma_joinid", "raw_mean_nnz", "raw_variance_nnz"]).concat().to_pandas()
        
        combined_df = pd.concat([obs_df.set_index("soma_joinid"), mv_df], axis=1)

combined_df
```

image

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
bkmartinjr: We are using the 'batch' term in too many ways!  The thing we are advocating for has nothing to do with batch correction.

What the current underlying implementation provides is a way to perform separate mean/var calculations, on subsets of the data identified by a batch key, all in one pass.  This is a performance win, and very important for algorithms that need the stats per-batch in order to do other work (e.g.,. HVG uses this)

If we don't add this, then a user that needs per-batch stats will have to do multiple passes over the data (perf hit)

api/python/cellxgene_census/tests/experimental/pp/test_online.py
ebezzi: Added as a separate test.

api/python/cellxgene_census/tests/experimental/pp/test_stats.py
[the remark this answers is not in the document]
ebezzi: Update: since this test requires materializing the matrix in memory to calculate mean/var without the zeros, it also requires way more memory. Therefore, I had to keep it separate and change the query strategy so that the matrix doesn't get too big. I also added back both axis.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -82,15 +87,18 @@ def iterate() -> Generator[Tuple[npt.NDArray[np.int64], Any], None, None]:
     )
 
     if calculate_variance:
-        mvn = MeanVarianceAccumulator(n_batches, n_samples, n_dim_0, ddof=ddof)
+        print("starting calc")
```
ebezzi, line 90:
```suggestion
```

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -46,6 +47,10 @@ def mean_variance(
         ddof:
             "Delta Degrees of Freedom": the divisor used in the calculation for variance is ``N - ddof``, where ``N`` represents the number of elements.
 
+        exclude_zeros:
```
bkmartinjr, line 50: In addition, with the decision to define `N` as `nnz` when this flag is True, I think we should rename it.  It is no longer `exclude zeros`, but rather `use nnz as N`.

My suggestion is to rename for clarity - possible candidates:
* nnz_only
* use_nnz_as_sample_count

etc.  @pablo-gar - do you have a good name that would be clear and descriptive?  `exclude_zeros` seems misleading.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
bkmartinjr: remove debugging code

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -82,15 +87,18 @@ def iterate() -> Generator[Tuple[npt.NDArray[np.int64], Any], None, None]:
     )
 
     if calculate_variance:
-        mvn = MeanVarianceAccumulator(n_batches, n_samples, n_dim_0, ddof=ddof)
+        print("starting calc")
+        mvn = MeanVarianceAccumulator(n_batches, n_samples, n_dim_0, ddof=ddof, exclude_zeros=exclude_zeros)
         for dim, data in iterate():
             mvn.update(dim, data)
+            print("update")
```
bkmartinjr: more debugging code

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
ebezzi: I agree. Let's wait for Pablo to chime in.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -82,15 +87,18 @@ def iterate() -> Generator[Tuple[npt.NDArray[np.int64], Any], None, None]:
...
         _, _, all_u, all_var = mvn.finalize()
+        print("finalized")
```
bkmartinjr: debugging code

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
ebezzi: Ah ok, good point. We can go for `np.maximum` then. I'll make the change, thanks!

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -74,12 +92,21 @@ def __init__(self, n_samples: int, n_variables: int):
 
         self.u = np.zeros(n_variables, dtype=np.float64)
         self.n_samples = n_samples
+        # If we want to exclude zeros, we need to keep track of the denominator
+        self.exclude_zeros = exclude_zeros
+        self.denominator = np.zeros(n_variables)
```
bkmartinjr, line 97: You have introduced a second naming scheme for the sample count (`n`), calling it `denominator`.  The existing code uses `n` which is the convention in stats.  See the code in `MeanVarianceAccumulator`.

Recommend being consistent, as it will make the code easier to understand.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -364,3 +391,32 @@ def _update_mean_vector(
...
+            numba.types.Array(numba.int64, 1, "C", readonly=True),
+            numba.types.Array(numba.float32, 1, "C", readonly=True),
+        ),
+    ],
+    nopython=True,
+    nogil=True,
+)  # type: ignore[misc]  # See https://github.com/numba/numba/issues/7424
+def _update_mean_and_denominator_vectors(
+    u: npt.NDArray[np.float64],
+    d: npt.NDArray[np.float64],
```
bkmartinjr, line 416: for consistency, this should be `n` -- see `_mbomv_update_single_batch`

api/python/cellxgene_census/tests/experimental/pp/test_online.py
```diff
@@ -54,14 +59,38 @@ def test_meanvar(matrix: sparse.coo_matrix, n_batches: int, stride: int, ddof: i
...
-    for batch in range(n_batches):
-        dense = matrix[batches == batch, :].toarray()
-        assert allclose(batches_u[batch], dense.mean(axis=0))
-        assert allclose(batches_var[batch], dense.var(axis=0, ddof=ddof, dtype=np.float64))
+    if should_exclude_zeros:
+        mask = np.ones(matrix.shape)
+        r, c = matrix.tolil().nonzero()
+        for x, y in zip(r, c):
+            mask[x, y] = 0
+        masked = ma.masked_array(dense, mask=mask)  # type: ignore[no-untyped-call, var-annotated]
```
bkmartinjr, line 68: Please don't disable mypy errors when they can just be fixed with a type declaration.

E.g., you can remove the `var-annotated` with a type declaration:
>         masked: np.ma.MaskedArray[np.bool_, np.dtype[np.float64]] = ma.masked_array(dense, mask=mask)  # type: ignore[no-untyped-call]

bkmartinjr (changes requested):
A bunch of super easy to fix nits, plus:
* fix docstrings so they are accurate
* we need a better name for the `exclude_zeros` argument, as it is not descriptive of the actual behavior.  I suggest `nnz_only` or something similar.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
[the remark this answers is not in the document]
ebezzi: In order to not over-complicate the main paragraph, I am proposing that we add `by default` and specify the default for `exclude_zeros` is False. This way the description should be consistent.

ebezzi: For the time being, I settled for `nnz_only`.

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
```diff
@@ -19,14 +19,15 @@ def mean_variance(
...
+    nnz_only: bool = False,
 ) -> pd.DataFrame:
     """
     Calculate  mean and/or variance along the ``obs`` axis from query results. Calculations
     are done in an accumulative chunked fashion. For the mean and variance calculations,
-    the total number of elements (N) is the corresponding dimension size:
+    the total number of elements (N) is, by default, the corresponding dimension size:
     for column-wise calculations (``axis = 0``) N is number of rows,
     for row-wise calculations (``axis = 1``) N is number of columns.
+    For metrics calculated only on nnz values of the sparse matrix, specify ``nnz_only=True``.
```
bkmartinjr, line 30: nit:  suggest "...only on nnz (explicitly stored) values of the ..."

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_stats.py
bkmartinjr: As long as @pablo-gar is good with this change, I'm good....

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
```diff
@@ -74,12 +92,21 @@ def __init__(self, n_samples: int, n_variables: int):
...
         self.u = np.zeros(n_variables, dtype=np.float64)
         self.n_samples = n_samples
+        # If we want to exclude zeros, we need to keep track of the denominator
+        self.nnz_only = nnz_only
+        self.n = np.zeros(n_variables)
 
     def update(self, var_vec: npt.NDArray[np.int64], val_vec: npt.NDArray[np.float32]) -> None:
-        _update_mean_vector(self.u, var_vec, val_vec)
+        if self.nnz_only:
+            _update_mean_and_denominator_vectors(self.u, self.n, var_vec, val_vec)
```
bkmartinjr, line 101: I'm still finding the use of `denominator` inconsistent, but I'll stop bugging you about it :-)

bkmartinjr (approved):
Approved.  There are a couple of additional "editorial" changes, but none that hold up approval!

api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_online.py
ebezzi: I can rename it to `_update_mean_and_n_vectors`.

Issue #457, opened 2023-05-03
Finalize conventions for recording LTS data releases


pablo-gar:
## Motivation

We currently don't have a final system to tag and surface LTS data release in the `release.json` and the API respectively.

**Definitions**

Please see definition of data releases [here](https://github.com/chanzuckerberg/cellxgene-census/blob/main/docs/cellxgene_census_docsite_data_release_info.md#long-term-supported-lts-cell-census-releases).

Labels used by API in `open_soma(census_version = "")`:

- "stable" is the latest LTS data release
- "latest" is the current weekly release

**User stories**

1. I want to access the latest Census data release that contains datasets from as late as last week. 
2. I want to access the latest LTS Census data release. And I would want to know which specific version it is so that in the future I can get use this exact version to get consistent results.
3. I want to access a specific LTS Census data release for consistency.
4. I want to understand if, why, and which LTS Census releases have been removed. 

**Product requirements**

This system must:

- Be backwards compatible with current implementation
- Allow a future version of the API to identify all data releases available to `open_soma`.
- Allow a future version of the API to identify all LTS data releases available to `open_soma`
- The LTS release names surfaced to the user must be clear and indicate they are LTS, as well as their build dates.
- `open_soma` should provide clear messaging of what version is being opened (**already satisfied**).
- The default `census_version = "stable"` of `open_soma` must open the latest LTS data release.
- `census_version = "latest"` of `open_soma` must open the current weekly release.
- In the unlikely edge-case where an LTS is deleted:
   - The system must retire the data release and flag it as such.
   - They system must be able to record the reason for its removal.
   - The (aforementioned future) API must provide access to this reason/metadata
- The APIs must:
    - Allow to get the aliases of all LTS's
    - Provide clear information of the release being opened upon calling `open_soma` (already implemented)   - The API must provide an error with clear messaging if a user tries to open it via `open_soma`

## Proposed implementation

The following serves as a _guide_ for eng but final design is up to engineers

The `release.json` has the following conventions:

- A dynamic alias `"stable"` that points to the latest LTS data release (already implemented)
- A dynamic alias `"latest"` that points to the current weekly release (already implemented)
- Each LTS data release must be named with the convention `"LTS-[YY.MM]"` (new)
- Each release must have a key `is_LTS` which must be True if the release is an LTS, False otherwise (new)

Latest example `release.json` as of 2023-08-23:

```json
{
  "stable": "2023-07-25",
  "latest": "2023-08-22",
[...]

PR #737, opened 2023-08-28
[builder][python api] Explicit LTS releases in releases.json

atolopko-czi:
For #457. 

Using this PR as means to review and potentially ratify the design for explicit LTS release support in `releases.json`. If the design is acceptable to all, we will need equivalent R API changes to complete #457.  

- Add support for an optional "flags" dictionary in releases.json. Currently, only allows for "lts" and "retracted" flags to specified.
- Add support for an optional "retraction" dictionary in releases.json
- ~Add `get_census_version_directory_of_lts_releases()` convenience method, which only returns LTS releases and excludes retracted releases by default, but can optionally included them.~
- Remove `alias` key/value pair from result of `get_census_version_directory()`; this was an internal attribute that should not have been exposed before and is no longer used in `_open_soma()`. This is a breaking API change, technically, but I don't think it requires a major version bump (all of our code is still experimental!)
- On `open_soma()`, handle "lts"- and "retracted"-flagged releases appropriately via updated log messaging
- Removed temporary logic for handling non-existence of a "stable" alias in releases.json; this was only needed to handle initial case where API was released before a "stable" release existed.
- Updated `docs/cellxgene_census_storage_and_release_policy.md` with policy for using "lts" and "retracted" flags


tools/cellxgene_census_builder/src/cellxgene_census_builder/release_manifest.py
```diff
@@ -25,11 +25,13 @@
 CensusVersionDescription = TypedDict(
     "CensusVersionDescription",
     {
-        "release_date": Optional[str],  # date of release, optional
+        "release_date": Optional[str],  # date of release (deprecated)
         "release_build": str,  # date of build
         "soma": CensusLocator,  # SOMA objects locator
         "h5ads": CensusLocator,  # source H5ADs locator
         "do_not_delete": Optional[bool],  # if set, prevents automated deletion
+        "flags": Optional[Dict[str, bool]],  # flags for the release
```
atolopko-czi: since Builder is not manipulating the contents of "flags" (for now), I left this more loosely typed than what we have on the API side, which defines `ReleaseFlags` type, et. al.

docs/cellxgene_census_storage_and_release_policy.md
```diff
@@ -42,7 +42,16 @@ The publication date along with the full URI paths for the `soma` folder and the
 
 * This file MUST be in `json` formats where the parent keys are release identifiers (alias or name).
 * The alias `latest` MUST be present and MUST point to the **Weekly Census release**.
-* The prefix `V` MUST be used followed by an integer counter to label long-term supported Census releases, e.g. `V1`.
+* For a long-term supported ("LTS") release:
+  * A `flags` key MUST be present with a dictionary-typed value, which MUST contain the key `lts` with the value `true`.
+  * An alias MUST exist that points to the release, formatted as `V<#>`, where `<#>` is integer  to label long-term supported Census releases, e.g. `V1`.
```
atolopko-czi, line 47: Since we already had the "V<#>" LTS naming policy in place, I stuck with it. Note that we discussed using "LTS-YYYY-MM-DD" in our design meeting. I don't feel strongly either way,

api/python/cellxgene_census/tests/test_open.py
```diff
@@ -190,35 +190,6 @@ def test_open_soma_works_if_no_relative_uri_specified(requests_mock: rm.Mocker)
         )
 
 
-def test_open_soma_defaults_to_latest_if_missing_stable(requests_mock: rm.Mocker) -> None:
```
atolopko-czi, line 193: no longer relevant, since this was testing the temporary case where we did not yet have a "stable" release in releases.json

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -158,38 +200,64 @@ def get_census_version_directory() -> Dict[CensusVersionName, CensusVersionDescr
...
     """
+    if invalid_release_flags := release_flags_kwargs.keys() - RELEASE_FLAGS:
+        raise ValueError(f"Unknown release flag(s): {invalid_release_flags}")
+
     response = requests.get(CELL_CENSUS_RELEASE_DIRECTORY_URL)
     response.raise_for_status()
 
     directory: CensusDirectory = cast(CensusDirectory, response.json())
+    directory_out: CensusDirectory = {}
+    aliases: Dict[CensusVersionName, List[CensusVersionName]] = {}
```
atolopko-czi, line 211: much of the below code changes are for handling alias mappings internally in the method, to avoid adding a synthentic "alias" key/value to each CensusVersionDescription.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -158,38 +200,64 @@ def get_census_version_directory() -> Dict[CensusVersionName, CensusVersionDescr
...
+    unordered_directory = cast(Dict[CensusVersionName, CensusVersionDescription], directory_out)
 
     # Sort by aliases and release date, descending
-
-    aliases = [(k, v) for k, v in unordered_directory.items() if v.get("alias") is not None]
-    releases = [(k, v) for k, v in unordered_directory.items() if v.get("alias") is None]
+    aliased_releases = [(k, v) for k, v in unordered_directory.items() if k in aliases]
+    concrete_releases = [(k, v) for k, v in unordered_directory.items() if k not in aliases]
     ordered_directory = OrderedDict()
-    for k, v in aliases + sorted(releases, key=lambda k: k[0], reverse=True):
```
atolopko-czi, line 192: aliases were themselves sorted before; now they are

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -124,18 +137,47 @@ def get_census_version_description(census_version: str) -> CensusVersionDescript
     return description
 
 
-def get_census_version_directory() -> Dict[CensusVersionName, CensusVersionDescription]:
+def get_census_version_directory_of_lts_releases(
```
bkmartinjr, line 140: What is the rationale for adding this wrapper?  It seems primarily to substitute for docs/examples of how to use `get_census_version_directory`.  It doesn't save the user any keystrokes or required level of understanding....

My suggestion is to remove it, and just improve the an Examples section of `get_census_version_directory`, highlighting suggested use of the `lts` param, etc.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -124,18 +137,47 @@ def get_census_version_description(census_version: str) -> CensusVersionDescript
...
     See Also:
+        :func:`get_census_version_directory`: get directory of all Census releases currently available.
         :func:`get_census_version_description`: get description by census_version.
+    """
+    release_flags = dict(lts=True) if include_retracted else dict(lts=True, retracted=False)
+    return get_census_version_directory(**release_flags)
+
+
+# TODO: Update example output and add example for use of release_flags
+def get_census_version_directory(**release_flags_kwargs: bool) -> Dict[CensusVersionName, CensusVersionDescription]:
```
bkmartinjr, line 165: need docstring info on flags - or better yet, a pointer to where they are documented.

also need additional examples highlighting most common use cases such as lts=True

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -184,25 +184,24 @@ def open_soma(
...
-            ) from None
-
-    if description["alias"]:
+        raise ValueError(
+            f'The "{census_version}" Census version is not valid. Use get_census_version_directory() to retrieve '
+            f"available versions."
+        ) from None
+
+    if description.get("flags", {}).get("retracted", False):
+        api_logger.info(
```
bkmartinjr, line 198: IMHO, this should be a `warning` level log call

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
bkmartinjr: Can you comment on the motivation for this change?  I'm not opposed, but also don't see any functional benefit (I'm probably missing something obvious!)

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -25,14 +25,27 @@
...
+    "CensusVersionRetraction",
+    {
+        "date": str,  # the date of retraction
+        "reason": Optional[str],  # the reason for retraction
+        "info_url": Optional[str],  # a permalink to more information
+        "replaced_by": Optional[str],  # the census version that replaces this one
+    },
+)
+RELEASE_FLAGS = {"lts", "retracted"}
+ReleaseFlag = Literal["lts", "retracted"]
```
bkmartinjr, line 38: this seems brittle/not-DRY, as the code later relies on the type definition always matching the set definition.

Suggest:
```
ReleaseFlag = Literal["lts", "retracted"]
RELEASE_FLAGS = { *typing.get_args(ReleaseFlag) }
```

api/python/cellxgene_census/src/cellxgene_census/_open.py
atolopko-czi: agreed, changed!

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
atolopko-czi: ah yes, was hoping to DRY this up, done!

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
atolopko-czi: This method is intended to be a "policy-aware" convenience method, built upon the "mechanism"-only `get_census_version_directory()` method. I thought this is what we had discussed in the design meeting, but perhaps I misunderstood. 

However, I do think API users read method names (and use them) before they carefully read docs, so it could be worth keeping from a UX perspective.

I'm not wedded to it, but let's hear from @pablo-gar and @mlin first.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
atolopko-czi: Since this is intended to be a "mechanism"-only method, I opted to have keep the docstring agnostic about the admissible flags. If we remove, `get_census_version_directory_of_lts_releases()`, then it definitely makes sense to document the flags here.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
atolopko-czi: This is _only_ needed later in the method to order the output, where it needs to know which releases are identified by aliases. Realizing now that this can be simplified, so I've now done that.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
bkmartinjr: the (large) downside IMHO is the loss of typing information (both IDE hints, linting, etc)

Explicit signatures are much preferred IMHO, and in real-world use no less flexible.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
bkmartinjr: When I initially made the "separate policy from mechanism" comment, I did not intend for it to imply additional functions, but rather additional operations.  Either approach satisfies that goal.  The question of a separate function name vs. flags is a style question only, and I think we should do whatever is easiest for users to understand/use.

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
atolopko-czi: As that's what I proposed in the original design changes, I'm not opposed to it. But I thought the outcome of the discussion was to remain agnostic about the specific flags in this method, and in the releases.json file. But maybe just the latter was the case?

bkmartinjr (changes requested):
Several open issues noted, only one of which I think must be resolved:  the untyped parameters for `get_census_version_directory` remove very useful/mportant typing information, and I don't think we should have any public API this loosely defined. Given all the typing elsewhere, lets define the arg types explicitly

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
bkmartinjr: I agree we wanted flexible flags over the long-term in the data model (release.json). But in any given API release, we are not agnostic and we code for a specific schema.  So the untyped signature doesn't really do much other than disable tooling help

atolopko-czi: Now typed!

bkmartinjr:
Seems to be failing lint:
> tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py:224: error: Missing keys ("flags", "retraction") for TypedDict "CensusVersionDescription"  [typeddict-item]

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -137,38 +137,18 @@ def get_census_version_description(census_version: str) -> CensusVersionDescript
...
-        :func:`get_census_version_description`: get description by census_version.
-    """
-    release_flags = dict(lts=True) if include_retracted else dict(lts=True, retracted=False)
-    return get_census_version_directory(**release_flags)
-
-
 # TODO: Update example output and add example for use of release_flags
-def get_census_version_directory(**release_flags_kwargs: bool) -> Dict[CensusVersionName, CensusVersionDescription]:
+def get_census_version_directory(
+    lts: Optional[bool] = None, retracted: Optional[bool] = None
```
bkmartinjr, line 142: I am wondering if an unspecified `retracted` flag should not default to `True`

Pros/cons:
* we really don't want people using retracted datasets, so defaulting `True` accomplishes it
* if somewhat _is_ using a dataset, which is retracted after they being using it, the API may start to fail in a surprising manner.  Arguably a good thing, as we _want_ them to notice this change in status.

Is there a use case or motivation for a default to False?

bkmartinjr (commented):
One additional small item related to the default for the `retracted` filter flag.  My gut is that it should default to `True`, rather than the current default of `False`

@pablo-gar - thoughts?

atolopko-czi:
> Seems to be failing lint:
> 
> > tools/cellxgene_census_builder/src/cellxgene_census_builder/**main**.py:224: error: Missing keys ("flags", "retraction") for TypedDict "CensusVersionDescription"  [typeddict-item]

Yeah, have been looking into it. I want to use `typing.NotRequired` for the `flags` and `retraction` entries, but that requires >=3.11.

bkmartinjr:
> typing.NotRequired

try `typing_extensions`

atolopko-czi:
> One additional small item related to the default for the `retracted` filter flag. My gut is that it should default to `True`, rather than the current default of `False`
> 
> @pablo-gar - thoughts?

Woops, unintentionally lost that inverted default value with the previous changes! Reinstated. @pablo-gar can still weigh in, of course. :)

api/python/cellxgene_census/src/cellxgene_census/_release_directory.py
```diff
@@ -139,7 +139,7 @@ def get_census_version_description(census_version: str) -> CensusVersionDescript
 
 # TODO: Update example output and add example for use of release_flags
 def get_census_version_directory(
-    lts: Optional[bool] = None, retracted: Optional[bool] = None
+    lts: Optional[bool] = None, retracted: Optional[bool] = False
```
bkmartinjr, line 142: not sure the docstring matches the default for `retracted`.

Also, do you really want the flags to accept `None`? Seems better to just force as keyword args, and explicitly default them, i.e.
>   def get_census_version_directory(*, lts: bool = False, retracted: bool = True)

atolopko-czi: It's tricky because (I think) we want to support filtering releases in 3 ways:  flag is set (`flag=True`), flag is explicitly not set (`flag=False`), flag is ignored and can have either way (`flag=None`).

bkmartinjr: ah, I did not understand the "do not filter" condition - I took "false" to mean "don't filter", which is clearly not what we want!

Given case,I agree with your ternary approach, but I would add the "no positionals" marker.  I also that if `lts` really defaults to true (rather than "ignore"), that the signature reflect this.

atolopko-czi: added keyword-only marker to method signature and added full docs & examples. PTAL

> Also, if lts really defaults to true (rather than "ignore"), the signature should reflect this.

`lts` defaults to `None`/ignore, so I think we're good as is?

bkmartinjr (approved):
LGTM

mlin (approved):
LGTM. I just filed #766 to sync the R analogs. @pablo-gar - high priority?

atolopko-czi:
Dear Github Actions Bot, 

I reminded team to review it again, just yesterday! As this PR is still active, please do not close. Thank you for your keeping our PR list up-to-date.

Sincerely,
A Concerned Engineer

docs/cellxgene_census_storage_and_release_policy.md
pablo-gar: Same here!

pablo-gar (approved):
LGTM

Issue #744, opened 2023-08-31
[builder] Failed at consolidation step


atolopko-czi:
Failed at consolidation step with `tiledb.cc.TileDBError: [TileDB::Reader] Error: Incorrect unfiltered tile size allocated.` [[logs](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/%2Fcellxgene-census-builder-prod/log-events/cellxgene-census-builder-prod%2Fdefault%2F2bfcf805029e4a19824294d0419d86bf)]

12 of 13 consolidation processed completed successfully.
The failed consolidation was for: `file:///census-build/2023-08-28/soma/census_data/homo_sapiens/ms/RNA/X/normalized`

atolopko-czi:
Fix is in place. Will wait until Monday's scheduled run to confirm success before closing out this issue.

atolopko-czi:
2023-09-04 build was successful, so closing this issue.

atolopko-czi:
2023-09-12 build failed again with same error ([logs](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/%2Fcellxgene-census-builder-prod/log-events/cellxgene-census-builder-prod%2Fdefault%2F63a21f7fdf2e46faaaf8e3915740aecd)). So not likely caused by disk space.

atolopko-czi:
- [ ] Share the following with TileDB at https://czi-sci.slack.com/archives/C013K0JB8F8/p1694537135163599?thread_ts=1694533729.655039&cid=C013K0JB8F8
* tiledbsoma.show_package_versions() as used by the process doing the consolidation
* re-run with "config.logging_level": 5 in the Ctx config and send me that

atolopko-czi:
https://github.com/single-cell-data/TileDB-SOMA/issues/1670

atolopko-czi:
A single, ad hoc builder run _succeeded_ on consolidation (with level 5 tiledb log level in place), but this did not reproduce the error. Running second attempt of consolidation for just `homo_sapiens/../X/normalized`.

[...]

PR #745, opened 2023-09-01
add builder dmesg logging on error

atolopko-czi:
For https://github.com/chanzuckerberg/cellxgene-census/issues/744

### DIFF
--- a/tools/cellxgene_census_builder/entrypoint.sh
+++ b/tools/cellxgene_census_builder/entrypoint.sh
@@ -2,7 +2,7 @@
 
 python3 -m cellxgene_census_builder .
 BUILDER_STATUS=$?
-# On error, log dmesg to aid troubleshooting
+# On error, log dmesg tail to aid troubleshooting
 # Note: requires docker --privileged option
-if [[ $BUILDER_STATUS -ne 0 ]]; then dmesg; fi
+if [[ $BUILDER_STATUS -ne 0 ]]; then dmesg | tail -n 128; fi
 exit $BUILDER_STATUS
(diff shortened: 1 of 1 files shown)


tools/cellxgene_census_builder/entrypoint.sh
```diff
@@ -1,3 +1,8 @@
 #!/bin/bash
 
 python3 -m cellxgene_census_builder .
+BUILDER_STATUS=$?
+# On error, log dmesg to aid troubleshooting
+# Note: requires docker --privileged option
+if [[ $BUILDER_STATUS -ne 0 ]]; then dmesg; fi
```
bkmartinjr, line 7: nit 1: is there any way to test if privileged and only call if affirmative?

nit 2: I"m not sure we need all of the kernel log.  `dmesg | tail -100` is probably sufficient for our OOM/etc purposes

bkmartinjr (approved):

tools/cellxgene_census_builder/entrypoint.sh
atolopko-czi: > nit 1: is there any way to test if privileged and only call if affirmative?

From within the container, it's (reportedly) not possible. Other than just trying the command and handling the error. Since this is only run in an builder erroring state, then container will exit with a non-zero status either way. So I'm just going to leave it as is.

> 
> nit 2: I"m not sure we need all of the kernel log. `dmesg | tail -100` is probably sufficient for our OOM/etc purposes

Done! Though I took poetic license on the tail size.

Issue #467, opened 2023-05-09
Add information to notebooks about `is_primary_data`


pablo-gar:
## Description

It has come to our attention that `is_primary_data` doesn't land with, and could potentially lead to silent errors in downstream analysis where they unintentionally get duplicate cells. 

## Mitigation

@jahilton suggests 

> Include is_primary_data in every single tutorial so a user is forced to be aware of it.
-even for all cell use cases, specify is_primary_data=[True, False]

## TODO

- [ ] Add the text below as a recommendation in the notebook editorial guidelines.
- [ ] Add the text below to all notebooks

```
⚠️ Note that the Census data includes:

* Full-gene sequencing read counts (e.g. Smart-Seq2) and molecule counts (e.g. 10X).
* Duplicate cells present across multiple datasets, these can be filtered in or out using the cell metadata variable `is_primary_data`.
```

pablo-gar:
Completed #754

PR #754, opened 2023-09-06
[docs][python] Reinforce knowledge of `is_primary_data` in all notebooks 

pablo-gar:
Addresses #467 for Python. Once approved, R work follows.


api/python/notebooks/analysis_demo/comp_bio_census_info.ipynb
```diff
@@ -18,6 +17,9 @@
     "- Census summary content tables\n",
     "- Understanding Census contents beyond the summary tables\n",
     "\n",
+    "⚠️ Note that the Census RNA data includes duplicate cells present across multiple datasets, these can be filtered in or out using the cell metadata variable `is_primary_data`.\n",
```
bkmartinjr, line 20: run-on sentence.  Suggest breaking into two, e.g.,

> Note that the Census RNA data includes duplicate cells present across multiple datasets. Duplicate cells can be filtered in or out using the cell metadata variable `is_primary_data`.

I would also add a link to the normative section of the CxG schema that defines this attribute.

bkmartinjr (changes requested):
suggest copyedit & link to CxG schema

atolopko-czi:
Do we need to consider an API-level change to better support the (seemingly) common use of wanting to filter primary data?

api/python/notebooks/analysis_demo/comp_bio_census_info.ipynb
atolopko-czi: +1

atolopko-czi (changes requested):

atolopko-czi (approved):

bkmartinjr (approved):

Issue #727, opened 2023-08-22
Develop a proposal for a design of a Census tokenizer that creates modeling-ready data for Geneformer 


pablo-gar:
### Relevance

Geneformer was identified as a potential model that could offer Census-integrated assets and/or models for the following high-priority Census work: [The Census includes pre-computed results with specific focus on integration algos which simplify access, usability and expand the user base (e.g. embeddings, pre-trained models, summary statistics).](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell/506) In particular for the following areas.

1. Embedding generation.
2. Cell type prediction.
3. Identification of key genes involved disease.

In addition there are other promising areas where Geneformer can provide insights to Census users. As such providing modeling-ready Census data for Geneformer is valuable for internal and external use of Census data.

### Tokenization

A key aspect of Geneformer is that it requires **tokenized** data after data normalization. The tokenization involves the following steps:

- Normalize cells by library size (row sums)
- Normalize cells by median expression per gene across all cells at ["Genecorpus" data](https://huggingface.co/datasets/ctheodoris/Genecorpus-30M/tree/main/example_input_files)
- Rank genes per cell, high-to-low. 
- Convert genes to integer tokens
- Export to hugging face `Dataset` class.

### The problem: lack of a scalable tokenizer.

[The current tokenizer provided by Geneformer](https://huggingface.co/ctheodoris/Geneformer/blob/main/geneformer/tokenizer.py) has proven hard to scale in the context of Census during internal testing. This is due to the following pain points of the tokenizer:

1. It is implemented for loom, so Census data needs to be exported to AnnData and then saved as loom.
2. It makes a hugging face Dataset using the `Dataset.from_dict()` method which somehow cannot convert large matrices due to an error in the inference of token vector encoding (encoded as int16 rather than int64).
4. It calculates token sizes after creating the dataset (using `.map()`) and it frequently breaks with multiprocessing of the length calculation.
5. It creates tokens in an inefficient way for sparse matrices, not taking advantage of the csr format, which Census guarantees. 

## Definition of Done

A proposed design of a Census tokenizer that:

- Performs iterative creation of tokenized `Datasets` for memory-efficient tokenization.
- Works natively with Census (`tiledbsoma`) 
- It's flexible and extensible for future potential different tokenization methods different from Geneformer. 

## Example tokenizer for in-memory Census data (from anndata)

[...]

PR #762, opened 2023-09-11
[python] Geneformer Dataset builder

mlin:
#727

Adds an experimental `GeneformerTokenizer` to generate a Hugging Face Dataset with Geneformer token sequences derived from Census query results. This is factored with an abstract `CellDatasetBuilder` that captures the general pattern of processing Census query results into a cell Dataset, and a concrete subclass implementing the Geneformer tokenizer particulars within that pattern.

The small unit test data are consistent with results generated by @pablo-gar's `census_adata_tokenizer.py`.


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/geneformer_tokenizer.py
```diff
@@ -0,0 +1,145 @@
...
+
+from .cell_dataset_builder import CellDatasetBuilder
+
+try:
+    import geneformer
+except ImportError:
+    # pyproject.toml can't express Geneformer git+https dependency
+    raise ImportError(
+        "Please install Geneformer with: " "pip install git+https://huggingface.co/ctheodoris/Geneformer@39ab62e"
+    ) from None
```
mlin, line 17: @pablo-gar Here's how I've handled the Geneformer dependency for now. As we discussed, there are a couple problems with making it a conventional Python dependency (pyproject.toml doesn't yet allow this `git+https` scheme; and one of its transitive dependencies needs cython). And it's just the two .pkl files I'm using from it.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/cell_dataset_builder.py
```diff
@@ -0,0 +1,135 @@
...
+
+from cellxgene_census.experimental.util import X_sparse_iter
+
+
+class CellDatasetBuilder(ContextManager["CellDatasetBuilder"], ABC):
+    """
+    Abstract base class for methods to process CELLxGENE Census query results into a
+    Hugging Face Dataset in which each item represents one cell. Subclasses implement
+    the `cell_item()` method to process the expression vector for a cell into a Dataset
+    item, and may optionally override `__init__()` and `__enter__()` to perform any
```
atolopko-czi, line 18:
```suggestion
    item. Acts as a context manager, so subclasses may optionally override `__init__()` and `__enter__()` to perform any
```
if only to explicitly mention in the docstring that it's a ContextManager

api/python/cellxgene_census/pyproject.toml
```diff
@@ -47,6 +47,11 @@ experimental = [
     "scikit-learn==1.2.2",
     "scikit-misc==0.2.0",
     "psutil==5.9.5",
+    "datasets~=2.14",
+    "tdigest~=0.5",
+    # Not expressible in pyproject.toml:
+    #"git+https://huggingface.co/ctheodoris/Geneformer",
+    # instead, experimental/ml/geneformer_tokenizer.py catches ImportError to ask user to install that.
```
atolopko-czi, line 54: Should add user documentation for this. Maybe in the `GeneformerTokenizer` class docstring.  I'm realizing that no where do we mention in user docs how to install `experimental` packages, so maybe we should add that and at the same time mention this GeneFormer special-case.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/cell_dataset_builder.py
```diff
@@ -0,0 +1,135 @@
...
+    subclasses after the context has been entered.
+    """
+
+    def __init__(
+        self,
+        experiment: tiledbsoma.Experiment,
+        *,
+        measurement_name: str = "RNA",
+        layer_name: str = "raw",
+        cells_query: Optional[tiledbsoma.AxisQuery] = None,
```
atolopko-czi, line 40: We use `{obs,var}_query` elsewhere, so maybe stick with that for consistency.

atolopko-czi (changes requested):

api/python/cellxgene_census/src/cellxgene_census/experimental/util/_csr_iter.py
```diff
@@ -81,42 +100,51 @@ def X_sparse_iter(
     obs_coords = query.obs_joinids().to_numpy()
     obs_coord_chunker = (obs_coords[i : i + stride] for i in range(0, len(obs_coords), stride))
 
-    # Lazy read into Arrow Table. Yields (coords, Arrow.Table)
     X = query._ms.X[X_name]
     var_coords = query.var_joinids().to_numpy()
+
+    # Lazy read of X chunk. Yields ((obs_coords, var_coords), (soma_data, soma_dim_0, soma_dim_1))
```
bkmartinjr, line 106: comment says this is lazy, but it is not.  It always returns a materialized tuple.

I suspect this is an error with the comment, as the `table_reader` is still lazy, and there is no benefit to making the table-to-tuple-of-numpy conversion lazy.  

Suggest just fixing the comment - perhaps by moving it down a few lines

api/python/cellxgene_census/src/cellxgene_census/experimental/util/_csr_iter.py
```diff
@@ -81,42 +100,51 @@ def X_sparse_iter(
...
         )
-        for (obs_coords_chunk, var_coords), tbl in table_reader
-    )
-    if use_eager_fetch:
-        coo_reindexer = (t for t in _EagerIterator(coo_reindexer, query._threadpool))
+        if use_eager_fetch:
+            coo_reindexer = (t for t in _EagerIterator(coo_reindexer, query._threadpool))
 
-    # lazy convert Arrow table to Scipy sparse.csr_matrix
+    # lazy convert to Scipy sparse.csr_matrix
```
bkmartinjr, line 137: This comment is also wrong - it can convert to CSC _or_ CSR (or it used to do that).  The variable name is also misleading.   (yes, some of this is my fault :-)

Suggest clarifying - the `fmt_ctor` controls the format, via the user-provided `fmt` flag

api/python/cellxgene_census/src/cellxgene_census/experimental/util/_csr_iter.py
```diff
@@ -81,42 +100,51 @@ def X_sparse_iter(
...
+                ),
+            )
+            for (obs_coords_chunk, var_coords), (soma_data, soma_dim_0, soma_dim_1) in table_reader
         )
-        for (obs_coords_chunk, var_coords), tbl in table_reader
-    )
-    if use_eager_fetch:
-        coo_reindexer = (t for t in _EagerIterator(coo_reindexer, query._threadpool))
+        if use_eager_fetch:
+            coo_reindexer = (t for t in _EagerIterator(coo_reindexer, query._threadpool))
```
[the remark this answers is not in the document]
bkmartinjr: No -- it won't pipeline (in a separate thread) the reindexing operation unless you use an _EagerIterator

Basically, passing `use_eager_fetch` stuffs the following into separate theads:
* read from SOMA
* reindex
* reorganize coo->cs{c,r}
if you omit that line, it would combine the last two steps into a single thread

bkmartinjr (changes requested):
@mlin - can you comment on the impact of this on memory usage?  In particular, when reindexing is disabled on the primary axis (the non-compressed axis), and the soma_joinids are very large.

I'm not sure the change as proposed makes sense, as it can't handle incremental reads of very large arrays anymore -- it will likely run out of memory (or best case, allocate but never use a great deal of memory).

Concrete example (using the SciPy spmatrix names for each field):
* do a query where the COO table looks like (obs is dim_0)
```
dim_0          dim_1      val
10000000       0          2.0
```
* convert CSR


The resulting length of each spmatrix field is now:
* data:  length 1
* indices:  length 1
* indptr:  10,000,000  (so total memory is 80MiB)

Size of `indptr` is always the max of the primary dimension coordinate value.

The original design required reindexing because it provided a sliding window over the X matrix query, in a memory effiicent manner.  The caller has access to the obs/query soma_joinids via the ExperimentAxisQuery, so there is no need to preserve them.  And the reindexing on both axis was required to keep the memory footprint small.

I don't know your actual use case, so it is hard for me to comment on what motivated this change, but I don't think it is a good idea.

mlin:
@bkmartinjr thanks, very educational for me. What would you think of constraining the new option to apply only to the sparse axis? i.e. setting `reindex_sparse_axis=False` would leave var indexed by soma_joinid for csr (and obs for csc).

The use cases in mind (for csr) are when retrieving all of var anyway, or when the app needs to reindex var in some other way regardless (applicable here).

I had no actual need/want to disable obs reindexing, supported it just for symmetry.

bkmartinjr:
> constraining the new option to apply only to the sparse axis

This would definitely fix the memory issue.

Can you elaborate the use case for disabling reindexing on the sparse axis?  It is a very trivial operation for the caller to reverse, so there is no information loss (i.e., the caller has the vector of join ids, and can simply do a lookup in that vector).   The only use case (i.e., motivation) I can think of is one of performance, and I'm wondering how much difference it makes in practice.

mlin:
@bkmartinjr performance and also just simplicity. I have an entirely different "space" that I have to project gene IDs (soma_dim_1) onto, namely the transformer token numbers defined by the Geneformer model. I precompute that mapping once before starting the chunked read, so it's nice to be able to index the X chunks directly with it, instead of first reversing the reindexing we just spent time computing. That's easily doable as you say, but the reindexing is entirely wasted and we have to think about three id spaces instead of two (and the additional space is theoretically specific to each chunk of X). 

I can certainly live without the opt-out though, since I can still write my API/ABC to hide the un-reindexing from the end developer. Let me know what you think, thanks- this has been really helpful.

bkmartinjr:
Given that this is still an experimental API, and that you find it helpful (i.e., there _is_ a use case :-), I"m fine with adding it on the sparse axis.  But it should definitely "throw" an error if anyone attempts it on the non-sparse axis.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/geneformer_tokenizer.py
pablo-gar: Just pasting my comment from slack to record it here on just copying the files into the repo 

> Yes I’ve thought about this. I don’t have any concerns from the license perspective. My concern is more related to future iterations of Geneformer, specifically when they train again and their files change. I think is less error prone for the end-user to have the dependency taking care of the versioning.
> I do understand the hurdles of adding a dependency on Geneformer, but my thought is … if you (user) are going to create a Dataset to be used in Geneformer why would you do that if you are not going to use it with Geneformer?

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/__init__.py
```diff
@@ -2,6 +2,14 @@
 An API to facilitate use of PyTorch ML training with data from the CZI Science CELLxGENE Census.
 """
 
+from .cell_dataset_builder import CellDatasetBuilder
+from .geneformer_tokenizer import GeneformerTokenizer
 from .pytorch import ExperimentDataPipe, Stats, experiment_dataloader
```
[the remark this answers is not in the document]
mlin: open to this (maybe huggingface & pytorch sub-packages), @pablo-gar POV?

api/python/cellxgene_census/tests/experimental/ml/test_geneformer.py
```diff
@@ -0,0 +1,55 @@
...
+        20987639: (1367, [8681, 10047, 9069, 6623, 14968, 16865, 7725]),
+        22222206: (2048, [15933, 15623, 8809, 754, 25306, 411, 6872]),
+        23456773: (2048, [15633, 11385, 16997, 650, 17184, 3408, 7066]),
+        25925907: (1589, [15602, 10824, 3106, 608, 8510, 13232, 24344]),
+        28395041: (589, [6556, 19788, 18489, 2124, 10509, 2218, 6567]),
+        30864175: (815, [10057, 9500, 2936, 21070, 13659, 5081, 9209]),
+        32098742: (2048, [4067, 948, 1324, 5261, 16985, 1511, 10268]),
+    }
+
+    with cellxgene_census.open_soma(census_version="2023-09-04") as census:
```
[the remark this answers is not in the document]
mlin: The trouble right now is we're supposed to use the new `normalized` layer (@pablo-gar correct?). I added a comment about updating to the next LTS release when we make it.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/cell_dataset_builder.py
mlin: @atolopko-czi thanks, when I got into addressing this and your other comment about obs/var terminology, I ended up significantly refactoring `CellDatasetBuilder` to inherit `ExperimentAxisQuery` directly instead of wrapping it. This simplified it significantly and I also improved the docstring showing the usage.

api/python/cellxgene_census/pyproject.toml
mlin: @atolopko-czi Good idea, added to the docstring. Also added a blurb about the experimental modules to cellxgene_census_docsite_installation.md (although I'm reluctant to discuss Geneformer specifically there to avoid one more place to repeat the desired git ref. I hope the specific `ImportError` message one gets at runtime suffices; as @pablo-gar pointed out, if you're going to be using this then it's probably not too surprising to have to install geneformer too)

mlin:
@bkmartinjr revised the diff to `X_sparse_iter` as we discussed, with `reindex_sparse_axis` that won't allow opting out of reindexng the dense-indptr axis. This streamlined the diff slightly since there's no path without any reindexing now (which, again, wasn't actually needed/wanted). I *think* I got the inline comments too but appreciate a double-check.

api/python/cellxgene_census/src/cellxgene_census/experimental/util/_csr_iter.py
```diff
@@ -94,35 +114,44 @@ def X_sparse_iter(
     if use_eager_fetch:
         table_reader = (t for t in _EagerIterator(table_reader, query._threadpool))
 
-    # lazy reindex of obs coordinates. Yields coords and (data, i, j) as numpy ndarrays
+    # Lazy reindexing of soma_dim_0 and soma_dim_1 to obs_coords and var_coords
+    # positions (except one or the other if not reindex_sparse_axis).
+    # Yields coords and COO (data, i, j) as numpy ndarrays
```
bkmartinjr, line 119: nit: I found this comment is slightly confusing - it yields `((obs_coords, var_coords), (data, i, j))` so perhaps be specific?

api/python/cellxgene_census/src/cellxgene_census/experimental/util/_csr_iter.py
```diff
@@ -19,12 +19,15 @@ def X_sparse_iter(
     stride: int = 2**16,
     fmt: Literal["csr", "csc"] = "csr",
     use_eager_fetch: bool = True,
+    reindex_sparse_axis: bool = True,
```
bkmartinjr, line 22: CC, *FYI only*: @beroy - this use case will need to eventually be ported to fast reindexing and COO conversion, when they are available.

bkmartinjr (approved):
@mlin - I only reviewed the changes to the CSR iterator.  It LGTM!  Thanks for your work on this.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/geneformer_tokenizer.py
```diff
@@ -0,0 +1,154 @@
...
+
+    ```
+    import cellxgene_census
+    import tiledbsoma
+    from cellxgene_census.experimental.ml import GeneformerTokenizer
+
+    with cellxgene_census.open_soma() as census:
+        with GeneformerTokenizer(
+            census["census_data"]["homo_sapiens"],
+            obs_query=tiledbsoma.AxisQuery(...),  # define some subset of Census cells
```
pablo-gar, line 32: Let's provide an actual example with a value_filter. I like filter for cells from tissue == 'tongue' cause it's a very small slice

api/python/cellxgene_census/tests/experimental/ml/test_geneformer.py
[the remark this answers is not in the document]
atolopko-czi: nit: instezd, open `stable`, check for normalized layer, and if it doesn't exist, revert to 2023-09-04.  This will save us frustration of future test breakage. Comment can then just indicate "remove after..."

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/__init__.py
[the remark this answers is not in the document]
pablo-gar: Yeah this makes sense. I'd recommend:

- `cellxgene_census.experimental.pytorch` and
- `cellxgene_census.experimental.huggingface`

@mlin for this PR you can just move your work to `cellxgene_census.experimental.huggingface` I'll make a separate ticket for moving the pytorch code

api/python/cellxgene_census/pyproject.toml
pablo-gar: `ImportError` works great for now

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/cell_dataset_builder.py
```diff
@@ -0,0 +1,106 @@
...
+        - `measurement_name`: Measurement in the experiment, default "RNA".
+        - `layer_name`: Name of the X layer to process, default "raw".
+        - `kwargs`: passed through to `ExperimentAxisQuery()`, especially `obs_query`
+           and `var_query`.
+        """
+        super().__init__(experiment, measurement_name, **kwargs)
+        self.layer_name = layer_name
+        self._cells_per_chunk = _cells_per_chunk
+
+    def build(self, from_generator_kwargs: Optional[Dict[str, Any]] = None) -> Dataset:
```
pablo-gar, line 60: To be tested in the near future @mlin `Dataset.from_generator` has a `num_proc` argument. We should test if when > 1 there's any conflicts with the thread pools from `X_sparse_iter`.

In my experience the multi-threading from `Dataset` methods work great and scale well with large data.

pablo-gar: Tagging @atolopko-czi to see if he has any thoughts as he may have (?) some working experience with this in PyTorch

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/geneformer_tokenizer.py
```diff
@@ -0,0 +1,175 @@
...
+    ) -> None:
+        """
+        - `experiment`: Census Experiment to query
+        - `obs_query`: obs AxisQuery defining the set of Census cells to process (default all)
+        - `obs_attributes`: names of attributes to propagate from the experiment obs dataframe
+           (cell metadata) into each Dataset item
+        - `max_input_tokens`: maximum length of Geneformer input token sequence (default 2048)
+        - `token_dictionary_file`, `gene_median_file`: pickle files supplying the mapping of
+          Ensembl human gene IDs onto Geneformer token numbers and median expression values.
+          By default, these will be loaded from the Geneformer package.
```
pablo-gar, line 72: missing 

`gene_median_file` docstring

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/geneformer_tokenizer.py
```diff
@@ -0,0 +1,175 @@
...
+        super().__init__(
+            experiment,
+            measurement_name="RNA",
+            layer_name="normalized",
+            # configure query to fetch the relevant genes only
+            var_query=tiledbsoma.AxisQuery(coords=(self.model_gene_ids,)),
+            **kwargs,
+        )
+
+    def load_geneformer_data(
```
pablo-gar, line 90: nit, unless we expect users to call this manually, let's underscore it
```suggestion
    def _load_geneformer_data(
```

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/geneformer_tokenizer.py
```diff
@@ -0,0 +1,175 @@
...
+        ), "Experiment RNA measurement lacks 'normalized' layer; try 'latest' Census version (2023-08-01 or newer)"
+
+        self.max_input_tokens = max_input_tokens
+        self.obs_attributes = set(obs_attributes) if obs_attributes else set()
+        self.load_geneformer_data(experiment, token_dictionary_file, gene_median_file)
+        super().__init__(
+            experiment,
+            measurement_name="RNA",
+            layer_name="normalized",
+            # configure query to fetch the relevant genes only
```
pablo-gar, line 85: Nice!

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/geneformer_tokenizer.py
```diff
@@ -0,0 +1,175 @@
...
+        self.obs_df = self.obs(column_names=obs_column_names).concat().to_pandas().set_index("soma_joinid")
+        return self
+
+    def cell_item(self, cell_joinid: int, cell_Xrow: scipy.sparse.csr_matrix) -> Dict[str, Any]:
+        """
+        Given the expression vector for one cell, compute the Dataset item providing
+        the Geneformer inputs (token sequence and metadata).
+        """
+        # project cell_Xrow onto model_gene_ids
+        model_counts = cell_Xrow[:, self.model_gene_ids]
```
pablo-gar, line 157: Hard to tell the impact on performance, but this may not be taking advantage of sparsity, unless scipy  during the multiplication step below is only using the non-zero values, which I doubt.

In other words we may be doing arithmetic operations over the full 20K genes when in reality most of those are 0s which can be left out during multiplication. For handful of cells that may not be a huge deal, but maybe it is impactful over millions of cells.

For reference see my use of csr's `data` slot for this operation in my prototype

https://gist.github.com/pablo-gar/7cbba76caa3f3cfb7b5ebf96b37d2644#file-census_adata_tokenizer-py-L43

api/python/cellxgene_census/tests/experimental/ml/test_geneformer.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def select_census_version_for_geneformer_tests() -> str:
+    # GeneformerTokenizer needs the "normalized" X layer which wasn't yet available in
+    # "stable" at the time this was written. This should provide a graceful transition
+    # once we next advance "stable" (after which it could be eliminated).
+    with cellxgene_census.open_soma(census_version="stable") as stable:
+        if "normalized" in stable["census_data"]["homo_sapiens"].ms["RNA"].X:
+            return "stable"
+    return "2023-09-04"
```
pablo-gar, line 20: This means that we should label this release to not be deleted anytime soon, otherwise the test will just fail. Can you coordinate with @atolopko-czi and @ebezzi  on this

pablo-gar (changes requested):
- Two small fixes, of for doc-strings and  another one for marking `load_geneformer_data` as "private"
- I'd appreciate your thoughts on my comment about arithmetic multiplication for each cell with mostly 0s
- And please mark in the release.json to not delete the Census version that is used for testing here 

Only the first bullet point needs to be addressed for the PR to be merged, should be very quick

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/geneformer_tokenizer.py
mlin: They're both covered by this line.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/geneformer_tokenizer.py
mlin: I think this is OK -- I added some assertions to confirm `model_counts` and `model_expr` remain sparse i.e. their `data` vectors contain no explicit zeroes. (Assertions checked out but are now commented out)

Then, [here is the scipy.sparse code](https://github.com/scipy/scipy/blob/ece199257f857f8a2418587433cfb8b4030a19e1/scipy/sparse/_compressed.py#L467-L469) for efficient multiplication of a `cs{r,c}_matrix` times a dense vector.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/geneformer_tokenizer.py
pablo-gar: duh! I missed it!

atolopko-czi (approved):
LGTM

Issue #751, opened 2023-09-05
[api] [python] [r] Implement nightly package builds to test integrity of Python and R APIs with fuzzy dependency pins 


pablo-gar:
We want to ensure that any patch-level updates to TileDB-SOMA will be readily tested against the Census API, to catch potential regressions. This will allow the TileDB-SOMA package dependency to more confidently use a fuzzy version pin for the patch level. E.g. `tiledbsoma~=1.4.0`.

Schedule the [Python unit tests GHA workflow](https://github.com/chanzuckerberg/cellxgene-census/blob/main/.github/workflows/py-unittests.yml
) and [R unit tests GHA workflow](https://github.com/chanzuckerberg/cellxgene-census/blob/main/.github/workflows/r-check.yml) to run on a nightly [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) 

Notes:
* It does not seem necessary to test the Census API against all previous patch levels (e.g. using a GHA matrix), since users installing the Census anew will always get the latest patch level of `tiledbsoma`. We only need to ensure the latest patch, as of "today", is working correctly. (Also, we do not expect > 1 patch level TileDB-SOMA release per day, sou nightly testing should suffice.)

Ticket can be closed when work is completed for the builds of:

- [x] Python
- [x] R

atolopko-czi:
@pablo-gar I think we agreed this shouldn't be high priority. That said, it's almost no work...

atolopko-czi:
Should we run the R tests on a nightly basis as well? Does the tiledbsoma r-universe dependency get pinned at build time or user installation time? If the latter, then we should.

pablo-gar:
> @pablo-gar I think we agreed this shouldn't be high priority. That said, it's almost no work...

I see as high priority but not a blocker for our last release.

bkmartinjr:
@pablo-gar - AFAIK, this is complete.  Can you close if you and @mlin agree?

pablo-gar:
Agreed! @mlin ?

bkmartinjr:
Closing this, as it appears all necessary changes have landed

PR #763, opened 2023-09-11
[python][misc] GHA for Python dependency checks

bkmartinjr:
Partial solution for #751

### DIFF
--- a/.github/workflows/py-dependency-check.yml
+++ b/.github/workflows/py-dependency-check.yml
@@ -29,7 +29,7 @@ jobs:
       - name: get latest release with tag
         id: latestrelease
         run: |
-          echo "::set-output name=releasetag::$(curl -s https://api.github.com/repos/chanzuckerberg/cellxgene-census/releases/latest | jq '.tag_name' | sed 's/\"//g')"
+          echo "releasetag=$(curl -s https://api.github.com/repos/chanzuckerberg/cellxgene-census/releases/latest | jq '.tag_name' | sed 's/\"//g')" >> "$GITHUB_OUTPUT"
 
       - name: confirm release tag
         run: |
@@ -39,6 +39,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ steps.latestrelease.outputs.releasetag }}
+          fetch-depth: 0
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
--- a/.github/workflows/py-dependency-check.yml
+++ b/.github/workflows/py-dependency-check.yml
@@ -12,6 +12,7 @@ name: Python dependency checks
 on:
   schedule:
     - cron: "30 1 * * *"
+  workflow_dispatch: # Used for testing
 
 jobs:
   python-dependency-check:
@@ -38,7 +39,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ steps.latestrelease.outputs.releasetag }}
-          fetch-depth: 0
+          fetch-depth: 1
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/py-dependency-check.yml
```diff
@@ -0,0 +1,76 @@
...
+# continues to function correctly using the latest upstream dependencies.
+#
+# Method:
+# 1. Check out the repo at the latest release tag
+# 2. Install latest cellxgene-census package from pypi
+# 3. Verify versions match
+# 4. Run tests from the repo
+
+on:
+  schedule:
```
bkmartinjr, line 13: do we want this to run on any other events?

if not, how will we monitor that it is working?  Badge on the REAMDE?

mlin (approved):
@bkmartinjr Sure, the R version will actually be a little simpler because the release package tarballs include the test code. (a default behavior of the R packager for whatever reason)

bkmartinjr:
@mlin - do you want to add a workflow to this PR, or send me notes on an approach?  I'm fine with either.

.github/workflows/py-dependency-check.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: confirm release tag
+        run: |
+          echo Latest GH release tag ${{ steps.latestrelease.outputs.releasetag }}
+
+      - name: checkout latest release of repo
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ steps.latestrelease.outputs.releasetag }}
+          fetch-depth: 0
```
atolopko-czi, line 41: `0` checks out full history--do we need that? if not, use `1`

atolopko-czi (changes requested):
LGTM, but a few minor points to consider and a suggestion for sending notifications.

.github/workflows/py-dependency-check.yml
bkmartinjr: @pablo-gar - you can make the call. Either in this PR or elsewhere.  This is not the only workflow that needs such a mechanism.

.github/workflows/py-dependency-check.yml
```diff
@@ -0,0 +1,76 @@
...
+            echo Release tags match
+          fi
+
+      - name: Test with pytest (API, main tests)
+        run: |
+          PYTHONPATH=. pytest -v -rP ./api/python/cellxgene_census/tests/
+
+      - name: Test with pytest (API, experimental)
+        run: |
+          PYTHONPATH=. pytest -v -rP --experimental ./api/python/cellxgene_census/tests/experimental
```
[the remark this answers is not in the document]
bkmartinjr: I literally implemented "execute same workflow as unit tests".  Do you recall the aforementioned rationale?

.github/workflows/py-dependency-check.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies (including experimental)
+        run: |
+          python -m pip install -U pip setuptools wheel
+          pip install -U -r ./api/python/cellxgene_census/scripts/requirements-dev.txt
+          pip install -U cellxgene-census[experimental]
```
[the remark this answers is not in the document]
bkmartinjr: no.  Not sure why that was done other than it is harmless and extremely cautious

.github/workflows/py-dependency-check.yml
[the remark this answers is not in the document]
atolopko-czi: I wish I remembered! I don't see a need for it now.

atolopko-czi (approved):
LGTM, and we can add notifications as a follow-up issue, if needed.

.github/workflows/py-dependency-check.yml
pablo-gar: Unless any strong concerns let's go with slack since that's already a convention used by the team

pablo-gar:
@mlin can you complete the R work in a different PR? 

I will add tasks to the original issue #751, one for Python one for R. We can close the ticket once both have been completed.

.github/workflows/py-dependency-check.yml
bkmartinjr: I am unable to figure out how this works with a quick review, so I'm putting this PR on a back-burner for a while.

mlin:
@pablo-gar @bkmartinjr Yes, no problem, I should be able to put in a separate PR for R tomorrow.

bkmartinjr:
After discussion with @pablo-gar, decided to add a status badge and consider slack integration for a later date.

.github/workflows/py-dependency-check.yml
bkmartinjr: added badge to README.md

PR #770, opened 2023-09-14
[builder] Add additional debug logging

ebezzi:
Add additional builder logging, namely: 

1. Output of `pip freeze`
2. TileDB configuration used
3. If the builder fails, output of `df` to check the disk space situation

### DIFF
--- a/tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
+++ b/tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
@@ -2,11 +2,14 @@ import argparse
 import logging
 import os
 import pathlib
+import subprocess
 import sys
 from typing import Callable, Sequence
 from urllib.parse import urlparse
 
 import s3fs
+
+from tools.cellxgene_census_builder.src.cellxgene_census_builder.build_soma.globals import DEFAULT_TILEDB_CONFIG
 
 from .build_state import CENSUS_BUILD_CONFIG, CENSUS_BUILD_STATE, CensusBuildArgs, CensusBuildConfig, CensusBuildState
 from .util import log_process_resource_status, process_init, urlcat
@@ -49,6 +52,7 @@ def main() -> int:
         "build": [
             do_prebuild_set_defaults,
             do_prebuild_checks,
+            do_log_environment,
             do_build_soma,
             do_validate_soma,
             do_create_reports,
@@ -66,6 +70,7 @@ def main() -> int:
         "full-build": [
             do_prebuild_set_defaults,
             do_prebuild_checks,
+            do_log_environment,
             do_build_soma,
             do_validate_soma,
             do_create_reports,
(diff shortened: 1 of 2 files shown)


tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -150,11 +155,30 @@ def do_prebuild_checks(args: CensusBuildArgs) -> bool:
...
+    # Log pip freeze output
+    import pkg_resources
+
+    deps = ""
+    for dep in pkg_resources.working_set:
+        deps += f"{dep.as_requirement()}\n"
+    logging.info(f"Pip freeze:\n{deps}")
+
+    # Log TileDB Config
+    logging.info(f"TileDB Config:\n{DEFAULT_TILEDB_CONFIG}")
```
ebezzi, line 168: This looks hacky, but the consolidate step is referencing this global variable, so I think this is the only way to do this.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
```diff
@@ -150,11 +155,30 @@ def do_prebuild_checks(args: CensusBuildArgs) -> bool:
...
+
+    return True
+
+
 def do_build_soma(args: CensusBuildArgs) -> bool:
     from .build_soma import build as build_a_soma
 
     if (cc := build_a_soma(args)) != 0:
         logging.critical(f"Build of census failed with code {cc}.")
+
```
bkmartinjr, line 178: What is the rationale for putting this logging in the package, vs. in the docker `entrypoint.sh`?  We currently do `dmesg` loggging in entrypoint.sh, and it would be good to be consistent.

My _opinion_ is that it belongs in entrypoint.sh, at the end of the run (as in the `dmesg` logging).  Rationale: we only need this introspection when running in Batch/container. Any other time, the user is running from the CLI and can do this type of system introspection on their own.

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
bkmartinjr: I don't think this is useful information, as it is just printing out the hard-coded constant.

What we want to run (i.e., what John requested) is: `tiledbsoma.show_package_versions()`

Or if from entrypoint.sh, `python -c 'print(tiledbsoma.show_package_versions())'`

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
bkmartinjr: https://github.com/chanzuckerberg/cellxgene-census/blob/deceee6411286d0f8915d5100c39a2d4621f0301/tools/cellxgene_census_builder/entrypoint.sh#L7

bkmartinjr: I don't like putting it into the mainline script, as it isn't needed except for debugging the container.

That said, I'm also not too opinionated.  @atolopko-czi established the convention of putting container/EBS specific stuff into entrypoint.sh, so I'll let you guys hash it out.

bkmartinjr (approved):
LGTM!

tools/cellxgene_census_builder/src/cellxgene_census_builder/__main__.py
atolopko-czi: In the other logging PR, `DEFAULT_TILEDB_CONFIG` will be changed to add logging config, so not entirely hard-coded anymore. Which led me to this thought: https://github.com/chanzuckerberg/cellxgene-census/pull/767#issuecomment-1719910261

PR #775, opened 2023-09-19
[python] Add pytorch global shuffling

### LINKED ISSUE #538: ExperimentDataPipe data shuffling
Shuffling of training data load via `ExperimentDataPipe` can currently be accomplished using PyTorch's [Shuffler](https://pytorch.org/data/beta/generated/torchdata.datapipes.iter.Shuffler.html#torchdata.datapipes.iter.Shuffler) pipe. However, this operation can only shuffle data within a fixed batch size (since data is streamed). If the shuffle batch size is large enough, the randomization may be acceptable. But this depends upon the ordering of the Census data (along the obs axis) and upon the model architecture being trained. E.g., model is being trained on multiple tissues, but the data from a given dataset is for a single tissue and larger than the shuffle batch size, then Shuffler will not produce a randomized order where different tissues are interleaved.

So it may be beneficial to support an initial full shuffling along the obs axis of the query result. This would allow Shuffler's batch size to be less of a concern, since Shuffler's batch-size shuffling would occur on an already-randomized ordering of Census data.

This could be implemented by ExperimentDataLoader fairly easily, since the obs soma_joinids are retrieved fully before X data iteration commences. The training-time retrieval of X/obs data would occur in the global randomized order.

### LINKED ISSUE #674: PyTorch soma_buffer_bytes tuning
Currently, the `soma_buffer_bytes` (indirectly) determines the number of rows (cells) to be returned by each SOMA batch. But, currently, it also controls the chunk sizes of X data retrieval for the SOMA batch, resulting in unnecessarily small chunk sizes, which is inefficient. Instead, the X data retrieval should be performed with large buffers to minimize the number of TileDB reads, and so it should not be constrained by the obs retrieval buffer size. Note that since the totality of the X data for a given SOMA batch needs to be held in memory anyway, using larger buffers for X should not result in increased peak memory usage during processing of a given SOMA batch.

- [x] Introduce a row count-based `soma_chunk_size` param to replace the less intuitive `soma_buffer_bytes` param. SOMA reads should use the default Census API `{soma,py}.init_buffer_bytes` for the low-level reads.
- [ ] ~Add documentation that explains how PyTorch memory usage can be tuned via the obs `soma_chunk_size` param. Possibly demonstrate in a notebook example, if useful.~ [split out into #782]

This is follow-up work to  #653.

atolopko-czi:
Resolves #538 
Resolves #674 

Changes:
- Add `ExperimentDataPipe()` `shuffle` param (flag). When set, shuffling is performed in two steps for performance reasons: 1) a global shuffling of contiguous ranges of data, where the rows are grouped into chunks and all chunks are shuffled, and 2) a local shuffling, where each chunk is shuffled internally.  The global shuffling ensures that the ordering of result data will be randomized across the entire obs axis, while also reducing the number of SOMA read operations by leveraging the spatial locality of contiguously stored data (i.e. reducing the number of times a tile is read). The local shuffling ensures the data is shuffled at the row-level within each chunk.
- Refactor "soma batch" term to "soma chunk", to better differentiate from the "pytorch batch" constructs in the code.
- Replace `ExperimentDataPipe()` `soma_buffer_bytes` param with `soma_chunk_size`, allowing the user to tune the SOMA chunk retrieval size using a row count instead of a byte count. While this is intended to be a more intuitive param for users (see #674), it is also used to tune shuffling randomness vs memory utilization and retrieval speed. 
- Add explicit `shuffle` param to `experiment_dataloader()` method, since shuffling is now handled by ExperimentDataPipe (this should have been a disallowed DataLoader option previously, since it was ineffectual when constructing a DataLoader with an IterDataPipe).
- Add explicit `seed` param for correct (disjoint) partitioning across worker training processes when used with `DistributedDataParallel`.
- Update `pytorch.ipynb` notebook for new shuffling param, replacing the `Shuffler` data pipe usage (still mentions its existence)


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -96,53 +99,83 @@ def __add__(self, other: "Stats") -> "Stats":
...
+        obs_column_names: Sequence[str],
+        obs_joinids: npt.NDArray[np.int64],
+        var_joinids: npt.NDArray[np.int64],
+        soma_batch_read_size: int,
+        shuffle: bool = False,
+    ):
+        self.obs = obs
         self.X = X
+        self.obs_column_names = obs_column_names
+        self.obs_joinids_chunks_iter = self.chunk_obs_joinids(obs_joinids, soma_batch_read_size, shuffle)
```
atolopko-czi, line 138: The SOMA reader for the obs DataFrame has been replaced with the list of `obs_joinids`. 
This was necessary to support global shuffling, since each soma chunk may comprise randomly-selected rows.  With the SOMA reader no longer determining the extent of each chunk, the chunking needs to be explicitly handled here now.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -77,82 +77,107 @@ class Stats:
...
+
+        obs_batch = (
+            self.obs.read(
+                coords=(obs_joinids_chunk,),
+                column_names=self.obs_column_names,
+            )
+            .concat()
+            .to_pandas()
+            .set_index("soma_joinid")
+        )
```
atolopko-czi, line 159: This replaces the single `obs.read()`, which returned an iterator PyArrow tables, with an `obs.read()` per chunk, using an explicit set of obs joinids.  These chunks may have been globally shuffled, so we need to request each chunk in this shuffled order.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -77,82 +77,107 @@ class Stats:
...
+        assert obs_batch.shape[0] == obs_joinids_chunk.shape[0]
 
         # handle case of empty result (first batch has 0 rows)
         if len(obs_batch) == 0:
             raise StopIteration
 
-        X_batch = _fast_csr.read_scipy_csr(self.X, obs_table["soma_joinid"].combine_chunks(), self.var_joinids)
+        # reorder obs rows to match obs_joinids_chunk ordering, which may be shuffled
+        obs_batch = obs_batch.reindex(obs_joinids_chunk, copy=False)
```
atolopko-czi, line 167: N.B.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -444,47 +484,46 @@ def _init(self) -> None:
         self._initialized = True
 
     @staticmethod
-    def _partition_obs_joinids(ids: npt.NDArray[np.int64]) -> npt.NDArray[np.int64]:
-        dist_partition = loader_partition = 0
-        dist_partitions = loader_partitions = 1
+    def _subset_ids_to_partition(
```
atolopko-czi, line 487: The new `_subset_ids_to_partition()` and `_compute_partitions()` methods are introduced only as a refactoring to accommodate the explicit list of (possibly) shuffled obs_joinids.  The partitioning logic is effectively the same as before, except that it now operates at the chunk level rather than the row level.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -495,25 +534,38 @@ def __iter__(self) -> Iterator[ObsAndXDatum]:
...
+        )
 
+        with _open_experiment(self.exp_uri, self.aws_region) as exp:
             obs_and_x_iter = _ObsAndXIterator(
-                X,
-                obs_tables_iter,
+                obs=exp.obs,
+                X=exp.ms[self.measurement_name].X[self.layer_name],
+                obs_column_names=self.obs_column_names,
+                obs_joinids_chunked=obs_joinids_chunked_partition,
```
atolopko-czi, line 561: We now pass in the obs ids rather than a obs_tables_iter (a SOMA Reader). The id are chunked, optionally shuffled, and optionally partitioned.

api/python/cellxgene_census/tests/experimental/ml/test_pytorch.py
```diff
@@ -38,6 +35,15 @@ def pytorch_x_value_gen(obs_range: range, var_range: range) -> spmatrix:
     return checkerboard_of_ones
 
 
+def pytorch_seq_x_value_gen(obs_range: range, var_range: range) -> spmatrix:
```
atolopko-czi, line 38: used by tests that verify obs and X are shuffled results are correctly aligned.

api/python/cellxgene_census/pyproject.toml
```diff
@@ -44,7 +44,7 @@ dependencies= [
 experimental = [
     "torch==2.0.1",
     "torchdata==0.6.1",
-    "scikit-learn==1.2.2",
+    "scikit-learn==1.3.1",
```
atolopko-czi, line 47: avoid "FutureWarning" messages related sparse matrices.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -379,16 +414,31 @@ def __init__(
...
-                The number of bytes to use for reading data from SOMA. If not specified, will use the default SOMA
-                value. Maximum memory utilization is controlled by this parameter, with larger values providing better
-                read performance.
+            soma_chunk_size:
+                The number of obs/X rows to retrieve when reading data from SOMA. This impacts two aspects of
+                ``ExperimentDataPipe`` behavior: 1) The maximum memory utilization is controlled by this
+                parameter, with larger values providing better read performance, but also requiring more memory. If
+                not specified, the value is set to utilize ~1 GiB of RAM per SOMA chunk read, based upon the number
+                of ``var`` columns (cells/features) being requested and assuming X data sparsity of 95%.
+                2) The granularity of the global shuffling step (see ``shuffle`` parameter).
```
pablo-gar, line 439: I wonder if it's worth providing some extra guidance similar to what you did in 1) e.g. "Large values lead to diminished global shuffling, and small values lead to increased global shuffling"

api/python/notebooks/experimental/pytorch.ipynb
```diff
@@ -113,17 +114,19 @@
...
     "\n",
-    "(Note: PyTorch also provides `DataSet` as a legacy interface for wrapping and accessing training data sources, but a `DataPipe` can be used interchangeably.)"
+    "The `shuffle` flag allows you to randomize the ordering of the training data for each training epoch. Note:\n",
+    "* You should use this flag instead of the `DataLoader` `shuffle` flag, as the latter does not support shuffling when used with an `IterDataPipe` dataset.\n",
+    "* PyTorch's TorchData library provides a [Shuffler](https://pytorch.org/data/main/generated/torchdata.datapipes.iter.Shuffler.html) `DataPipe`, which is alternate mechanism one can use to perform shuffling of an `IterDataPipe`. However the `Shuffler` will not \"globally\" randomize the training data, as it only \"locally\" randomizes the ordering of the training data within fixed-size \"windows\". Due the layout of Census data, a given \"window\" of Census data may be highly homogeneous in terms of its `obs` axis attribute values, and so this shuffling strategy may not provide sufficient randomization for certain types of models."
```
pablo-gar, line 129:
```suggestion
    "* PyTorch's TorchData library provides a [Shuffler](https://pytorch.org/data/main/generated/torchdata.datapipes.iter.Shuffler.html) `DataPipe`, which is an alternate mechanism one can use to perform shuffling of an `IterDataPipe`. However the `Shuffler` will not \"globally\" randomize the training data, as it only \"locally\" randomizes the ordering of the training data within fixed-size \"windows\". Due the layout of Census data, a given \"window\" of Census data may be highly homogeneous in terms of its `obs` axis attribute values, and so this shuffling strategy may not provide sufficient randomization for certain types of models.\n",
    "### Inspecting data in the `ExperimentDataPipe`\n",
    "\n"
```

api/python/notebooks/experimental/pytorch.ipynb
```diff
@@ -113,17 +114,19 @@
...
+    "This class provides an implementation of PyTorch's [DataPipe interface](https://pytorch.org/data/main/torchdata.datapipes.iter.html), which defines a common mechanism for wrapping and accessing training data from any underlying source. The `ExperimentDataPipe` class encapsulates the details of querying and retrieving Census data from a single SOMA `Experiment` and returning it to the caller as PyTorch Tensors. Most importantly, it retrieves the data lazily from the Census in batches, avoiding having to load the entire training dataset into memory at once. (Note: PyTorch also provides `DataSet` as a legacy interface for wrapping and accessing training data sources, but a `DataPipe` can be used interchangeably.)\n",
```
pablo-gar, line 117: This text section is getting pretty long, let's add some headers to aid the reader with the content
```suggestion
    "### Census `ExperimentDataPipe` in a nutshell.\n",
    "\n",
    "This class provides an implementation of PyTorch's [DataPipe interface](https://pytorch.org/data/main/torchdata.datapipes.iter.html), which defines a common mechanism for wrapping and accessing training data from any underlying source. The `ExperimentDataPipe` class encapsulates the details of querying and retrieving Census data from a single SOMA `Experiment` and returning it to the caller as PyTorch Tensors. Most importantly, it retrieves the data lazily from the Census in batches, avoiding having to load the entire training dataset into memory at once. (Note: PyTorch also provides `DataSet` as a legacy interface for wrapping and accessing training data sources, but a `DataPipe` can be used interchangeably.)\n",
```

api/python/notebooks/experimental/pytorch.ipynb
```diff
@@ -113,17 +114,19 @@
...
     "\n",
     "The constructor only requires a single parameter, `experiment`, which is a `soma.Experiment` containing the data of the organism to be used for training.\n",
```
pablo-gar, line 119:
```suggestion
    "###  The parameters of `ExperimentDataPipe` explained\n",
    "\n",
    "The constructor only requires a single parameter, `experiment`, which is a `soma.Experiment` containing the data of the organism to be used for training.\n",
```

api/python/notebooks/experimental/pytorch.ipynb
```diff
@@ -113,17 +114,19 @@
...
     "\n",
     "To retrieve a subset of the Experiment's data, along either the `obs` or `var` axes, you may specify query filters via the `obs_query` and `var_query` parameters, which are both `soma.AxisQuery` objects.\n",
     "\n",
     "The values for the prediction label(s) that you intend to use for training are specified via the `obs_column_names` array. \n",
     "\n",
-    "Finally, the `batch_size` allows you to specify the number of obs rows (cells) to return in each training iteration step. You may exclude this parameter if you want single rows (`batch_size=1`).\n",
+    "The `batch_size` allows you to specify the number of obs rows (cells) to return in each training iteration step. You may exclude this parameter if you want single rows (`batch_size=1`).\n",
```
pablo-gar, line 125: Can you also add an explanation for the SOMA chunk size?

api/python/notebooks/experimental/pytorch.ipynb
```diff
@@ -103,7 +103,8 @@
     "    X_name=\"raw\",\n",
     "    obs_query=soma.AxisQuery(value_filter=\"tissue_general == 'tongue' and is_primary_data == True\"),\n",
     "    obs_column_names=[\"cell_type\"],\n",
-    "    batch_size=16,\n",
+    "    batch_size=128,\n",
+    "    shuffle=True,\n",
```
pablo-gar, line 107: Can we add SOMA chunk size?

pablo-gar (approved):
Small changes to the docs, looks good otherwise from prod requirements.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
atolopko-czi: I'm going with:

>  A large chunk size will result in fewer chunks, resulting in an overall less uniform distribution (unless chunk count is 1); conversely, smaller chunk sizes will result in a larger number of chunks, which will result in an overall more uniform distribution.

This is a statistically interesting problem, in fact, and it's not immediately clear to me how _exactly_ the distribution uniformity changes as function of chunk size, or inversely, chunk count. When either reaches 1, a uniform distribution should result, as a single chunk or single-rowed chunks are both equivalent to a one-step shuffling process, where all the rows are sorted at once. However my LLM-based sources indicate the above is true; but I haven't found any directly-related literature on the matter. I'll have to start running some Monte Carlo simulations...

pablo-gar: I believe we are aligned, see my comments below.

I think speaking of uniformity may confuse the reader. I would suggest a slight change in language

> A large chunk size will result in fewer chunks, resulting in reduced shuffling across all cells in the query; conversely, smaller chunk sizes will result in a larger number of chunks, which will result in an increased shuffling across all cells in the query.

------

I think in isolation your global shuffler without TorchData’s shuffler, would do effectively no shuffling (or very minimal) for training if the “SOMA chunk size” >> “Training batch size”

So the power and complexity comes when mixing your global shuffler with the TorchData’s shuffler. But the amount of true randomness, i.e. how close we get to effectively have shuffling that represents random ordering of all individual cells would be defined by the following variables:

1. Batch modeling size
2. SOMA chunk size.

So fixing all other variables except SOMA chunk size:

1. Global shuffler is enabled
2. TorchData’s shuffler is enabled
3. Batch modeling size is fixed at X value (e.g. 128)

Then the smaller the SOMA chunk size the closer we get to complete random reshuffling of all cells, with the extremes SOMA chunk size of 1 leading to full reshuffled representation and SOMA chunk size of N (where N is the total number of cells) leads the lowest level of shuffling (i.e. only shuffling within batch size)

I think it may good to get our theory straight, and do some simulations if we are still not sure about it

atolopko-czi: In fact, there is no need for the TorchData shuffler now.  ExperimentDataPipe's global/local shuffling ensures that every cell gets shuffled into a new position for the entire query result.  Torch batches are drawn from the fully shuffled result.  The soma_chunk_size only determines how far apart certain cells can end up from one another in the shuffled result. Cells within a given chunk can never be farther than the chunk size.

This might require further discussion in person to make sure we're aligned.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -77,82 +78,106 @@ class Stats:
...
-        self.n_soma_batches += other.n_soma_batches
+        self.n_soma_chunks += other.n_soma_chunks
         return self
 
 
 @contextmanager
 def _open_experiment(
-    uri: str, aws_region: Optional[str] = None, soma_buffer_bytes: Optional[int] = None
+    uri: str,
+    aws_region: Optional[str] = None,
```
bkmartinjr, line 98: side issue: this ties the implementation to AWS (or an AWS-like API). Is there a reason you pass only the `vfs.s3.region` value, rather than simply passing down the entire context config dictionary? In theory, the latter would allow portability to other environments, when/if we decide to host on non-AWS object stores.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
bkmartinjr: I am reasonably certain that TileDB only sorts if you ask it to do so (and by default, it does not - see the `result_order` parameter). I believe what you are observing is that we _write_ the Census in row sorted order, and tiledbsoma is just reading them back in on-disk order (without a sort).

I don't think this distinction matters for your code - just clarifying for shared understanding.

Note: if you do care about order, you can explicitly ask for it using the above param.

api/python/notebooks/experimental/pytorch.ipynb
```diff
@@ -810,9 +632,7 @@
    "cell_type": "markdown",
    "id": "c2fd35fa",
    "metadata": {},
-   "source": [
-    "As we can see, the predictions are poor. It is as left as an exercise to the reader to find a better performing model."
-   ]
+   "source": []
   }
  ],
  "metadata": {
```
bkmartinjr, line 638: there seems to be an empty cell at the end of the notebook

pablo-gar (approved):

api/python/notebooks/experimental/pytorch.ipynb
atolopko-czi: added!

api/python/notebooks/experimental/pytorch.ipynb
atolopko-czi: added new section header

api/python/notebooks/experimental/pytorch.ipynb
atolopko-czi: added section header

api/python/notebooks/experimental/pytorch.ipynb
atolopko-czi: 🔥

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
atolopko-czi: Ah, that makes sense, ty!

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
atolopko-czi: Well, it's a private method, and the region is determined from the `Experiment` passed in by the user (`experiment.context.tiledb_ctx.config().get("vfs.s3.region")`). So I think that sufficiently hides the AWS-specific details, right?

bkmartinjr: My point was that it doesn't support non-AWS storage locations so it isn't very future-proof. Just transparently passing the entire config would be future-proof.

This is _totally_ a side issue, so feel free to ignore!

PR #779, opened 2023-09-21
[python] limit runtime of --expensive unit tests

bkmartinjr:
We have several `--expensive` unit tests, used for acceptance testing upstream packages such as tiledbsoma.  As the Census has grown, their run-times have gotten out of control.  

This PR adds caps on the max size of the queries performed by a handful of the tests, to keep their run time reasonable.  The primary mechanism is to add an obs coordinate range to queries, restricting them to some maximum number of results.


atolopko-czi (approved):
LGTM. 46m run time seems reasonable for --expensive.

ebezzi (approved):
LGTM

PR #783, opened 2023-09-25
[builder] Restore h5ad replication

ebezzi:
Restores h5ad replication to mirrors.


atolopko-czi (approved):
LGTM! You really have to appreciate when changes are this easy. Well-built systems... 😄

PR #784, opened 2023-09-26
[python] GeneformerTokenizer speed optimizations

mlin:
1. Disable the `var_query` for now since we have seen it actually slow things down. Still investigating why. Without it, we get the full expression vector for all genes, and select the ones of interest.
2. Refactor some of the tokenizer computations, equivalent but faster.


pablo-gar (approved):
Makes sense! Plus I don't see a situation where we would need to filter genes out, in principle Geneformer was designed to take all genes available up to 2048.

PR #822, opened 2023-10-23
[python] add PEP 561 py.typed marker

### LINKED ISSUE #806: [python] add PEP 561 typing marker
Add support for type checking code that _uses_ the cellxgene-census python package.

Currently, running mypy on code that _uses_ the cellxgene-census package receives an error from mypy:
> Skipping analyzing "cellxgene_census": module is installed, but missing library stubs or py.typed marker

Minimally, we should add a [PEP 561](https://peps.python.org/pep-0561/) py.typed marker.

bkmartinjr:
Fixes #806 

Add marker file indicating use of line typing for the cellxgene_census package.

### DIFF
--- a/api/python/cellxgene_census/src/cellxgene_census/py.typed
+++ b/api/python/cellxgene_census/src/cellxgene_census/py.typed
@@ -1,1 +1,1 @@
-# Marker file for PEP 561.  The cellxgene_census package uses inline types.
+# Marker file for PEP 561. The cellxgene_census package uses inline types.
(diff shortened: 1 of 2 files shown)


api/python/cellxgene_census/src/cellxgene_census/py.typed
```diff
@@ -0,0 +1 @@
+# Marker file for PEP 561.  The cellxgene_census package uses inline types.
```
ebezzi, line 1:
```suggestion
# Marker file for PEP 561. The cellxgene_census package uses inline types.
```

ebezzi (approved):

api/python/cellxgene_census/src/cellxgene_census/py.typed
bkmartinjr: Lolz, will fix.  This was a literal cut & paste from the mypy marker file.

Issue #831, opened 2023-10-30
[python] dependency check GHA failing due to GeneFormer deps


bkmartinjr:
The last few runs of the Python dependency check GHA have failed with a MacOS install issue.  Ubuntu actions work (if manually started).

At first glance, it looks like an old dependency, `accumulation-tree` is failing, which I believe was installed for GeneFormer work.

An example from https://github.com/chanzuckerberg/cellxgene-census/actions/runs/6687305266

```
Building wheels for collected packages: accumulation-tree, pyudorandom, umap-learn, session-info, pynndescent
  Building wheel for accumulation-tree (setup.py): started
  Building wheel for accumulation-tree (setup.py): finished with status 'error'
  error: subprocess-exited-with-error
  
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [42 lines of output]
      /Users/runner/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/site-packages/setuptools/__init__.py:80: _DeprecatedInstaller: setuptools.installer and fetch_build_eggs are deprecated.
      !!
  Running setup.py clean for accumulation-tree
      
              ********************************************************************************
              Requirements should be satisfied by a PEP 517 installer.
              If you are using pip, you can try `pip install --use-pep517`.
              ********************************************************************************
      
      !!
        dist.fetch_build_eggs(dist.setup_requires)
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/private/var/folders/3s/vfzpb5r51gs6y328rmlgzm7c0000gn/T/pip-install-2rip_z8i/accumulation-tree_60848aaeacb6405f8c531d16c44c44eb/setup.py", line 4, in <module>
          setup(
        File "/Users/runner/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/site-packages/setuptools/__init__.py", line 102, in setup
          _install_setup_requires(attrs)
        File "/Users/runner/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/site-packages/setuptools/__init__.py", line 75, in _install_setup_requires
          _fetch_build_eggs(dist)
        File "/Users/runner/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/site-packages/setuptools/__init__.py", line 80, in _fetch_build_eggs
          dist.fetch_build_eggs(dist.setup_requires)
[...]

PR #833, opened 2023-10-30
fix installation of geneformer's transitive dependency in py-dependency-check.yml

mlin:
RE: #831 

Ports a workaround for installing a gnarly transitive dependency of Geneformer, [from py-unittests.yml](https://github.com/chanzuckerberg/cellxgene-census/blob/1d071d6e189a32c3a040e67a43e981ba7f082701/.github/workflows/py-unittests.yml#L28-L29) to py-dependency-check.yml.

Due to this dependency that's difficult on certain platforms (cython etc.), we leave Geneformer out of the pyproject.toml dependencies and instead raise [a runtime error](https://github.com/chanzuckerberg/cellxgene-census/blob/1d071d6e189a32c3a040e67a43e981ba7f082701/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/huggingface/geneformer_tokenizer.py#L104-L111) asking end users to install it when needed. However, we do need to install it in GHA in order to run the unit tests (which albeit are temporarily disabled for unrelated reasons #818).

### DIFF
--- a/.github/workflows/py-dependency-check.yml
+++ b/.github/workflows/py-dependency-check.yml
@@ -53,6 +53,7 @@ jobs:
       - name: Install dependencies (including experimental)
         run: |
           python -m pip install -U pip setuptools wheel
+          pip install --use-pep517 accumulation-tree # Geneformer dependency needs --use-pep517 for Cython
           pip install -U -r ./api/python/cellxgene_census/scripts/requirements-dev.txt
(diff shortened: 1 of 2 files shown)


mlin:
Pending: https://github.com/chanzuckerberg/cellxgene-census/actions/runs/6697238492

pablo-gar (approved):

PR #853, opened 2023-11-08
Contrib embeddings

bkmartinjr:
Ingestion and data model for contributed embeddings.  PR includes:
* ingestion tool (`census_contrib` package) to ingest, transform and save embedding + metadata. Also supports injecting single embeddings into a Census build where appropriate
* notebook for functional access demo
* Python cellxgene-census package extension `cellxgene_census.experimental.get_embedding()` to make reading simpler

The only open issue at the moment is the dependency on a TileDB-SOMA bug fix coming in the next release.

Reviewer requests (happy to take input on anything, but at least cover these):
* Sidney, Alec, Pablo: notebook, get_embedding API
* Emanuele: implementation


api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding.py
```diff
@@ -0,0 +1,113 @@
...
+import tiledbsoma as soma
+
+from .._open import _build_soma_tiledb_context, open_soma
+from .._release_directory import get_census_version_directory
+
+
+def get_embedding(
+    census_version: str,
+    embedding_uri: str,
+    obs_soma_joinids: Union[npt.NDArray[np.int64], pa.Array],
```
atarashansky, line 27: Any plans to add support for passing in value filters and/or tiledbsoma.AxisQuery objects?

atarashansky: The current logic in the notebook seems quite circular.

 - First, you need to figure out what census version to use, so you load the embedding metadata.
 - Then, you open write out the verbose, nested context managers to get the query and soma join IDs.
 - Finally, you get the slice of the embedding, passing in the census version you retrieved from the embedding metadata.

Why are the first two steps necessary? It seems much cleaner to require only the embedding URI and the AxisQuery. The census version can be inferred from the metadata and the right version of census can be opened to fetch the soma join ids if a value filter was provided in AxisQuery.

bkmartinjr: You are raising multiple issues and I think maybe conflating them.

1. the notebook logic - I get the census version from the embedding, as I choose to make the embedding the only input to the notebook. We could make the user specify both, but i"m not sure why that is better.  Keep in mind that they must be consistent or the results are garbage.

2. the API design - it is a requirement that the sugar API check that the user has not accidentally used an embedding and census which are mismatched.  CC: @pablo-gar as this is his design.

Ultimately, to "get" info, we need:
* an embedding URI
* a census URI
* and they must match (and it is a requirement that we check that they match)

I personally prefer a low-level API, which is more flexible for a variety of use cases. This one provides that by accepting an array of joinids, rather than some higher level abstraction such as an OBS dataframe filter.

We could do both.  Easy peasy to implement.

Alec - can you shortcut this conversation and propose a concrete API signature that you would like to see?

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding.py
```diff
@@ -0,0 +1,113 @@
...
+        obs_soma_joinids:
+            The slice of the embedding to fetch and return.
+        context:
+            A custom :class:`SOMATileDBContext` which will be used to open the SOMA object.
+            Optional, defaults to None.
+
+    Returns:
+        A :class:`numpy.ndarray` containing the embeddings. Embeddings are positionally
+        indexed by the obs_soma_joinids. In other words, the cell identified by
+        `obs_soma_joinids[i]` corresponds to the `ith` positoin in the returned ndarray.
```
ebezzi, line 49:
```suggestion
        `obs_soma_joinids[i]` corresponds to the `ith` position in the returned ndarray.
```

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding.py
```diff
@@ -0,0 +1,113 @@
...
+        with open_soma(census_version=census_version, context=context) as census:
+            experiment_name = embedding_metadata["experiment_name"]
+            if experiment_name not in census["census_data"]:
+                raise ValueError("Census and embedding mismatch - experiment_name does not exist")
+            measurement_name = embedding_metadata["measurement_name"]
+            if measurement_name not in census["census_data"][experiment_name].ms:
+                raise ValueError("Census and embedding mismatch - measurement_name does not exist")
+
+        embedding_shape = (len(obs_soma_joinids), E.shape[1])
+        embedding = np.full(embedding_shape, np.NaN, dtype=np.float32, order="C")
```
ebezzi, line 102: Based on some experiments I did while implementing the `to_anndata` in SOMA, embeddings have some values that are exactly zero (_very_ few - I guess those are very small values that underflow). This will cause them to be NaN instead, which I believe it's not what we want.

atarashansky: The embeddings are stored in sparse format, so zeros correspond to missing data. Some models were not trained on certain cells. For example, most models were only trained on primary data.

bkmartinjr: Actually, this will cause them to be zero, not NaN, as we _explicitly store zero values for all embeddings.  The only missing values are "no embedding", not "zero-valued embedding".

See the spec [here](https://github.com/chanzuckerberg/cellxgene-census/blob/bkmartinjr/contrib-embeddings/tools/census_contrib/embedding_metadata.md), in particular, the paragraph labelled *important*

ebezzi: Yeah, I realized that my experiments were run against an initial prototype where the zeros were not explicitly marked. Good to know that the final version won't have this issue.

pablo-gar (approved):
API signatures and readmes look good. I defer to @ebezzi for code review.

As I mentioned I will do some editorial changes to the notebook, but in order to avoid blocking the release I will do so as a PR. 

Can you please remove the symlink to the notebook for now? Otherwise the notebook will become public upon our upcoming release, which will trigger the website rendering.

bkmartinjr:
> please remove the symlink to the notebook 

@pablo-gar  - this is done

mlin (approved):
I reviewed the notebook and _embedding.py, discussed one or two things IRL, LGTM

PR #855, opened 2023-11-08
Acceptance/Builder tests misc fixes

ebezzi:
1. Acceptance tests now use unsigned requests, which is required if you want to run them on an EC2 machine without explicit IAM permissions
2. Fixes an issue where matrices generated for builder tests fixtures sometimes contain rows with all zeros, causing the presence matrix to have a different nnz count

### DIFF
--- a/api/python/cellxgene_census/tests/test_acceptance.py
+++ b/api/python/cellxgene_census/tests/test_acceptance.py
@@ -26,6 +26,7 @@ def make_context(census_version: str, config: Optional[Dict[str, Any]] = None) -
     version = cellxgene_census.get_census_version_description(census_version)
     s3_region = version["soma"].get("s3_region", "us-west-2")
     config.update({"vfs.s3.region": s3_region})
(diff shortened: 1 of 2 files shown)


api/python/cellxgene_census/tests/test_acceptance.py
```diff
@@ -26,6 +26,7 @@ def make_context(census_version: str, config: Optional[Dict[str, Any]] = None) -
     version = cellxgene_census.get_census_version_description(census_version)
     s3_region = version["soma"].get("s3_region", "us-west-2")
     config.update({"vfs.s3.region": s3_region})
+    config.update({"vfs.s3.no_sign_request": "true"})
     return soma.options.SOMATileDBContext(tiledb_ctx=tiledb.Ctx(config))
```
bkmartinjr, line 30: slightly off-topic:  this code would be better written:
> return soma.options.SOMATileDBContext(tiledb_config=config)

I.e., remove the `tiledb` dependency.

PR #863, opened 2023-11-20
[builder] Absolute uris should use the default mirror

ebezzi:
Mirror resolution is done via the `mirrors.json` file on newer versions of the census API, but for legacy reasons we're still supporting absolute URIs. These don't support mirrors and should just point to the default one.

Also note that the h5ad functions also do not support mirroring, and still read from the absolute uris. This was done on purpose due to the original intent of deprecating those h5ads locations, but we should probably update them now. I can submit a separate PR for it.


atolopko-czi (approved):
LGTM.

Are we also planning to (manually) update existing absolute URIs in release.json for LTS data releases?

ebezzi:
> LGTM.
> 
> Are we also planning to (manually) update existing absolute URIs in release.json for LTS data releases?

Yes, we're gonna do the first update manually. This is for next week's run.

bkmartinjr (approved):

Issue #832, opened 2023-10-30
deprecate and remove X_sparse_iter when tiledbsoma blockwise iterators are available


bkmartinjr:
When SOMA support for blockwise iterators are released:

- [x] deprecate cellxgene_census.experimental.util.X_sparse_iter
- [x] port all uses to blockwise iterators:
    - [x] geneformer/mul
    - [x] find_dups notebook
    - [x] unit tests
    - [x] pca notebook
- [x] remove from the cellxgene_census package

Blockwise iterators will ship in TileDB-SOMA 1.5, so this needs to be deprecated in cellxgene-census version 1.8 and removed in 1.9

mlin:
I have to update the geneformer tokenizer.

PR #866, opened 2023-11-22
[python] migrate GeneformerTokenizer to blockwise iterator

mlin:
#832


bkmartinjr (approved):
visual inspection only, but LGTM

PR #882, opened 2023-12-05
[builder] Fix builder release manifest validation to not check for absolute uri

ebezzi:
Fixes an issue introduced by the switch to the new bucket in the `uri` field. To keep the code simple, I removed the explicit `uri` check in the manifest validation - I think this is acceptable since that is now a legacy field. Alternatively, we can pass the "default mirror bucket" all the way to that function, but that will make the code more complex. @bkmartinjr let me know what you think.


bkmartinjr (approved):
I'm fine removing the check for a legacy field

PR #885, opened 2023-12-06
[python] upgrade to tiledbsoma and add obsm to get_anndata

### LINKED ISSUE #874: Add `obsm_layers` parameter to `get_anndata`
With the following signature

`obsm_layers: Sequence[str] = ()`

And the following doc string

```
obsm_layers:
  Additional obsm layers to read and return in the ``obsm`` slot.
```

ebezzi:
Resolves #874 

Note: currently there are no unit tests that do not rely on the live corpus for data layers. Since `obsm` isn't yet available in any live Census distributions, the unit tests are marked as skipped.


bkmartinjr (approved):
LGTM.  At some point we need to add varm, obsp, varp, but those are not urgent.

PR #888, opened 2023-12-08
[builder] Update release cleanup code to support relative URIs

ebezzi:
These assert do not work anymore because the absolute uri doesn't match the uri of the primary bucket anymore. I don't believe other assertions are necessary here. We could force a check that the URI matches a hardcoded pattern if we want to keep a defensive assert, but it will break again if we (e.g.) decide to change the location of the primary bucket.


atolopko-czi (approved):
This seems fine since (absolute) `uri` is deprecated anyway.

bkmartinjr (approved):

PR #895, opened 2023-12-15
Make default soma context available for customization

atolopko-czi:
- Add get_default_soma_context() method to obtain a `TileDBSOMAContext` with reasonable defaults for opening the Census. This allows user to customize the context without having to also specify the default settings that are usually always needed.
- Changed the default context to set `vfs.s3.no_sign_request` to `true`, as this is the more frequent case. This avoids users having to set it explicitly when customizing other soma settings in a default soma context.


api/python/cellxgene_census/tests/test_open.py
```diff
@@ -38,30 +43,54 @@ def test_open_soma_latest() -> None:
         assert isinstance(census, soma.Collection)
 
 
-@pytest.mark.live_corpus
-def test_open_soma_with_context() -> None:
```
atolopko-czi, line 42: refactored this existing test into 2 tests, and added a third for the default soma context

api/python/cellxgene_census/tests/conftest.py
```diff
@@ -49,7 +49,6 @@ def small_mem_context() -> soma.SOMATileDBContext:
     cfg = {
         "tiledb_config": {
             "soma.init_buffer_bytes": 32 * 1024**2,
-            "vfs.s3.no_sign_request": True,
         },
     }
     return soma.SOMATileDBContext().replace(**cfg)
```
ebezzi, line 54:
```suggestion
    return get_default_soma_context().replace(**cfg)
```

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -68,38 +71,52 @@ def _open_soma(
...
+
+    Examples:
+
+        To reduce the amount of memory used by TileDB-SOMA I/O operations:
+
+        .. highlight:: python
+        .. code-block:: python
+
+            ctx = cellxgene_census.get_default_soma_context().replace(
+                tiledb_config={"py.init_buffer_bytes": 128 * 1024**2,
```
bkmartinjr, line 101: why not just allow the user to pass a `config` to the `open_soma()` call? That makes it closer to the tiledbsoma syntax, and I think easier to use.

Example:
```
census = cellxgene_census.open_soma(
    uri="s3://my-private-bucket/census/soma",
    tiledb_config={"vfs.s3.region": "us-east-1"}
)
```

api/python/cellxgene_census/tests/conftest.py
bkmartinjr, line 54: this pattern is obsolete - should be:

> soma.SOMATileDBContext(tiledb_config=cfg)

which avoids the create/delete/create waste

bkmartinjr (changes requested):
I think the forced use of `.replace()` is no bueno, especially given that we fixed this issue upstream in tiledbsoma.

Recommend just adding a `tiledb_config` argument to the function signature, mimicking the signature for SOMATileDBContext()

api/python/cellxgene_census/src/cellxgene_census/_open.py
atolopko-czi: `TileDBObject.open()` calls still just take a `context`, afaict (https://github.com/single-cell-data/TileDB-SOMA/blob/24b57de4fad1828d4fdcda99ca630bf048b22cd6/apis/python/src/tiledbsoma/_tiledb_object.py#L44).  But the `SOMATileDBContext` object does now accept a config dict.  So for your suggestion, would it make sense to accept a `tiledb_config` in the `get_default_soma_context()` call instead? My thinking is that if the design of `SOMATileDBContext` changes, the `open_soma()` method doesn't have to change, though the get_default_soma_context() would.  But since it's more "context savvy", maybe that's a little better?

atolopko-czi: The other consideration is that `get_default_soma_context()` is also useful when opening `TileDBObject`s directly, and wanting to use the Census's context defaults. So if we eliminated the method in favor of _only_ supporting context overrides via `open_soma()`, this use case has to resort back to specifying _all_ the context settings correctly.

bkmartinjr: I'm not suggesting one over the other, but doing both.  Common case is `open_soma` while overriding one param such as memory.  But you are right that the `get context` is also useful.

So, I'd do three things:
1. add a `tiledb_config={}` argument to the `open_soma` call.
2. retain the newly added get_default_soma_context
3. add the same `tiledb_confg` argument to get_default_soma_context

I'm primarily suggesting that the dict-override is superior, and should be pulled through our API

atolopko-czi: As discussed separately, will merge this and implement this suggestion on a new PR.

bkmartinjr (approved):

PR #896, opened 2023-12-15
[builder] port to use enums in schema

bkmartinjr:
*WORK IN PROGRESS*

Initial port to use Arrow dictionary / TileDB enums for all DataFrame columns where it is useful.


tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/source_assets.py
```diff
@@ -63,7 +63,11 @@ def _copy_file(n: int, dataset: Dataset, asset_dir: str, N: int) -> str:
         raise last_error
 
     # verify file size is as expected, if we know the size a priori
-    assert (dataset.asset_h5ad_filesize == -1) or (dataset.asset_h5ad_filesize == os.path.getsize(dataset_path))
+    assert (dataset.asset_h5ad_filesize == -1) or (
```
bkmartinjr, line 66: added better message for assert.  Change is unrelated to this PR - tripped over this during my testing, which overlapped with the schema four migration, which in turn triggered this assert to fail.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/schema_util.py
```diff
@@ -0,0 +1,148 @@
...
+
+2. allow explicit type choices where there are options (e.g., string vs large_string)
+
+"""
+
+OptDataFrame = TypeVar("OptDataFrame", pd.DataFrame, None, Optional[pd.DataFrame])
+
+
+@attrs.define(frozen=True, kw_only=True, slots=True)
+class FieldSpec:
```
ebezzi, line 24: Even if this is not intended as a public class, adding some docstrings could help better understand some of the methods (and especially parameters) to first-time readers.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/summary_cell_counts.py
```diff
@@ -94,6 +95,7 @@ def accumulate_summary_counts(current: pd.DataFrame, obs_df: pd.DataFrame) -> pd
                 columns="is_primary_data",
                 index=["organism", "ontology_term_id", "label"],
                 fill_value=0,
+                aggfunc="sum",
```
ebezzi, line 98: Curious why this wasn't specified before?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/validate_soma.py
```diff
@@ -192,6 +192,12 @@ def _validate_axis_dataframes(args: Tuple[str, str, Dataset, List[ExperimentSpec
                 .reset_index(drop=True)
             )
 
+            # decategorize census obs slice, as it will not have the same categories as H5AD obs,
+            # preventing Pandas from performing the DataFrame equivalance operation.
```
ebezzi, line 196:
```suggestion
            # preventing Pandas from performing the DataFrame equivalence operation.
```

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/summary_cell_counts.py
bkmartinjr: the aggfunc is a noop no matter what it is specified as, because each value is unique (i.e., no aggregation occurs). However, the default aggfunc ('mean') causes the int values to be cast to a float. As the dataframe column is defined as int, this is no good. The sum aggfunc is specified because Python sum will leave ints as ints, bypassing the cast.

This was a latent bug I only found when the new TableSpec/FieldSpec class actually checked that the resulting schema was as expected.  Hence the fix.

bkmartinjr: and I'll add a comment to this effect

ebezzi: Thanks! This explains it.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/schema_util.py
bkmartinjr: done

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/schema_util.py
```diff
@@ -0,0 +1,176 @@
...
+                else:
+                    index_type = pa.int8()  # smallest, as a default
+                pa_fields.append((field_spec.name, pa.dictionary(index_type, field_spec.type, ordered=0)))
+            else:
+                pa_fields.append((field_spec.name, field_spec.type))
+
+        return pa.schema(pa_fields)
+
+    def field_names(self) -> Sequence[str]:
+        """Return field names for this TableSpec as a seuqence of string"""
```
atolopko-czi, line 152:
```suggestion
        """Return field names for this TableSpec as a sequence of string"""
```

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/globals.py
```diff
@@ -58,126 +72,118 @@
...
-  Columns with non-repetative values (e.g., var.feature_id) are NOT dictionary coded.
-* obs/var DataFrame show significant improvement in on-disk size and read performance at high Zstd compression level,
-  making it worth the extra build/write compute time. Level set to highest value that does not require additional
-  resources at decompression (nb. 20+ requres additional memory).
-* int64 index columns (soma_joinid, soma_dim_0, etc) empirically show wins from ByteShuffle followed by Zstd.
-  First dimension of axis dataframes are always monotonically increasing, and also beneit from DoubleDelta.
-* Benchmarking X slicing (using lung demo notebook) used to tune X[raw]. Read / query performance did not benefit from
-  higher Zstd compression beyond level=5, so the level was not increased further (and level=5 is still reasonable for
-  writes)
-"""
```
atolopko-czi, line 117: I appreciated these notes, but also understand if they're now sufficiently outdated to not bother keeping (and updating).

atolopko-czi (approved):
LGTM. I like the `{Field,Table}Spec` design. One minor concern is that there's a non-trivial amount of code involved in generating the arrow table schemas. Is it worth having a basic test or two to verify the translation of a TableSpec to a schema? Or do we have enough checks in the Census validator?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/schema_util.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+        if not ignore_dict_type and self.is_dictionary and not pa.types.is_dictionary(self.type):
+            raise TypeError("Unable to determine final Pandas type for dictionary.")
+        return cast(npt.DTypeLike, self.type.to_pandas_dtype())
+
+    def is_type_equivalent(self, other_type: pa.DataType, *, null_non_primitive_equivalence: bool = False) -> bool:
+        """
+        Return True if this FieldSpec is equivalent to the Arrow `other_type`.
+        For convenience in comparing with types inferred from Pandas DataFrames,
+        where strings and other Arrow non-primtives are stored as objects, allow a
```
ebezzi, line 50:
```suggestion
        where strings and other Arrow non-primitives are stored as objects, allow a
```

ebezzi (approved):

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/schema_util.py
bkmartinjr: fixed

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/schema_util.py
bkmartinjr: fixed

bkmartinjr:
> worth having a basic test or two 

good call - added (and fixed a bug immediately revealed by said tests!)

PR #897, opened 2023-12-16
[misc] acceptance test GHA workflow

bkmartinjr:
Add GHA which peforms the R and Python "acceptance" tests on a schedule - currently every Saturday at 1AM UTC. Can also be run manually via workflow_dispatch.

By default, will install from `main` and run the latest acceptance tests in `main`. You can run it against a branch with:

> gh workflow run full-unittests.yml --ref _branch_name_

The python (not R) job supports installing a custom `tiledbsoma` version, allowing the `main` of this repo to test against any tiledbsoma version, including branches from that repo. This will allow pre-qualification of tiledbsoma releases against the latest cellxgene-census code base. For example, to test against the head of main, do:

> gh workflow run full-unittests.yml -f 'tiledbsoma_python_dependency=git+https://github.com/single-cell-data/TileDB-SOMA.git#egg=tiledbsoma&subdirectory=apis/python/'

Other changes:
* update to R acceptance test to reduce memory use
* minor fix to version dep on setuptools for python package
* minor fix to Python tests to work with latest LTS Census release

Current test run status:
* Python - works
* R - generates warnings about long vectors (32 bit limitation of R)

Overhang:
* single-cell-data/TileDB-SOMA#1937 contains a fix that will allow removal of the `editable` flag on the tiledbsoma version override when fixed. Filed as #907


api/python/cellxgene_census/pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=45", "setuptools_scm[toml]>=6.2"]
+requires = ["setuptools>=64", "setuptools_scm[toml]>=8"]
```
bkmartinjr, line 2: This is only tangentially related to the new GHA, but important as the existing dependencies were quite obsolete and willl fail in some scenarios.  See https://setuptools-scm.readthedocs.io/en/latest/

.github/workflows/full-unittests.yml
```diff
@@ -0,0 +1,91 @@
...
+          lscpu
+
+      - name: install OS dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y build-essential git-all libxml2-dev libssl-dev libcurl4-openssl-dev cmake
+
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
atolopko-czi, line 33:
```suggestion
          fetch-depth: 1
```

atolopko-czi: for just the head commit

.github/workflows/full-unittests.yml
```diff
@@ -0,0 +1,91 @@
...
```
atolopko-czi, line 74:
```suggestion
          fetch-depth: 1
```

.github/workflows/full-unittests.yml
```diff
@@ -0,0 +1,91 @@
+name: cellxgene_census package full unit tests
+
+# Run all unit tets, including those that are too expensive to run frequently.
```
atolopko-czi, line 3:
```suggestion
# Run all unit tests, including those that are too expensive to run frequently.
```

.github/workflows/full-unittests.yml
```diff
@@ -0,0 +1,91 @@
...
+# This workflow requires a very large capacity runner, e.g., 1+TiB RAM. There
```
atolopko-czi, line 4:
```suggestion
# This workflow requires a very large capacity, self-hosted runner, e.g., 1+TiB RAM. There
```
(or mention in some comment that this depends on a self-hosted runner)

atolopko-czi (approved):
LGTM, w/some very minor suggestions.

.github/workflows/full-unittests.yml
bkmartinjr: we don't actually care if it is self-hosted or not, as it is a generic ubuntu image.  Just care if it has large resources

.github/workflows/full-unittests.yml
```diff
@@ -0,0 +1,91 @@
...
+          working-directory: ./api/r/cellxgene.census
+          extra-packages: any::rcmdcheck, any::remotes
+          cache: true
+
+      - name: testthat
+        run: |
+          Rscript -e 'remotes::install_local("./api/r/cellxgene.census")'
+          Rscript -e 'library("tiledbsoma"); tiledbsoma::show_package_versions()'
+          Rscript -e 'library("testthat"); library("cellxgene.census"); test_dir("./api/r/cellxgene.census/tests/")'
+          Rscript -e 'library("cellxgene.census"); library(testthat); test_file("./api/r/cellxgene.census/tests/testthat/acceptance-tests.R")'
```
ebezzi, line 88: Pablo created an `acceptance-tests-run-script.R` that does some extra output (for instance adds timestamps, etc). I don't know if this is redundant with GHA, but if might be worth considering.

bkmartinjr: I know about it, but I was unable to get it to work in this context.  Not sure why - seems to be some sort of R sensitivity to current working dir or something like that - so I just cut & pasted the contents of it into this workflow

ebezzi: Yeah, I think it needs to be run from the `testthat` directory. I wouldn't worry - we can adjust them later if we want a more detailed output.

.github/workflows/full-unittests.yml
```diff
@@ -0,0 +1,107 @@
...
+          echo '--------- tiledbsoma.show_package_version():'
+          python -c 'import tiledbsoma; tiledbsoma.show_package_versions()'
+          echo '--------- PIP package versions:'
+          pip freeze
+
+          PYTHONPATH=. pytest -v --durations=0 -rP --experimental --expensive ./api/python/cellxgene_census/tests/
+
+  r_unit_tests:
+    if: false # TEMPORARILY DISABLE FOR DEBUGGING
+    runs-on: single-cell-1tb-runner
```
jakeyheath, line 73: woot! just wanted to say 👋🏻 and introduce myself. I worked with @beroy getting these self-hosted runners set up. I'm watching them periodically and recording metrics on the nodes and autoscaler, but ping me if you see any unusual behavior or have other requirements you want us to add to them. Have fun!

bkmartinjr: hi @jakeyheath - these are a game changer.  Really happy to have them!

So far they seem to work great, other than missing some "nice-to-have" packages installed in their default image.  I have given @beroy an initial list.

.github/workflows/full-unittests.yml
bkmartinjr: the other related issue is capturing timings. The python job will print out timings, but R will not. As you said, happy to refine this later.

.github/workflows/full-unittests.yml
jakeyheath: Easy to solve, we already chatted this morning about building a custom container for this project to use that has all the packages pre-installed. 

Let's meet in the new year and see how things are going. Right now, everything is running in our czi-playground account. Might be good to put this project is a more specific location and tune the performance as needed. Here's a screenshot of the metrics I collected on the latest run. Happy to show yall how to pull these yourselves and tune to your needs.


Screenshot 2023-12-18 at 12 15 17 PM

api/python/cellxgene_census/tests/experimental/pp/test_hvg.py
```diff
@@ -209,7 +209,7 @@ def test_hvg_vs_scanpy(
             "Homo sapiens",
             "is_primary_data == True",
             "dataset_id",
-            slice(500_000, 1_000_000),
+            slice(1_000_000, 4_000_000),
```
bkmartinjr, line 212: the latest LTS census has no primary cells in the [500_000,1_000_000] soma joniid range, so this test will fail.  Expanded the range to ensure we pick up some cells.

api/r/cellxgene.census/tests/testthat/acceptance-tests.R
```diff
@@ -102,7 +102,7 @@ test_that("test_incremental_read_X_human", {
 })
 
 test_that("test_incremental_read_X_human-large-buffer-size", {
-  census <- open_soma_latest_for_test(soma.init_buffer_bytes = paste(4 * 1024**3))
+  census <- open_soma_latest_for_test(soma.init_buffer_bytes = paste(1 * 1024**3))
```
bkmartinjr, line 105: on very high cpu count machines, this will OOM due to per-core buffer allocation.  Reducing to a GiB resolves the issue

.github/workflows/full-unittests.yml
ebezzi: I can take a stab at it later. For now, resolving this.

ebezzi (approved):
LGTM. Let's ship those now and monitor how they go.

PR #902, opened 2023-12-18
[python] Syntactic sugar for specifying tiledb_config options

atolopko-czi:
Follow-on work for #895 

The `open_soma()` and `get_default_soma_context()` methods now take an optional `tiledb_config` dict param, to simplify caller code for the common case of overriding selected `tiledb_config` options in the default SOMA context.


bkmartinjr (approved):

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -74,18 +74,24 @@ def _open_soma(
...
     """Return a ``SOMATileDBContext`` with sensible defaults that can be further customized by the user.  The customized
-    context can then be passed to ``open_soma()`` or a ``SOMAObject.open()`` method,
+    context can then be passed to ``open_soma(context=...)`` or a ``SOMAObject.open(context=...)`` method,
     such as ``tiledbsoma.Experiment.open()``.  Use the ``replace()`` method on the returned object to customize its
-    settings.
+    settings further.
+
+    Args:
+        tiledb_config:
+            A dictionary of TileDB configuration parameters.  If specified, the specified parameters will override the
```
ebezzi, line 93:
```suggestion
            A dictionary of TileDB configuration parameters.  If specified, the parameters will override the
```

ebezzi (approved):

api/python/cellxgene_census/src/cellxgene_census/_open.py
bkmartinjr: I'll fix this and then land the PR

Issue #899, opened 2023-12-17
[python] unit test `test_open_soma_with_context()` failing


bkmartinjr:
Running the full Python acceptance tests has one failure using the latest tests from `main`:

```
_________________________ test_open_soma_with_context __________________________

    @pytest.mark.live_corpus
    def test_open_soma_with_context() -> None:
        description = cellxgene_census.get_census_version_description("latest")
        uri = description["soma"]["uri"]
        s3_region = description["soma"].get("s3_region")
        assert s3_region == "us-west-2"
    
        # Verify the default region is set correctly in the TileDB context object.
>       with cellxgene_census.open_soma(census_version="latest", context=soma.SOMATileDBContext()) as census:

api/python/cellxgene_census/tests/test_open.py:49: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../_tool/Python/3.11.7/x64/lib/python3.11/site-packages/cellxgene_census/_open.py:211: in open_soma
    return _open_soma(locator, context)
../../_tool/Python/3.11.7/x64/lib/python3.11/site-packages/cellxgene_census/_open.py:76: in _open_soma
    return soma.open(locator["uri"], mode="r", soma_type=soma.Collection, context=context)
../../_tool/Python/3.11.7/x64/lib/python3.11/site-packages/tiledbsoma/_factory.py:114: in open
    obj = _open_internal(_tdb_handles.open, uri, mode, context, tiledb_timestamp)
../../_tool/Python/3.11.7/x64/lib/python3.11/site-packages/tiledbsoma/_factory.py:143: in _open_internal
    handle = opener(uri, mode, context, timestamp)
../../_tool/Python/3.11.7/x64/lib/python3.11/site-packages/tiledbsoma/_tdb_handles.py:48: in open
    obj_type = tiledb.object_type(uri, ctx=context.tiledb_ctx)
tiledb/libtiledb.pyx:3506: in tiledb.libtiledb.object_type
    ???
tiledb/libtiledb.pyx:356: in tiledb.libtiledb.check_error
    ???
tiledb/libtiledb.pyx:350: in tiledb.libtiledb._raise_ctx_err
    ???
[...]

PR #905, opened 2023-12-19
fix tests with invalid vfs.s3.no_sign_request setting

atolopko-czi:
- reinstate `vfs.s3.no_sign_request=True` setting to allow these tests to pass on EC2 instances that have an IAM role and that does not grant explicit access to the bucket
- remove one test, `test_open_soma_with_plain_soma_context()` that can never be successfully run under these circumstances; there is still sufficient coverage from other tests
- tweak test assertions

Tests passing on GHA standard runner and on EC2 instance with IAM role, where tests previously failed.  Relevant tests pass on GHA custom runner: test run [here](https://github.com/chanzuckerberg/cellxgene-census/actions/runs/7266987244)

### DIFF
--- a/api/python/cellxgene_census/tests/test_open.py
+++ b/api/python/cellxgene_census/tests/test_open.py
@@ -35,24 +35,22 @@ def test_open_soma_stable() -> None:
             assert census.context.tiledb_ctx.config()[k] == str(v)
 
 
-@pytest.mark.live_corpus
-def test_open_soma_latest() -> None:
-    # There should _always_ be a 'latest'
-    with cellxgene_census.open_soma(census_version="latest") as census:
-        assert census is not None
-        assert isinstance(census, soma.Collection)
-
-
 @pytest.fixture(scope="module")
 def latest_locator() -> CensusLocator:
     return cellxgene_census.get_census_version_description("latest")["soma"]
 
 
 @pytest.mark.live_corpus
-def test_open_soma_with_plain_soma_context(latest_locator: CensusLocator) -> None:
-    with cellxgene_census.open_soma(census_version="latest", context=soma.SOMATileDBContext()) as census:
-        # Verify the default region is set correctly in the TileDB context object.
-        assert census.context.tiledb_ctx.config()["vfs.s3.region"] == latest_locator.get("s3_region")
+def test_open_soma_latest(latest_locator: CensusLocator) -> None:
+    with cellxgene_census.open_soma(census_version="latest") as census:
+        # There should _always_ be a 'latest'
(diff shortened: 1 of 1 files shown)


api/python/cellxgene_census/tests/test_open.py
```diff
@@ -63,34 +61,36 @@ def test_open_soma_with_customized_plain_soma_context(latest_locator: CensusLoca
...
     with cellxgene_census.open_soma(census_version="latest", context=custom_context) as census:
-        # Verify the soma context defaults are set correctly in the TileDB context object.
-        assert census.context.tiledb_ctx.config()["vfs.s3.no_sign_request"] == "false"
+        # Verify the non-overriden soma context defaults are set correctly in the TileDB context object.
+        assert census.context.tiledb_ctx.config()["vfs.s3.no_sign_request"] == "true"
         assert census.context.tiledb_ctx.config()["vfs.s3.region"] == latest_locator.get("s3_region")
+        assert census.context.tiledb_ctx.config()["py.init_buffer_bytes"] == f"{1 * 1024 ** 3}"
 
-        # Verify that the user-provided custom config is set correctly in the TileDB context object.
+        # Verify that the user-overriden config settings are set correctly in the TileDB context object.
```
bkmartinjr, line 93: typo in overriden (missing 'd')

bkmartinjr (approved):

api/python/cellxgene_census/tests/test_open.py
atolopko-czi: ty fixed

ebezzi (approved):
LGTM. @atolopko-czi, with this new paradigm, I would recommend to change `test_opening_census_without_anon_access_fails_with_bogus_creds` to use a context where the unsigned request is disabled, instead of using `context=soma.SOMATileDBContext()`.

atolopko-czi:
> LGTM. @atolopko-czi, with this new paradigm, I would recommend to change `test_opening_census_without_anon_access_fails_with_bogus_creds` to use a context where the unsigned request is disabled, instead of using `context=soma.SOMATileDBContext()`.

It's a good idea, but it's hard to make that a reliable test, since it depends upon the test environment and whether there is an IAM role configured. See discussion [here](https://github.com/chanzuckerberg/cellxgene-census/issues/899#issuecomment-1863163031), as it would take implementing one of Bruce's suggestions there. It could be worth a separate issue, I suppose. But I think it's more than we should implement on this particular PR.

bkmartinjr:
>  a separate issue

definitely agree that we should land this PR and unblock the acceptance tests, then fine tune

PR #922, opened 2024-01-04
remove OS dependencies from Action workflow

bkmartinjr:
PR removes the OS package installs that are already installed in the self-hosted docker image.


ebezzi (approved):

beroy (approved):
It must be now synced with the latest runner [images](https://github.com/chanzuckerberg/cell_census_github_runners/commit/6bd4587a51c622804cc4a18771c02056ae1252d4).

PR #923, opened 2024-01-05
[builder] delete bounding box metadata on sparse arrays

### LINKED ISSUE #912: [builder] sparse array bounding box metadata - delete after build
Re: single-cell-data/TileDB-SOMA#1969

The Census build definitely fall prey to this issue.  After discussions with @mojaveazure and @johnkerl, it appears best if we just delete the bounding box after the Census build. TileDB-SOMA implementation should then fall back to either shape or nonempty_domain, both of which will work fine for the Census.

bkmartinjr:
Fixes #912 

Remove the bounding box metadata from all SparseNDArray objects in the Census, as a work-around for single-cell-data/TileDB-SOMA#1969


atolopko-czi (approved):
LGTM

PR #924, opened 2024-01-05
[misc] move ubuntu runner to self-hosted for py-unittest workflow

bkmartinjr:
Move python unit test workflow move from `ubuntu-latest` runner to self-hosted `single-cell-8c64g-runner` runner

Other minor changes:
* removed a not-very-useful test that made bad assumptions about the AWS credentials on the running.
* updated to latest upload/download action


.github/workflows/py-unittests.yml
```diff
@@ -63,28 +70,30 @@ jobs:
...
+          name: coverage-builder-${{ matrix.os }}-${{ matrix.python-version }}
           path: ./.coverage*
           retention-days: 3
 
   submit-codecoverage:
     needs:
       - unit_tests_builder
       - unit_tests_python_api
-    runs-on: ubuntu-20.04
+    runs-on: [self-hosted, Linux]
```
jakeyheath, line 83: this one is okay to run on arm or x86 right? otherwise, we should pick one architecture

jakeyheath (approved):
Minor question, but woot!

.github/workflows/py-unittests.yml
bkmartinjr: this step is pure python, so it doesn't care....

bkmartinjr: I spoke too soon - it failed!

The version '3.10' with architecture 'arm64' was not found for Ubuntu 20.04. The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json

will add the x86 tag....

jakeyheath: It is wild how many things still don't work on ARM, even with languages like python. I've had to do similar things for a few of our github actions

ebezzi (approved):

PR #925, opened 2024-01-07
[builder] memory use optimization

### LINKED ISSUE #776: [builder] add group meta consolidation
When TileDB-Inc/TileDB-Py#1833 is finalized, enable consolidation/vacuuming of `group_meta` in `consolidate.py::consolidate_tiledb_object()`

Currently it is commented out, awaiting the fix.

---

2023-11-02 update:  fix requires release of 2.18 core / TileDB-Py 0.24.  ETA before end of 2023.

2023-10-28 update: landed in 2.17.2, but was not incorporated into TileDB-Py.  Waiting on TileDB-Inc/TileDB-Py#1833 to be released.

2023-10-17 update:  fix is awaiting release of 2.17.2

bkmartinjr:
First of several refactoring PRs focused on improving the state of the builder.

This is focused on memory optimizations, in preparation for further simplification and cleanup of the work scheduling.  As an end result, builder will now run on an `r` class instance (i.e., 8GiB/CPU) without swap, and will should run without excessive paging/thrash on an m-class (4GiB/cpu) instance.

Memory usage improvements:
* AnnData loading: most of the highly variable memory allocation (due to different H5AD sizes) is removed.
    * The only allocation driven by H5AD size is the loading of the H5AD obs and var dataframes, which are fairly small compared to the rest of the object.
    * Dataframe loading now allows the caller to specify which columns are required, bypassing the need to allocate memory for unused columns.
* build steps operating on X now work with a fixed size memory budget via better chunking
* Removed unnecessary copies and added explicit large object deletion in inner loops, reducing peak memory use

Primary code changes:
* Added a system resource usage logging thread to assist in monitoring/debugging memory usage
* Replaced use of AnnData with an AnnData-like proxy object. Proxy does not materialize X, or load unused slots (obs{m,p}, etc). This removes the major source of variable memory use per worker.
* open_anndata and make_anndata_celll_filter reimplemented to provide (mostly) the same API / behavior. The one major change is that open_anndata is no longer an iterator, but instead is a one-shot function (i.e., it just opens a single anndata).
* Consolidation API made more flexible, allowing more concurrency (allows this step to overlap with others). NOTE: this API is not yet advantaged, but will be in a later PR.
* fine-tune concurrency caps (guardrails to prevent bumping into process and kernel resource limits on high CPU boxes)
* explicit calls to delete large objects, reducing peak memory usage

Minor changes:
* auto-updated pre-commit versions
* add group metadata consolidation. Fixes #776 
* update all tests to match new anndata loader

This update has been successfully tested on:
* r6i.32xlarge - with and without swap
* r6a.48xlarge - without swap

In addition, I confirmed that the *old* builder (ie, prior to this PR) validates the output of this modified builder.


tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/experiment_builder.py
```diff
@@ -470,77 +471,106 @@ def _accumulate_all_X_layers(
...
                 ),
                 AxisStats(eb.name, obs_stats, var_stats),
             )
         )
 
         # tidy
-        del ad, raw_X, raw_var, local_var_joinids, row, col, X_remap, pres_data, pres_cols, obs_stats, var_stats
+        del ad, local_var_joinids, obs_stats, var_stats
 
     gc.collect()
```
atolopko-czi, line 571: indent to gc to free the mem for `ad, local_var_joinids, obs_stats, var_stats` after each iteration?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/experiment_builder.py
```diff
@@ -602,7 +632,22 @@ def _accumulate_X(
...
+                (
+                    max(dataset.mean_genes_per_cell, 3000)
+                    * min(ACCUM_X_STRIDE, dataset.dataset_total_cell_count)
+                    * 20
+                    * 8
+                )
+                + (100_000 * 40 + dataset.dataset_total_cell_count * 32) * 8
+                + (dataset.dataset_total_cell_count * 40) * 8
+                + (2 * 1024**3)
+            ),
```
atolopko-czi, line 650: just curious, is the "overhead" value a guesstimate? i.e. "assumed inefficiency of data representation due to inefficient typing, copying, etc.". Or is it concretely known to be 8?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/anndata.py
```diff
@@ -19,99 +23,244 @@
...
+        filename: str | PathLike[str],
+        *,
+        view_of: Self | None = None,
+        obs_idx: slice | npt.NDArray[np.int64] | None = None,
+        var_idx: slice | npt.NDArray[np.int64] | None = None,
+        obs_column_names: Optional[Tuple[str, ...]] = None,
+        var_column_names: Optional[Tuple[str, ...]] = None,
+    ):
+        self.filename = filename
+        self.is_view = view_of is not None
```
atolopko-czi, line 100: not used; keep anyway?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/anndata.py
```diff
@@ -19,99 +23,244 @@
...
+    * removes the overhead when you need to index a view (where AnnData materializes)
+
+    In the future, if we need additional slots accessible, we may need to add sub-proxies
+    for uns, obs*, var*, etc. But at the moment, the Builder has no use for these so they
+    are omitted.
+    """
+
+    _obs: pd.DataFrame
+    _var: pd.DataFrame
+    _X: h5py.Dataset | CSRDataset | CSCDataset
```
atolopko-czi, line 87: Currently, is only CSR being tested by unit tests? Do we need unit test all three?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/anndata.py
```diff
@@ -19,99 +23,244 @@
...
+            self._obs, self._var, self._X = (view_of._obs, view_of._var, view_of._X)
+            assert obs_idx is not None
+            assert var_idx is not None
+            self._obs_idx = obs_idx
+            self._var_idx = var_idx
+
+    @property
+    def X(self) -> sparse.spmatrix | npt.NDArray[np.integer[Any] | np.floating[Any]]:
+        # Do two separate slices, as the underlying AnnData CS*Dataset
+        # does bad things when you do slicing on both axis simultaneously.
```
atolopko-czi, line 116:
```suggestion
        # does bad things when you do slicing on both axes simultaneously.
```

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/anndata.py
```diff
@@ -19,99 +23,244 @@
...
-        # cell/feature organism values.
-        feature_reference_organisms = set(ad.var.feature_reference.unique()) - FEATURE_REFERENCE_IGNORE
-        if len(feature_reference_organisms) > 1:
-            logging.info(f"H5AD ignored due to multi-organism feature_reference: {h5ad.dataset_id}")
-            continue
+def _index_index(prev: Index1D, new: Index1D, length: int) -> slice | npt.NDArray[np.int64]:
+    """Index an index"""
+    if isinstance(prev, slice):
+        if isinstance(new, slice):
+            # conviently, ranges support indexing!
```
atolopko-czi, line 35:
```suggestion
            # conveniently, ranges support indexing!
```

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/anndata.py
```diff
@@ -19,99 +23,244 @@
...
+        return len(self.var)
+
+    @property
+    def shape(self) -> tuple[int, int]:
+        return self.n_obs, self.n_vars
+
+    def __len__(self) -> int:
+        return self.shape[0]
+
+    def __getitem__(self, key: Index) -> "AnnDataProxy":
```
atolopko-czi, line 151: This method and/or _index_index could probably use some explicit tests. It handles lots cases like stepped slices, using slices vs masks vs points, etc..  If nothing else, the tests will help document the explicit requirements of how an anndataproxy view can be sliced.  Maybe also a comment, "supports slicing in the same manner as AnnData"

I might rename `_index_index` to `_slice_index`, though naming is a personal matter. :) AnnData docs do refer to this operation "slicing", so seems reasonable.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/anndata.py
bkmartinjr: I think there is a use for it in some scenarios, but not in the builder - will delete in the name of simplicity

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/anndata.py
bkmartinjr: We need all three, and they are all actively in use.  As of today, in our CxG datasets, in both X and raw.X, there are:
* 1965 CSR matrices
* 3 CSC matrices
* 18 dense ndarrays

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/experiment_builder.py
```diff
@@ -470,77 +471,106 @@ def _accumulate_all_X_layers(
...
+
+            X = ad[idx : idx + ACCUM_X_STRIDE].X
+            if isinstance(X, np.ndarray):
+                X = sparse.csr_matrix(X)
+
+            if not is_nonnegative_integral(X):
+                logging.error(f"{dataset.dataset_id} contains non-integer or negative valued data")
+
+            X.eliminate_zeros()
+            gc.collect()
```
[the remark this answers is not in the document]
bkmartinjr: pretty much in the noise, so I don't see much value

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/experiment_builder.py
```diff
@@ -470,77 +471,106 @@ def _accumulate_all_X_layers(
...
+
+            # accumulate obs/var axis stats
+            _obs_stats, _var_stats = _get_axis_stats(X, idx + dataset_obs_joinid_start, local_var_joinids)
+            obs_stats = pd.concat([obs_stats, _obs_stats])
```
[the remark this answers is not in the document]
bkmartinjr: I don't see how it would - it is my understanding that Pandas requires the column (series) to be contiguous (a la numpy ndarray).

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/experiment_builder.py
bkmartinjr: guesstimate, informed by empirical testing. I've been doing all my test builds on hosts w/o swap, to catch the egregious cases of under-reserving memory.

tools/cellxgene_census_builder/src/cellxgene_census_builder/util.py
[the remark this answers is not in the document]
bkmartinjr: I'm open to that, but I have not found an obvious way to organize it. It also significantly increases the cost of these kinds of utilities, as they now need to start worrying about backward compat, docstrings, etc.

Perhaps if we can talk through how you might organize & manage such a depdenency, we can think think about what code is has the right ROI (i.e., value >> overhead) to migrate into it...

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/experiment_builder.py
```diff
@@ -373,16 +370,18 @@ def write_X_normalized(self, args: CensusBuildArgs) -> None:
         feature_length = self.var_df.feature_length.to_numpy()
 
         if args.config.multi_process:
-            STRIDE = 1_000_000  # controls TileDB fragment size, which impacts consolidation time
-            # memory budget: 3 attribute buffers * 3 threads * 100% buffer
-            mem_budget = 3 * 3 * 2 * int(SOMA_TileDB_Context().tiledb_ctx.config()["soma.init_buffer_bytes"])
+            WRITE_NORM_STRIDE = 2**18  # controls TileDB fragment size, which impacts consolidation time
+            mem_budget = (
+                int(20 * WRITE_NORM_STRIDE * 4000 * 2) + (3 * 1024**3) + feature_length.nbytes + is_smart_seq.nbytes
```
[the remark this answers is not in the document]
bkmartinjr: comment added.  But these are rough estimates based on my read of the code, plus a bunch of empirical testing.  The goal is "good enough" (ie.., budget correlates with reality, not exactly models it)

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/experiment_builder.py
```diff
@@ -760,46 +814,31 @@ def _write_X_normalized(args: Tuple[str, int, int, npt.NDArray[np.int64], npt.ND
...
-                        pa.Table.from_pydict(
-                            {
-                                "soma_dim_0": obs_indices[d0],
-                                "soma_dim_1": d1,
-                                "soma_data": data,
-                            }
-                        )
+            for tbl, (obs_indices, _) in (
+                X_raw.read(coords=(slice(obs_joinid_start, obs_joinid_start + n - 1),))
+                .blockwise(axis=0, size=2**16, reindex_disable_on_axis=[1], eager=False)
```
ebezzi, line 819: Just curious, why did we drop eager here?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/anndata.py
```diff
@@ -19,99 +23,248 @@
...
+
+    def _load_dataframe(self, elem: h5py.Group, column_names: Optional[Tuple[str, ...]]) -> pd.DataFrame:
+        # if reading all, just use the built-in
+        if not column_names:
+            return cast(pd.DataFrame, read_elem(elem))
+
+        # else read each user-specified column/index separately, taking care to preserve the
+        # original dataframe column ordering
+        assert len(column_names) > 0
+        assert elem.attrs["encoding-type"] == "dataframe" and elem.attrs["encoding-version"] == "0.2.0"
```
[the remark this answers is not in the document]
bkmartinjr: I am explicitly not supporting other users, and it would require more work to do so. Will add the explicit error message.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/experiment_builder.py
```diff
@@ -475,72 +480,101 @@ def _accumulate_all_X_layers(
...
+
+        for idx in range(0, ad.n_obs, ACCUM_X_STRIDE):
+            logging.debug(f"{eb.name}/{layer_name}: X chunk {idx//ACCUM_X_STRIDE + 1} {dataset.dataset_id}")
+
+            X = ad[idx : idx + ACCUM_X_STRIDE].X
+            if isinstance(X, np.ndarray):
+                X = sparse.csr_matrix(X)
+
+            if not is_nonnegative_integral(X):
+                logging.error(f"{dataset.dataset_id} contains non-integer or negative valued data")
```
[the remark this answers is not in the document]
bkmartinjr: The current code is identical to previous behavior.  That said, good question - if this fails, it indicates a dataset that has violated the schema. We don't have a strict policy about how these are handled, but for many errors it will cause a hard failure. So I think the right approach is to simply change this to an assert.

ebezzi (approved):

PR #934, opened 2024-01-10
[builder] add http fetch retries

bkmartinjr:
Adds HTTP get retry to the builder where it was missing such logic.  Hardening only - no change in baseline behavior

### DIFF
--- a/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/util.py
+++ b/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/util.py
@@ -5,6 +5,7 @@ from typing import Any, Iterator, Optional, Union
 import numpy as np
 import numpy.typing as npt
 import requests
+import urllib3
 from scipy import sparse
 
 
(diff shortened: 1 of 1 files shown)


tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/util.py
```diff
@@ -63,7 +64,13 @@ def array_chunker(
 
 
 def fetch_json(url: str, delay_secs: float = 0.0) -> object:
-    response = requests.get(url)
+    s = requests.Session()
+    retries = urllib3.util.Retry(
+        backoff_factor=1,  # i.e., sleep for [0s, 2s, 4s, 8s, ...]
+        status_forcelist=[500, 502, 503, 504],  # 5XX can occur on CDN failures, so retry
```
ebezzi, line 70: Something interesting - according to CloudWatch, Cloudfront is only reporting 4xx errors and no 5xx. Are we sure the errors you saw were 5xx?

bkmartinjr: Yes, we absolutely get 500's on occasion.  I don't think these get reported, as they are hard fails in the AWS infra.  But still to be expected as nothing is 100% reliable.

ebezzi: Wondering if those 4xxs are related then. Likely not.

ebezzi (approved):

PR #945, opened 2024-01-22
scib-metrics scripts

ebezzi:


mlin:
@ebezzi is there a distinct `scib-metrics-config-plot.yaml` to add here?

ebezzi:
> @ebezzi is there a distinct `scib-metrics-config-plot.yaml` to add here?

No, if I recall correctly, you're just supposed to call the script with the config file as a parameter. The one included here should be fine. We can delete that if branch.

tools/models/metrics/run-scib-plots.py
```diff
@@ -0,0 +1,174 @@
...
+import plotnine as pt
+
+from typing import List, Dict
+
+
+def main():
+    try:
+        file = sys.argv[1]
+    except IndexError:
+        file = "scib-metrics-config-plot.yaml"
```
mlin, line 14: @ebezzi Isn't there a `scib-metrics-config-plot.yaml` not yet checked in here, distinct from `scib-metrics-config.yaml` which is checked in?

ebezzi: See https://github.com/chanzuckerberg/cellxgene-census/pull/945#issuecomment-2622837505

mlin: @ebezzi This PR includes one YAML file `scib-metrics-config.yaml` which feeds into `run-scib.py`, with keys such as `census` `embeddings` `tissues`

Here we have a different script `run-scib-plots.py` which appears to expect a different YAML file named `scib-metrics-config-plot.yaml` with keys such as `metrics_files` `metrics_plots` `decimal_values` `normalize_entropy` etc. accessed below.

Did I miss something?

PR #946, opened 2024-01-23
[builder] remove unused H5AD files from build

### LINKED ISSUE #935: The builder should only save h5ads that pass filtering
## Motivation

Currently, the builder syncs all the downloaded h5ads on `do_data_copy`, including the ones that do not pass the filtering conditions. We should only sync the datasets that are actually part of the census build.

## Definition of Done

- Only datasets that belong to the census should be copied.
- A test (using a mocked S3 storage) should be created to verify this behavior.

bkmartinjr:
Fixes #935 

The Census build will only include H5AD files that contribute to the SOMA objects.

### DIFF
--- a/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/build_soma.py
+++ b/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/build_soma.py
@@ -1,5 +1,7 @@
 import gc
 import logging
+import os
+import pathlib
 from datetime import datetime, timezone
 from typing import Iterator, List
 
@@ -82,6 +84,9 @@ def build(args: CensusBuildArgs) -> int:
         args.h5ads_path.as_posix(), datasets, experiment_builders, args
     )
 
+    # Prune datasets that we will not use, and do not want to include in the build
+    prune_unused_datasets(args.h5ads_path, datasets, filtered_datasets)
+
     # Step 4 - populate X layers
     build_step4_populate_X_layers(args.h5ads_path.as_posix(), filtered_datasets, experiment_builders, args)
     gc.collect()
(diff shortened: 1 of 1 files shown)


tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/build_soma.py
```diff
@@ -104,6 +109,17 @@ def build(args: CensusBuildArgs) -> int:
     return 0
 
 
+def prune_unused_datasets(assets_path: pathlib.Path, all_datasets: List[Dataset], used_datasets: List[Dataset]) -> None:
+    unused_datasets = [d for d in all_datasets if d.dataset_id not in set(d.dataset_id for d in used_datasets)]
```
atolopko-czi, line 113: nit: use set.difference() or `-` for readability?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/build_soma.py
```diff
@@ -104,6 +109,17 @@ def build(args: CensusBuildArgs) -> int:
...
+    assert all(d.dataset_total_cell_count == 0 for d in unused_datasets)
+    assert all(d.dataset_total_cell_count > 0 for d in used_datasets)
+    assert not (set(d.dataset_id for d in used_datasets) & set(d.dataset_id for d in unused_datasets))
```
atolopko-czi, line 116: nit: use `set.isdisjoint()`

atolopko-czi (approved):
LGTM, with minor/optional suggestions to leverage more set functions for readability

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/build_soma.py
bkmartinjr: I'm not clear how that simplifies things given that the Dataset object isn't hashable, and I need to end up with a list of them.

If I understand, you are proposing:
>     unused_datasets = [d for d in all_datasets if d.dataset_id in (set(d.dataset_id for d in all_datasets) - set(d.dataset_id for d in used_datasets))]

which seems verbose and not any clearer.

I guess I could allocate another data structure at some expense which is hashable, eg.,, which would look like:
```
    all_datasets_by_id = {d.dataset_id: d for d in all_datasets}
    unused_dataset_ids = {d.dataset_id for d in used_datasets}
    used_dataset_ids = set(all_datasets_by_id) - unused_dataset_ids
    assert all(all_datasets_by_id[id].dataset_total_cell_count == 0 for id in unused_datasets)
    assert all(all_datasets_by_id[id].dataset_total_cell_count > 0 for id in used_datasets)
    assert set(all_datasets_by_id[id].dataset_id for id in used_datasets).isdisjoint(
        set(all_datasets_by_id[id].dataset_id for d in unused_datasets)
    )

    for dataset_id in used_dataset_ids:
        os.remove(assets_path / all_datasets_by_id[dataset_id].dataset_h5ad_path)
```

Also doesn't seem like a win.

Can you clarify your intent?  I think I'm missing something.

ebezzi (approved):

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/build_soma.py
atolopko-czi: Indeed, dataset objs would need to be hashable for improve readability, so best as is!

PR #947, opened 2024-01-23
[misc] move more workflows to self-hosted runners

bkmartinjr:
Move an additional 4 workflows to self-hosted runners

### DIFF
--- a/.github/workflows/close-stale-prs.yml
+++ b/.github/workflows/close-stale-prs.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   close-issues:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, Linux, X64]
     permissions:
       issues: write
       pull-requests: write
--- a/.github/workflows/py-dependency-check.yml
+++ b/.github/workflows/py-dependency-check.yml
@@ -20,7 +20,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [single-cell-8c64g-runner, macos-latest]
         python-version: ["3.8", "3.9", "3.10", "3.11"]
 
     runs-on: ${{matrix.os}}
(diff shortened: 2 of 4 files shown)


.github/workflows/py-dependency-check.yml
```diff
@@ -20,7 +20,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [single-cell-8c64g-runner, macos-latest]
```
beroy, line 23: Why does the runner appear in the `os` field?

.github/workflows/close-stale-prs.yml
```diff
@@ -5,7 +5,7 @@ on:
 
 jobs:
   close-issues:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, Linux, X64]
```
beroy, line 8: Just to be sure, in this PR you are not using large your intention is not to use `single-cell-8c64g-runner` or `single-cell-1tb-runner` runners.

beroy (approved):

.github/workflows/py-dependency-check.yml
bkmartinjr: if you look a few lines below, these are the labels we provide to `runs-on`, which accepts a runner.  I suppose we could rename `os` to something better (e.g., `runner`?), but using `os` has been our convention for a while

.github/workflows/close-stale-prs.yml
bkmartinjr: correct - this workflow can run on a much smaller runner. This set of labels is based on Jake's previous advice (see py-unittests.yml for an example)

.github/workflows/py-dependency-check.yml
beroy: I see it now

beroy (approved):

ebezzi (approved):

PR #959, opened 2024-01-29
Early version of profiler harness

beroy:
Include a basic benchmark as the starting point and needed scripts

### DIFF
--- a/.github/workflows/profiler.yml
+++ b/.github/workflows/profiler.yml
@@ -1,0 +1,40 @@
+name: Profiler
+
+on:
+  pull_request:
+
+jobs:
+  run_profiler:
+    name: Run Profiler
+    strategy:
+      matrix:
+        os: [single-cell-8c64g-runner]
+        python-version: ["3.11"]
+    runs-on: ${{matrix.os}}
+    permissions: # these permissions must be set for AWS auth to work!
+      id-token: write
+      contents: read
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-region: us-west-2
+          role-to-assume: arn:aws:iam::401986845158:role/MyNewPlayground
+          role-session-name: PushDockerImage
+
+      - name: Install AWS CLI v2
+        run:  |
+                sudo apt install unzip
+                curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o /tmp/awscliv2.zip
+                unzip -q /tmp/awscliv2.zip -d /tmp
+                rm /tmp/awscliv2.zip
+                sudo /tmp/aws/install --update
+                rm -rf /tmp/aws
+      - name: Run all tests
+        run: |
+          ./tools/perf_checker/perf_checker.sh
--- a/tools/perf_checker/perf_checker.py
+++ b/tools/perf_checker/perf_checker.py
@@ -1,0 +1,59 @@
+import argparse
+import json
+import logging
+import re
+from typing import Any
+
+import attr
+import profiler
+
+
+def format_string(text: str) -> Any:
+    return re.sub("\n", " ", text)
+
+
+# The script takes a command and a database path and looks
+# the performance anomalies in the performance history of that
+# command across the profiled runs.
+
+logger = logging.getLogger()
+logger.setLevel(logging.DEBUG)
+
+parser = argparse.ArgumentParser()
+parser.add_argument("command", type=str)
+parser.add_argument("db_path", type=str)
+
+args = parser.parse_args()
+
+# Processes the set of previously written logs
+# The threshold (ratio) of allowable performance degradation between profiling runs
+threshold = 1.10
+
+db = profiler.data.S3ProfileDB(args.db_path)
+command_profiles = db.find(f"{args.command}")
+
+
+if len(command_profiles) >= 2:
+    first_profile = command_profiles[0]
+    curr_profile = command_profiles[-1]
+    first_time = first_profile.elapsed_time_sec
+    curr_time = curr_profile.elapsed_time_sec
+
+    formatted_first_profile = json.dumps(format_string(str(attr.asdict(first_profile))))
+    formatted_curr_profile = json.dumps(format_string(str(attr.asdict(curr_profile))))
+
+    logging.info("****************************")
+    logging.info(f"Current time {curr_time} vs first time {first_time}")
+    logging.info("****************************")
+    logging.info(f"First profile: {formatted_first_profile}")
+    logging.info("****************************")
+    logging.info(f"Current profile: {formatted_curr_profile}")
+    logging.info("****************************")
+    logging.info(
+        f"TileDB version ver = first: {first_profile.tiledbsoma_version} curr: {curr_profile.tiledbsoma_version}"
+    )
+    if float(curr_time) > threshold * float(first_time):
+        raise SystemExit(f"Major performance degradation detected on {args.command}")
+
+    if threshold * float(curr_time) < float(first_time):
+        logging.info(f"Major performance increase detected on {args.command}")
--- a/tools/perf_checker/perf_checker.sh
+++ b/tools/perf_checker/perf_checker.sh
@@ -1,0 +1,30 @@
+.sh
+#!/bin/sh
+set -euox pipefail
+
+dbpath="census-profiler-tests-trial"
+sudo apt install time
+
+python3.11 -m venv ~/venv
+. ~/venv/bin/activate
+
+pip install psutil; pip install gitpython; pip install somacore; pip install tiledbsoma; pip install cellxgene_census; pip install boto3
+
+# Download the repo including the profiler
+cd ../
+git clone https://github.com/single-cell-data/TileDB-SOMA.git
+# Downloading TileDB-SOMA (remove the next line once the branch is merged)
+cd TileDB-SOMA/profiler
+git checkout census_profiler
+pip install .
+cd ../../cellxgene-census/
+
+# New benchmarks must be added to this list
+declare -a benchmarks=("./tools/perf_checker/test_anndata_export.py")
+
+# Running all benchmarks and checking performance changes
+for benchmark in ${benchmarks}
+do
+  python -m profiler "python ${benchmark}" $dbpath
+  python ./tools/perf_checker/perf_checker.py "python ${benchmark}" $dbpath
+done
--- a/tools/perf_checker/test_anndata_export.py
+++ b/tools/perf_checker/test_anndata_export.py
@@ -1,0 +1,14 @@
+import cellxgene_census
+import tiledbsoma as soma
+
+
+def main() -> None:
+    with cellxgene_census.open_soma(census_version="stable") as census:
+        with census["census_data"]["homo_sapiens"].axis_query(
+            measurement_name="RNA",
+            obs_query=soma.AxisQuery(value_filter="""tissue_general == 'eye'"""),
+        ) as query:
+            query.to_anndata(X_name="raw")
+
+
+main()


.github/workflows/profiler.yml
```diff
@@ -0,0 +1,30 @@
+name: Performance check
```
atolopko-czi, line 1: would align this name with the workflow file name

.github/workflows/profiler.yml
```diff
@@ -0,0 +1,30 @@
...
+    name: Run Profiler
+    runs-on: ubuntu-latest
+    permissions: # these permissions must be set for AWS auth to work!
+      id-token: write
+      contents: read
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
atolopko-czi, line 17: `1` if you just want the HEAD commit.

tools/perf_checker/perf_checker.sh
```diff
@@ -0,0 +1,41 @@
...
+declare -a benchmarks=("./tools/perf_checker/benchmark1.py")
+
+# Downloading TileDB-SOMA (branch or main once merged)
+git clone https://github.com/single-cell-data/TileDB-SOMA.git
+cd TileDB-SOMA
+git checkout census_profiler
+cd ../
+
+# Running all benchmark and checking performance changes
+arraylength=${#benchmarks[@]}
```
atolopko-czi, line 36: rename `num_profiling_runs`, e.g.?

tools/perf_checker/benchmark1.py
atolopko-czi, line 1: This is is testing "export anndata", so why not name the file as such?

tools/perf_checker/perf_checker.sh
```diff
@@ -0,0 +1,41 @@
...
+dbpath=`pwd`/mount-s3
+echo "Mount-S path = ${dbpath}"
+
+# New benchmarks must be added to this list
+declare -a benchmarks=("./tools/perf_checker/benchmark1.py")
+
+# Downloading TileDB-SOMA (branch or main once merged)
+git clone https://github.com/single-cell-data/TileDB-SOMA.git
+cd TileDB-SOMA
+git checkout census_profiler
```
atolopko-czi, line 32: This will be removed before merging to main, right? (I think that's what the comment above is saying)

tools/perf_checker/perf_checker.sh
```diff
@@ -0,0 +1,41 @@
...
+cd ../
+
+# Running all benchmark and checking performance changes
+arraylength=${#benchmarks[@]}
+for (( i=0; i<${arraylength}; i++ ))
```
atolopko-czi, line 37: you can probably just iterate the array, like `for benchmark in ${benchmarks}` and avoid the indexing

tools/perf_checker/perf_checker.sh
```diff
@@ -0,0 +1,41 @@
...
+sudo wget https://s3.amazonaws.com/mountpoint-s3-release/latest/x86_64/mount-s3.deb
+sudo apt install -y ./mount-s3.deb
+
+# Setting up mount-s3
+mkdir ./mount-s3
+mkdir ./s3_cache
+mount-s3 census-profiler-tests ./mount-s3 --cache ./s3_cache  --metadata-ttl 300
+
+dbpath=`pwd`/mount-s3
+echo "Mount-S path = ${dbpath}"
```
atolopko-czi, line 24: `mount-s3`?

tools/perf_checker/perf_checker.sh
```diff
@@ -0,0 +1,41 @@
...
+pip install somacore
+pip install tiledbsoma
+pip install cellxgene_census
+
+# Installing mount-s3
+sudo wget https://s3.amazonaws.com/mountpoint-s3-release/latest/x86_64/mount-s3.deb
+sudo apt install -y ./mount-s3.deb
+
+# Setting up mount-s3
+mkdir ./mount-s3
```
atolopko-czi, line 19: would rename to indicate its contents instead? could be same as bucket name: `census-profiler-tests`. Would also add comment explaining that this is necessary to persist the profiling run data that are performed below.

tools/perf_checker/perf_checker.sh
atolopko-czi: echo not needed if using `set -x` above

tools/perf_checker/benchmark1.py
```diff
@@ -0,0 +1,21 @@
...
+def main():
+    t1 = perf_counter()
+    with cellxgene_census.open_soma(**census_S3_latest) as census:
+        with census["census_data"]["homo_sapiens"].axis_query(
+            measurement_name="RNA",
+            obs_query=soma.AxisQuery(value_filter="""tissue_general == 'eye'"""),
+        ) as query:
+            query.to_anndata(X_name="raw")
+    t2 = perf_counter()
+    print(f"End to end time {t2 - t1}")
```
atolopko-czi, line 18: Is timing necessary here, since this is being called by the profiler, which is necessarily timing this script?

tools/perf_checker/benchmark1.py
atolopko-czi: and maybe put all the benchmarks into a subdir. The `perf_checker.sh` script could just iterate through all the files, allowing the explicit definition (and future update) of the `benchmarks` array in the script.

tools/perf_checker/perf_checker.sh
```diff
@@ -0,0 +1,41 @@
...
+git clone https://github.com/single-cell-data/TileDB-SOMA.git
+cd TileDB-SOMA
+git checkout census_profiler
+cd ../
+
+# Running all benchmark and checking performance changes
+arraylength=${#benchmarks[@]}
+for (( i=0; i<${arraylength}; i++ ))
+do
+  python  ./TileDB-SOMA/profiler "python ${benchmarks[$i]}" $dbpath -t time
```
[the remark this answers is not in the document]
atolopko-czi: is `-t time` going to run `gnu-time`? the profiler requires `gnu-time`-formatted output.  would install `gnu-time` if needed, in profiler.yml

tools/perf_checker/perf_checker.sh
atolopko-czi: also, per another comment, if all benchmark scripts are placed in a subdir, you can just glob and iterate them: `for benchmark in benchmarks/*py`

tools/perf_checker/benchmark1.py
beroy: Good point

tools/perf_checker/perf_checker.sh
beroy: Great ideas both

tools/perf_checker/perf_checker.sh
beroy: Mount-s3 is an s3 fused drive

tools/perf_checker/perf_checker.sh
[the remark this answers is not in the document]
beroy: Good point. I'll fix it need to figure how to enable it on linux

tools/perf_checker/perf_checker.sh
beroy: Yes you are right

tools/perf_checker/perf_checker.sh
beroy: Correct

tools/perf_checker/benchmark1.py
beroy: For the time being the agreement is for me to have this benchmark and later @bkmartinjr and @pablo-gar decide on the rest of them

tools/perf_checker/perf_checker.sh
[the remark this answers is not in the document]
beroy: I actually noticed that gtime only works on macos and was not able to install on ubunutu. LMK if there's a way to get it to work on ubuntu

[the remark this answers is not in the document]
atolopko-czi: It's usually available under `/usr/bin/time`. If not try, try `sudo apt install time`

.github/workflows/profiler.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-region: us-west-2
+          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
+          role-session-name: PushDockerImage
+
+      - name: Run all tests
+        run: |
+          python -m venv profiler_env
```
ebezzi, line 28: Any reason to use venv here? This will run on a sandboxed environment regardless.

tools/perf_checker/perf_checker.py
```diff
@@ -0,0 +1,42 @@
...
+parser.add_argument("db_path", type=str)
+
+args = parser.parse_args()
+
+# Processes the set of previously written logs
+# The threshold (ratio) of allowable performance degradation between profiling runs
+threshold = 1.10
+
+db = profiler.data.FileBasedProfileDB(args.db_path)
+dt = db.find(f"{args.command}")
```
ebezzi, line 20: Suggest renaming `dt`. It is not clear what the variable contains from its name.

tools/perf_checker/perf_checker.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+if len(dt) >= 2:
+    first_profile = dt[0]
+    curr_profile = dt[len(dt) - 1]
```
ebezzi, line 25: Can rewrite as: `dt[-1]`

tools/perf_checker/perf_checker.py
```diff
@@ -0,0 +1,42 @@
...
+    first_time = first_profile.elapsed_time_sec
+    curr_time = curr_profile.elapsed_time_sec
+
+    formatted_first_profile = str(first_profile).replace("\\n", "\n").replace("\\t", "\t")
```
ebezzi, line 29: This is a bit suspicious - why the double escaping in the first place?

tools/perf_checker/perf_checker.py
```diff
@@ -0,0 +1,42 @@
...
+    formatted_curr_profile = str(curr_profile).replace("\\n", "\n").replace("\\t", "\t")
+
+    print(f"********************\n\nFirst profile:\n {formatted_first_profile}")
```
ebezzi, line 32: I would recommend switching to using a logger and keep each log to a single line. Having multiple lines will make logs hard to read in an environment where lines are split. Consider using json logging if they span multiple lines. This may take a while so it can be done in a separate PR or just deferred to a later date.

tools/perf_checker/perf_checker.py
```diff
@@ -0,0 +1,42 @@
...
+    print(f"********************\n\nCurrent profile:\n {formatted_curr_profile}\n********************\n")
+    print(f"TileDB version ver = first: {first_profile.tiledbsoma_version} curr: {curr_profile.tiledbsoma_version}")
+    if float(curr_time) > threshold * float(first_time):
+        print(f"Major performance increase detected on {args.command}: curr: {first_time} vs first: {curr_time}")
+        raise SystemExit(
+            f"Potential performance degradation detected on {args.benchmark}: curr: {first_time} vs first: {curr_time}"
```
ebezzi, line 38: `first_time` and `curr_time` are swapped.

tools/perf_checker/perf_checker.py
```diff
@@ -0,0 +1,42 @@
...
+        )
+
+    if threshold * float(curr_time) < float(first_time):
```
ebezzi, line 41: Same here.

ebezzi: Also, how does this line compare to the previous one (35)? The log is the same but the condition is different.

tools/perf_checker/perf_checker.py
ebezzi: Also, is this `Potential` or `Major`? The print and return do not agree.

tools/perf_checker/perf_checker.sh
```diff
@@ -0,0 +1,51 @@
+.sh
+#!/bin/sh
+set -euox pipefail
+
+# Download the right version of python
+sudo apt install python3.11 -y
```
ebezzi, line 6: Can put those two on the same line.

tools/perf_checker/perf_checker.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# Setting up mount-s3. We use S3 file system as it is necessary to persist the
+# profiling run data that are performed below
+mkdir ./census-profiler-tests
+mkdir ./s3_cache
+mount-s3 census-profiler-tests ./census-profiler-tests --cache ./s3_cache  --metadata-ttl 300
+dbpath=`pwd`/census-profiler-tests
+
+# Download gnu time tool
+sudo apt-get update -y
```
ebezzi, line 21: Might as well more apt-get update at the beginning of the file.

tools/perf_checker/test_anndata_export.py
```diff
@@ -0,0 +1,19 @@
+from sys import stderr
+
+import cellxgene_census
+import tiledbsoma as soma
+
+print("Starting bm 1", file=stderr)
+census_S3_latest = dict(census_version="2024-01-01")
```
ebezzi, line 7:
```suggestion
census_S3_latest = dict(census_version="latest")
```

ebezzi (changes requested):

tools/perf_checker/perf_checker.py
beroy: Fixed

tools/perf_checker/test_anndata_export.py
beroy: The issue with using the latest is that it may lead to performance degradation once the latest changes

tools/perf_checker/benchmark1.py
beroy: Done

tools/perf_checker/perf_checker.sh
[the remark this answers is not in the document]
beroy: Fixed

.github/workflows/profiler.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-region: us-west-2
+          role-to-assume: arn:aws:iam::401986845158:role/MyNewPlayground #arn:aws:iam::401986845158:role/PlaygroundS3 #
```
ebezzi, line 27: Consolidate/remove comments

tools/perf_checker/perf_checker.sh
```diff
@@ -0,0 +1,36 @@
...
+#!/bin/sh
+set -euox pipefail
+
+dbpath="census-profiler-tests-trial"
+sudo apt install time
+
+python3.11 -m venv ~/venv
+. ~/venv/bin/activate
+
+pip install psutil
```
ebezzi, line 11: These can go on a single line

tools/perf_checker/test_anndata_export.py
ebezzi: If you pin a non LTS version it will go away after one month. That said, this is a test file so it's probably fine, but leave a comment for posterity.

ebezzi (approved):
LGTM overall. Left a few nitpicks.

tools/perf_checker/perf_checker.sh
bkmartinjr: This was not removed and is now present in `main`, and has broken the CI

PR #967, opened 2024-02-01
[builder] update dependency for security advisory

bkmartinjr:
Update aiohttp package only due to security advisory


bkmartinjr:
redundant - fix made in another PR

PR #971, opened 2024-02-01
[misc] chore: upgrade and loosen experimental package requirements

### LINKED ISSUE #968: [python] dependency pins are overly specific
The package experimental optional dependencies hard-pin a number of packages. It would be ideal if we could loosen these version dependencies to what is actually needed, allowing downstream users a bit more flexibility in choosing package versions.

Currently the required dependencies follow this style, but the optional do not.  Examples:

```
dependencies= [
    "anndata",
    "numpy>=1.21,<1.25",  # numpy is constrained by numba and the old pip solver
    "requests",
    "typing_extensions",
    "s3fs>=2021.06.1",
]

[project.optional-dependencies]
experimental = [
    "torch==2.0.1",
    "torchdata==0.6.1",
    "scikit-learn==1.3.1",
    "scikit-misc==0.2.0",
    "psutil==5.9.5",
    "datasets~=2.14",
    "tdigest~=0.5",
]
```

In an ideal world, would like something like `pytorch>=2.0.0` or some such....  Especially for the very general packages such as `datasets` and `psutil` which are used downstream

atolopko-czi:
Resolves #968 

Upgraded packages to latest version. 

For requirements restrictions:
- For packages that are post-1.0, allowed for any minor bump
- For packages that are pre-1.0 (`0.x.y`), allowed for any patch bump

Update:
- for packages which no longer support Python 3.8, allow older variants that are known to work (specific to scikit-misc)


api/python/cellxgene_census/pyproject.toml
```diff
@@ -41,12 +41,12 @@ dependencies= [
...
-    "scikit-learn==1.3.1",
-    "scikit-misc==0.2.0",
-    "psutil==5.9.5",
-    "datasets~=2.14",
+    "torch~=2.0",
+    "torchdata~=0.7",
+    "scikit-learn~=1.0",
+    "scikit-misc~=0.3",
+    "psutil~=5.0",
+    "datasets~=2.16",
```
bkmartinjr, line 49: Mike suggested `datasets~=2.0` - I assume you need a feature introduced in 2.16?

api/python/cellxgene_census/pyproject.toml
```diff
@@ -41,12 +41,12 @@ dependencies= [
...
-    "torch==2.0.1",
-    "torchdata==0.6.1",
-    "scikit-learn==1.3.1",
-    "scikit-misc==0.2.0",
-    "psutil==5.9.5",
-    "datasets~=2.14",
+    "torch~=2.0",
+    "torchdata~=0.7",
+    "scikit-learn~=1.0",
+    "scikit-misc>=0.2",  # scikit-misc 0.3 dropped Python 3.8 support
```
bkmartinjr, line 47: FYI  @atolopko-czi

bkmartinjr (approved):

Issue #948, opened 2024-01-24
Add tests for the differential expression query output to ensure correctness of implementation


prathapsridharan:
The entrypoint to query for differential expression [compute_all()](https://github.com/chanzuckerberg/cellxgene-census/blob/atol/memento/epic/api/python/cellxgene_census/src/cellxgene_census/experimental/diffexp/memento/diff_expr.py#L63) does not have any tests to ensure that the differential expression outputs are correct.

1. _We definitely need unit tests_. But unit tests work on dummy data that are too small to capture the distribution of real data and therefore _cannot tell you if the outputs align with the expected output of the memento methodology_

2. _Another type of test is a regression test_. In this type of test you pick a set of characteristic queries that distinguish 2 populations to compare for differential expression. You will also need differential expression outputs of these 2 populations to serve as the truth (expected output) - faithful to the `memento` methodology. Then run the comparison against an implementation that is backed by real data. As of now, this means running the queries against a real estimator cube. You would then compare the differential expression output of the implementation against the "truth".

The regression test is more of an integration test and is complex to implement and likely expensive to run especially if there are performance bottlenecks.

PR #976, opened 2024-02-06
chore: Write unit test for differential expression computation

prathapsridharan:
- Add a unit/regression test reading off a real memento estimator cube
- See memento epic ticket #948 for details


api/python/cellxgene_census/tests/experimental/diffexp/memento/test_diff_expr.py
```diff
@@ -1,156 +1,97 @@
...
 
 
-@pytest.fixture(scope="function")
-def estimators_df(dim_counts: Dict[str, int]) -> pd.DataFrame:
-    """Create a dummy estimators DataFrame, with all combinations of variables."""
-    columns = CUBE_LOGICAL_DIMS_OBS + ["feature_id"]
-    dim_counts_full = {var: 1 for var in columns}
-    dim_counts_full.update(dim_counts or {})
-    dim_values = {var: [f"{var}_{chr(ord('a') + i)}" for i in range(n)] for var, n in dim_counts_full.items()}
+class TestDiffExprRealDataset:
```
atolopko-czi, line 11: why introduce a class? (doing so is non-idiomatic for pytest)

api/python/cellxgene_census/tests/experimental/diffexp/memento/test_diff_expr.py
```diff
@@ -1,156 +1,97 @@
...
+    using real data, it is strongly recommended that the functions under
+    test are all PUBLIC functions. Private functions are best encapsulated
+    in separate class.
+    """
 
-    for c, estimator in enumerate(["n_obs", "mean", "sem", "var", "selv"], start=1):
-        estimators[estimator] = range(1, len(estimators) + 1)
-        estimators[estimator] *= c
+    @pytest.fixture(scope="class", params=["test_case_1", "test_case_2"])
+    def test_cases_for_compute_all_fn(self, request: Any) -> Dict[str, Any]:
```
atolopko-czi, line 22: decorate all test functions with `@pytest.mark.experimental`, since these are all for experimental code (we probably should rethink this, since all tests are under /experimental dir, but it's the repo's convention for now).

api/python/cellxgene_census/tests/experimental/diffexp/memento/test_diff_expr.py
atolopko-czi: pytest supports classes, yes, but I don't often see them used with pytest.  anyway, it's just a nit! :)

atarashansky (approved):
LGTM!

api/python/cellxgene_census/tests/experimental/diffexp/memento/test_diff_expr.py
```diff
@@ -1,156 +1,97 @@
...
+            "test_case_2": {
+                "diff_exp_query": {
+                    "cube_path": estimator_cube_path,
+                    "query_filter": "tissue_general_ontology_term_id in ['UBERON:0001723'] and cell_type_ontology_term_id in ['CL:0000066', 'CL:0000057']",
+                    "treatment": "cell_type_ontology_term_id",
+                    "num_sampled_genes": 2,
+                },
+                "expected_diff_exp_result": [
+                    ("ENSG00000000419", 0.868715, 6.048411, 1.462810e-09),
+                    ("ENSG00000002330", 0.834346, 5.218739, 1.801458e-07),
```
atolopko-czi, line 65: Unless we're confident that these values are "correct", I would add an issue and corresponding comment to indicate the values require manual validation (by a comp bio person). I'm assuming we have plans to do another level of biological validation. Once the implementation is shown to be producing valid results, we should link out to the outcomes of the manual validation effort (a notebook, etc.) as "proof" of these data's correctness.

atolopko-czi (approved):
One comment about ensuring regression test expected out is actually _valid_, and how to communicate that current state. Other than that, LGTM!

api/python/cellxgene_census/tests/experimental/diffexp/memento/test_diff_expr.py
prathapsridharan: Thanks @atolopko-czi - Does this capture your suggestion? https://github.com/chanzuckerberg/cellxgene-census/issues/977

atolopko-czi: yup!

tools/models/memento/tests/test_cube_builder.py
```diff
@@ -15,13 +15,13 @@ def test_cube_builder_regression() -> None:
     cd tools/models/memento/
 
     python tests/fixtures/census_fixture.py \
-    s3://cellxgene-data-public/cell-census/2023-10-23/soma/census_data/homo_sapiens \
+    s3://cellxgene-data-public/cell-census/2023-10-30/soma/census_data/homo_sapiens \
```
ebezzi, line 18: Don't use `cellxgene-data-public` for tests and use one of the LTS versions, otherwise these won't be replicable.

prathapsridharan: @ebezzi - These are test fixtures generated from a census version (the date) and stored in the repo. So we can replicate test results. Moreover, my understanding from @atolopko-czi is that we do want to keep in sync with census schema changes. If we need to regenerate the fixture, then we might as well move to more recent census version to generate the fixture and deal with data/schema problems right then and there if the requirement is to keep reasonably up-to-date with census evolution.

Am I misunderstanding something?

atolopko-czi: Yes, that was the idea, @prathapsridharan. But the discussion does make me think that we should have 2 tests:  1) test to validate the diff expr computation, using (possibly older) Census data, and 2) an "acceptance" test that ensures diff expr runs on the latest census, but doesn't care about the correctness of computations, just that it runs to completion.    This regression test addresses (1). Note that "acceptance" tests are run prior to cutting new Census data and api releases, and we expect such tests to hit "real" data.

prathapsridharan: @ebezzi @atolopko-czi - Are y'all ok if we figure out what data version specific test types should depend on when work starts on this ticket: https://github.com/chanzuckerberg/cellxgene-census/issues/977

I think these are valid points but it doesn't need to hold up merging this PR

PR #978, opened 2024-02-07
Update copyright notice

ebezzi:


api/r/cellxgene.census/LICENSE.md
```diff
@@ -1,6 +1,6 @@
 # MIT License
 
-Copyright (c) 2023 Chan Zuckerberg Initiative Foundation
+Copyright (c) 2023-2024 Chan Zuckerberg Initiative Foundation
```
ebezzi, line 3: Not sure if we want to uniform this to 2022-2024. I guess the R APIs were released in 2023, but this could be confusing.

bkmartinjr (approved):

ebezzi:
Update: Wyatt is trying to figure out if we can adopt a universal convention for our copyrights. I'll update this PR according to his recommendations.

ebezzi:
After a discussion with Wyatt, he recommended the following options:
- Only the initial date (prevents us from updating the copyright every year)
- Only mention CZI and not CZIF (not relevant for the copyright)

We should apply those changes to all of our other repos for consistency.

PR #980, opened 2024-02-07
Remove upperbound on numpy

### LINKED ISSUE #979: Support NumPy >= 1.25
The numpy dependency is currently upperbounded with `<1.25`, which is causing dependency issues when attempting to install scvi-tools and cellxgene-census in the same environment. I believe the newest numba releases support numpy 1.25, so would it be possible to increase the upperbound here? Thanks.

martinkim0:
Closes #979

### DIFF
--- a/api/python/cellxgene_census/pyproject.toml
+++ b/api/python/cellxgene_census/pyproject.toml
@@ -33,7 +33,7 @@ dependencies= [
     # Make sure this version does not fall behind the builder's tiledbsoma version.
     "tiledbsoma~=1.7.0",
     "anndata",
-    "numpy>=1.21,<1.25",  # numpy is constrained by numba and the old pip solver
+    "numpy>=1.21",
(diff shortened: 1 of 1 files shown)


bkmartinjr (approved):
LGTM.  Will continue to install old versions of numpy on Python 3.8 due to upstream pin (in tiledbsoma), but Python 3.9+ will get latest.

bkmartinjr:
@martinkim0 - do you need this soon?  If so, we will need to drop a release containing the change.  Please let us know.

martinkim0:
Great, thanks for approving @bkmartinjr! Not very urgent, but it would be nice to have as soon as possible. Is there a schedule for the next release?

bkmartinjr:
> Is there a schedule

We do them as needed, and not on a regular schedule.  I'll raise the issue with the team and see if we can get it moving.

PR #984, opened 2024-02-09
[builder] fix string size overflow in Arrow

bkmartinjr:
Resolve array index overflow caused by single-cell-data/TileDB-SOMA#2120  This issue was triggered by the size of a Census `obs` column exceeding 2**31-1 bytes.

Specific changes:
* all obs/var string types are now Arrow `large_string`
* explicit schema constraint used in `Table.from_pandas` calls used to generate Arrow Tables for write

Side effect: this increases memory use as the `large_string` encoding uses int64 indices.


ebezzi (approved):
To ensure I got this right, the issue here is with TileDBSOMA aggregating more chunks during writes and failing if one of those chunks cannot fit in 2**31-1 bytes, correct? In that case, I will probably add a comment/link to the issue in the schema, otherwise one would think that the `large_string`s are necessary because single values can exceed the integer size (which is obviously impossible). Otherwise, LGTM.

bkmartinjr:
> To ensure I got this right

You mostly have it right.  The key concept with Arrow string/binary is that there are two limits:
* each string must be less than the index size limit (e.g., for `string`, any single string must be less than 2^31-1 in length)
* AND, the total array length (all strings) must be less than the same limit

In this case, we are continuing to add cells to obs, and the _total_ array size is exceeding 2^31-1 bytes when tiledbsoma flattens the chunked array.

This only affects the write path (census_builder) memory use.  The read path already defaults to `large_string` due to tiledbsoma behavior, but that is a transient overhead if you then convert to a Pandas DataFrame.

I will add a comment in the builder schema definition, as that is the only place it has an effect on behavior.

PR #995, opened 2024-02-13
[misc] add support for var embeddings to contrib curation tool

bkmartinjr:
Extend the embedding contrib ingestion tool to support var/gene encodings.

### DIFF
--- a/tools/census_contrib/src/census_contrib/load.py
+++ b/tools/census_contrib/src/census_contrib/load.py
@@ -11,7 +11,7 @@ import pyarrow as pa
 import tiledbsoma as soma
 from typing_extensions import Self
 
-from .census_util import get_obs_soma_joinids
+from .census_util import get_axis_soma_joinids
 from .config import Config
 from .util import blocksize, blockwise_axis0_tables, get_logger, has_blockwise_iterator, soma_context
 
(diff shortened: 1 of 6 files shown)


tools/census_contrib/src/census_contrib/load.py
```diff
@@ -126,19 +126,19 @@ def __enter__(self) -> Self:
         self.joinids_sort = np.argsort(self.joinids)
         self.embeddings = np.load(self.embeddings_uri, mmap_mode="r")
 
-        self.n_obs = len(self.joinids)
+        self.n_embeddings = len(self.joinids)
```
ebezzi, line 129: Nit: I feel like `n_embeddings` could be a bit confusing since embedding also refers to the matrix as a whole. Maybe `n_rows`?

ebezzi (approved):

tools/census_contrib/src/census_contrib/load.py
bkmartinjr: each row _is_ an embedding of one cell into the latent space.  Hence the name.

ebezzi: Yeah, the name is indeed correct, but the term `embeddings` is overloaded and I feel like it might confuse the reader. That said, I don't think it really matters, so I'll resolve the conversation.

PR #999, opened 2024-02-16
feat: 'native cell' -> 'unknown' on geneformer label_blocklist -- DO NOT MERGE (yet)

### LINKED ISSUE #1019: [misc] `native cell` obsolescence (GeneFormer config)
In the near future CxG schema update 5.0, use of `native cell` (CL:0000003) in `cell_type_ontology_term_id` will be replaced with the non-term string `unknown`.

GeneFormer config is currently the only place this seems to be used in the Census code base.  As `native cell` should no longer appear, it may be feasible to just add `unknown` to the same config.

CC: @brianraymor @danieljhegeman 

See also chanzuckerberg/single-cell#647

danieljhegeman:
Fixes #1019

Schema 5.0 will use an updated ontology that does not include `CL:0000003 - native cell`. We will accept instead `"unknown"` for the `cell_type_ontology_term_id`, and accordingly we will annotate `cell_type` (label) as `"unknown"`.

This should (probably) be merged once we migrate. @mlin let me know if you foresee any issues.

### DIFF
--- a/tools/models/geneformer/finetune-geneformer.config.yml
+++ b/tools/models/geneformer/finetune-geneformer.config.yml
@@ -2,7 +2,7 @@
 label_feature: cell_subclass
 # Specific labels to exclude from training and evaluation
 label_blocklist:
-  - native cell
+  - unknown
 # Also exclude labels with too few examples
 label_min_examples: 10
 # Fraction of the input Dataset to hold out for evaluation


metakuni:
@danieljhegeman , I connected this PR to:
* https://github.com/chanzuckerberg/cellxgene-census/issues/1019

mlin (approved):

PR #1002, opened 2024-02-17
[docs] Fix docstrings for various functions

martinkim0:
Changes include fixing some typos, adding or removing whitespaces for correct rendering of docs, and changing `get_census_version_directory` to use `Args:` instead of `Params:`.


martinkim0:
For whoever reviews this: is there a particular reason that intersphinx mappings are not used in some parts of the docs such as [here](https://github.com/chanzuckerberg/cellxgene-census/blob/main/api/python/cellxgene_census/src/cellxgene_census/_open.py#L97)? Otherwise, I can add changes for those in this PR too.

bkmartinjr:
> particular reason that intersphinx mappings are not used

To my knowledge, lack of clue on our part and lack of a linter in our CI.  

Very happy to take the PR changes.

martinkim0:
@bkmartinjr Thanks for taking a look! I'll add the additional changes

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
```diff
@@ -29,20 +29,20 @@ def get_anndata(
     column_names: Optional[soma.AxisColumnNames] = None,
 ) -> anndata.AnnData:
     """
-    Convience wrapper around ``soma.Experiment`` query, to build and execute a query,
+    Convenience wrapper around :class:`tiledbsoma.Experiment` query, to build and execute a query,
```
bkmartinjr, line 32: recommend consistently using `soma.` or `tiledsoma.`

We tend to import as `soma`, but either is fine as long as it is consistent

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -352,14 +352,14 @@ class ExperimentDataPipe(pipes.IterDataPipe[Dataset[ObsAndXDatum]]):  # type: ig
                  [2416,    0,    4],
                  [2417,    0,    3]], dtype=torch.int64))
 
-    The ``return_sparse_X`` parameter controls whether the ``X`` data is returned as a dense or sparse Tensor.  If the
-    model supports use of sparse Tensors, this will reduce memory usage.
+    The ``return_sparse_X`` parameter controls whether the ``X`` data is returned as a dense or sparse
+    :class:`torch.Tensor`.  If the model supports use of sparse :class:`torch.Tensor`s, this will reduce memory usage.
```
bkmartinjr, line 356: Extra space after period in:  `.  If the`

api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -83,18 +83,19 @@ def _open_soma(
 
 
 def get_default_soma_context(tiledb_config: Optional[Dict[str, Any]] = None) -> soma.options.SOMATileDBContext:
-    """Return a ``SOMATileDBContext`` with sensible defaults that can be further customized by the user.  The customized
-    context can then be passed to ``open_soma(context=...)`` or a ``SOMAObject.open(context=...)`` method,
-    such as ``tiledbsoma.Experiment.open()``.  Use the ``replace()`` method on the returned object to customize its
-    settings further.
+    """Return a :class:``tiledbsoma.SOMATileDBContext` with sensible defaults that can be further customized by the
+    user.  The customized context can then be passed to :func:``cellxgene_census.open_soma`` with the ``context``
```
bkmartinjr, line 87: double space after period

bkmartinjr (changes requested):
A couple of very tiny nits - otherwise looks good.

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
martinkim0: I opted for `tiledbsoma.` as I wasn't sure if `soma` would render correctly since the package is called `tiledbsoma`.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
martinkim0: Addressed

api/python/cellxgene_census/src/cellxgene_census/_open.py
martinkim0: Addressed

bkmartinjr (approved):

bkmartinjr:
BTW, if you know of a linter that would catch missing formatting (e.g., single quotes vs double), happy to add it to the CI

martinkim0:
We use the Ruff linter on our end - I can put in a PR for that if that works for you!

bkmartinjr:
We have ruff in our pre-commits.  But at least as configured it doesn't flag this stuff for us.  Happy to take a PR!!  Many thanks

PR #1020, opened 2024-02-23
[docs] misc docsite adjustments

ebezzi:
- Fixes intersphinx bindings for tiledbsoma
- Adds more intersphinx bindings (scanpy, torch, etc)
- Fix wrong formatting
- Other misc fixes (typos, dates, etc).
- Updates copyright

This has been tested locally and the docsite renders correctly.

### DIFF
--- a/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
+++ b/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
@@ -327,7 +327,7 @@ class _ObsAndXIterator(Iterator[ObsAndXDatum]):
 
 
 class ExperimentDataPipe(pipes.IterDataPipe[Dataset[ObsAndXDatum]]):  # type: ignore
-    """An :class:`torchdata.datapipes.iter.IterDataPipe` that reads ``obs`` and ``X`` data from a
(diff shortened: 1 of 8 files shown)


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -436,7 +436,7 @@ def __init__(
             return_sparse_X:
                 Controls whether the ``X`` data is returned as a dense or sparse :class:`torch.Tensor`. As ``X`` data is
```
bkmartinjr, line 437: can you clarify when double-backquote vs. single-backquote is desirable?  I see double here, single elsewhere....

I assume double is emphasis (ie.., translate to a single in the final markdown), and single is intersphinx literal?

bkmartinjr (approved):

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ebezzi: With our current Sphinx setup:
1. Normal single backticks will render as _italic_
2. Double backticks will render as `code`
3. Any time you use references (like `:class`), you need to use single backticks.

PR #1021, opened 2024-02-26
[docs] Switch R docsite to custom gh runner

ebezzi:
R docsite generation now runs on GHA again using a runner with enough RAM.


.github/workflows/docsite-build-deploy.yml
```diff
@@ -33,19 +33,39 @@ jobs:
...
+          cd ../../../
           mkdir -p docsite/r
           cp -r api/r/cellxgene.census/docs/* docsite/r/.
 
       - name: Deploy 🚀
         uses: JamesIves/github-pages-deploy-action@v4
         with:
           folder: docsite # The folder the action should deploy.
-          branch: gh-pages
+          branch: gh-pages-test
```
ebezzi, line 71:
```suggestion
          branch: gh-pages
```

.github/workflows/docsite-build-deploy.yml
```diff
@@ -9,7 +9,7 @@ permissions:
 jobs:
   build-and-deploy:
     concurrency: ci-${{ github.ref }}
-    runs-on: ubuntu-latest
+    runs-on: single-cell-1tb-runner
```
bkmartinjr, line 12: Naive question:  why do we need such a large instance type to generate docs?  Would `single-cell-8c64g-runner` suffice (I believe it uses the same base image, but on a cheaper/smaller instance type)

bkmartinjr (approved):
Couple of optional small items noted inline. At your discretion.

.github/workflows/docsite-build-deploy.yml
ebezzi: It's actually a good question - I think historically some of the vignettes were very memory hungry, hence the 1T choice, but it's worth giving it a try with the 64G.

bkmartinjr: I converted several to the small runner, to save $$

ebezzi: Switched to `single-cell-8c64g-runner` and doing a test run right now.

api/r/cellxgene.census/vignettes/census_citation_generation.Rmd
```diff
@@ -96,15 +96,15 @@ print(slice_datasets$citation)
 ```{r,message=FALSE,warning=FALSE}
 # Fetch a Seurat object
 sce_obj <- get_single_cell_experiment(
-    census = census,
-    organism = "homo_sapiens",
-    measurement_name = "RNA",
-    obs_value_filter = "tissue == 'cardiac atrium'",
-    var_value_filter = "feature_name == 'MYBPC3'",
-    obs_column_names = c("dataset_id", "cell_type")
+  census = census,
```
pablo-gar, line 99: Any particular reason we are removing indentation here?

.github/workflows/docsite-build-deploy.yml
ebezzi: It worked! Leaving the change.

mlin (commented):
@ebezzi This is great, thank you!!!! One thing, I think you can delete vignettes/README.md and update_docs.sh. Delighted to be rid of that stuff of course!

ebezzi:
> @ebezzi This is great, thank you!!!! One thing, I think you can delete vignettes/README.md and update_docs.sh. Delighted to be rid of that stuff of course!

Good point, I also updated `release_process.md` and removed the outdated steps.

PR #1023, opened 2024-02-28
[python] New embeddings API

ebezzi:
New embeddings API. Provides a unified access pattern to embeddings (regardless of whether they're collaboration or hosted) through `get_anndata`.


api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
```diff
@@ -22,6 +22,7 @@ def get_anndata(
     X_name: str = "raw",
     X_layers: Optional[Sequence[str]] = (),
     obsm_layers: Optional[Sequence[str]] = (),
```
ebezzi, line 24: Two options:
1. Reuse `obsm_layers` (and `varm_layers`, that needs to be added) and modify the behavior so they retrieve the suitable embedding. This would effectively avoid introducing a breaking change
2. Rename the field, thus introducing a breaking change. 

If we don't plan to re-introduce a "mixed" scenario where we have both obsm objects in the census and external embeddings, I recommend 1.

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding.py
```diff
@@ -136,3 +136,48 @@ def get_embedding(
...
+
+def get_all_available_embeddings(census_version: str) -> list[dict[str, Any]]:
+    """Return a dictionary of all available embeddings for a given Census version.
+
+    Args:
+        census_version:
+            The Census version tag, e.g., ``"2023-12-15"``.
+
+    Returns:
+        A list of dictionaries, each containing metadata describing an available embedding.
```
ebezzi, line 149: Options:
1. Return a subset of the metadata that only has relevant information (name, organism, etc). The example listed here is only for reference
2. Return the full metadata.

tools/census_contrib/src/census_contrib/metadata.py
```diff
@@ -140,6 +142,13 @@ def validate_metadata(args: Arguments, metadata: EmbeddingMetadata) -> Embedding
             "Metadata: description must be string between 1 and {MAX_DESCRIPTION_LENGTH} characters in length",
         )
 
+    # 7. Name must have length < 24 characters
+    MAX_NAME_LENGTH = 24
```
ebezzi, line 146: I figured this could be a sensible value.

tools/census_contrib/embedding_metadata.md
```diff
@@ -21,6 +21,7 @@ Each embedding will contain a variety of metadata stored in the SOMA `metadata`
 | Field name             | Required | Type          | Description                                                                           |
 | ---------------------- | -------- | ------------- | ------------------------------------------------------------------------------------- |
 | id                     | required | string        | CZI-assigned accession ID for this embedding                                          |
+| embedding_name         | required | string        | Name of the algorithm used to generate the embedding                                  |
```
ebezzi, line 24: If we keep this as `required`, we should also modify the metadata in the existing embeddings, other than `contributions.json`. Otherwise, we can leave this as optional and only define it where we needed. Easier but not as clean. @bkmartinjr thoughts?

bkmartinjr: I would make it required for simplicity.  it is also quite easy to manually patch the array metadata if that is desired.

I would add to the schema a note that this value is NOT guaranteed to be unique across all embeddings (only the ID is guaranteed unique).

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
bkmartinjr: I don't like 1, as it is extremely confusing in a (likely future) world where we have both types.

I'd keep it super simple:  a param that says "add contributed embedding with algo name X".  You might also consider accepting the accession ID, as that resolves the case where there are non-unique algo names

tools/census_contrib/src/census_contrib/metadata.py
bkmartinjr: any reason to have it so short?  Unless there is an external dependency on 24, I'd go for something that can handle a long project name (e,g, < 128)

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
ebezzi: Yeah, I agree it could cause issues in the future. Let's go for the new param, unless @pablo-gar disagrees. I agree that accession id should be a valid choice. I'll add that to the docstrings.

tools/census_contrib/src/census_contrib/metadata.py
ebezzi: Nah, we can leave it to 128.

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
ebezzi: Added `add_obs_embeddings` and `add_var_embeddings`. Do we want to leave `obsm_layers` (and for completeness, add `varm_layers`)? The main concern is that right now, they would pull the "bogus" scvi embeddings so if we do, we should either make them no-op or add a warning. @pablo-gar any thoughts?

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
```diff
@@ -75,14 +92,31 @@ def get_anndata(
...
-        return query.to_anndata(
+        adata = query.to_anndata(
             X_name=X_name,
             column_names=column_names,
             X_layers=X_layers,
             obsm_layers=obsm_layers,
+            varm_layers=varm_layers
         )
+
+        # If add_obs_embeddings or add_var_embeddings are defined, inject them in the appropriate slot
```
ebezzi, line 109: `add_var_embeddings` TBA. Need to add `var` support in get_embeddings before we do it.

api/python/cellxgene_census/src/cellxgene_census/_util.py
```diff
@@ -18,3 +19,7 @@ def _uri_join(base: str, url: str) -> str:
         p_url.fragment,
     ]
     return urllib.parse.urlunparse(parts)
+
+def _extract_census_version(census: soma.Collection):
```
ebezzi, line 23: I created a live corpus unit test for this method. This should ensure that this parsing method remains consistent across releases.

api/python/cellxgene_census/tests/test_get_anndata.py
```diff
@@ -195,3 +193,21 @@ def test_get_anndata_obsm_two_layers(census: soma.Collection, obsm_layers: List[
     for obsm_layer in obsm_layers:
         assert obsm_layer in ad.obsm.keys()
         assert ad.obsm[obsm_layer].shape[0] == 100
+
+@pytest.mark.live_corpus
+@pytest.mark.parametrize("add_obs_embeddings", [["scvi", "geneformer"]])
+def test_get_anndata_add_obs_embeddings(census: soma.Collection, add_obs_embeddings: List[str]) -> None:
```
ebezzi, line 199: Test won't pass until we add `embedding_name` to the manifest file, as well as the new embeddings. Also add different organisms.

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
```diff
@@ -75,14 +92,31 @@ def get_anndata(
...
+        if add_obs_embeddings or add_var_embeddings:
+            obs_soma_joinids = query.obs_joinids()
+            from cellxgene_census.experimental import get_embedding, get_embedding_metadata_by_name
+            census_version = _extract_census_version(census)
+            for emb in add_obs_embeddings:
+                emb_metadata = get_embedding_metadata_by_name(emb, organism, census_version, "obs_embedding")
+                uri = f"{CENSUS_EMBEDDINGS_LOCATION_BASE_URI}/{census_version}/{emb_metadata['id']}"
+                embedding = get_embedding(census_version, uri, obs_soma_joinids)
```
ebezzi, line 117: Note: this will cause the census object to be re-opened. While this shouldn't be an issue, it will result into an extra call. With some effort I can refactor `get_embedding` to also accept an existing Census object, but I'm not sure if it's worth it.

bkmartinjr: IMHO, you should refactor the code to have a (common, shared) function that accepts an already open Census handle

api/python/cellxgene_census/src/cellxgene_census/_util.py
bkmartinjr: this code has some lint - please run (an up to date) pre-commit across it

bkmartinjr (commented):
No real concerns with functionality, but suggest you do an API review with Pablo and Mike to figure out if the signatures are friendly/comprehensible from a UX perspective

metakuni:
> No real concerns with functionality, but suggest you do an API review with Pablo and Mike to figure out if the signatures are friendly/comprehensible from a UX perspective

metakuni:
Sorry, fat fingers clicked the wrong button!

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
ebezzi: Done.

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
```diff
@@ -58,12 +64,20 @@ def get_anndata(
...
+            Embeddings to be returned as part of the ``varm`` slot.
+            Use :func:`get_all_available_embeddings` to retrieve available embeddings
+            for this Census version and organism.
 
     Returns:
         An :class:`anndata.AnnData` object containing the census slice.
 
     Lifecycle:
-        maturing
+        experimental
```
ebezzi, line 80: Reverting this to `experimental`.

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
ebezzi: I added a condition where `obsm_layers` and `add_obs_embedding` cannot share the same value. @atarashansky any idea on a better name for `add_obs_embedding` (from a user perspective)?

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
```diff
@@ -58,12 +64,20 @@ def get_anndata(
...
         obsm_layers:
             Additional obsm layers to read and return in the ``obsm`` slot.
+        add_obs_embeddings:
+            Additional embeddings to be returned as part of the ``obsm`` slot.
+            Use :func:`get_all_available_embeddings` to retrieve available embeddings
+            for this Census version and organism.
+        add_var_embeddings:
+            Additional embeddings to be returned as part of the ``varm`` slot.
+            Use :func:`get_all_available_embeddings` to retrieve available embeddings
+            for this Census version and organism.
```
bkmartinjr, line 74: Naming seems off.  When you want to include/add/fetch an X or obsm layer, you use `X_layers` and `obsm_layers`, but for an embedding it is `add_obs_embeddings` (i.e., includes a verb).  Why not stick with the consistent schema, a la `obs_embeddings` and `var_embeddings`?  or alternatively `obs_contributed_embeddings`, ....?

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
```diff
@@ -75,14 +89,46 @@ def get_anndata(
...
+        if add_obs_embeddings or add_var_embeddings:
+            from .experimental._embedding import _get_embedding, get_embedding_metadata_by_name
+
+            census_version = _extract_census_version(census)
+            experiment_name = _get_experiment_name(organism)
+            census_directory = get_census_version_directory()
+
+            if add_obs_embeddings:
+                if obsm_layers and [x for x in add_obs_embeddings if x in obsm_layers]:
+                    raise ValueError(
```
bkmartinjr, line 115: Given the (high) cost of calling query.to_anndata(), you should do all error checking before the costly ops - ie., move this kind of stuff to a prologue

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
```diff
@@ -75,14 +89,46 @@ def get_anndata(
...
+                        "Cannot request both `obsm_layers` and `add_obs_embeddings` for the same embedding name"
+                    )
+
+                obs_soma_joinids = query.obs_joinids()
+                for emb in add_obs_embeddings:
+                    emb_metadata = get_embedding_metadata_by_name(emb, experiment_name, census_version, "obs_embedding")
+                    uri = f"{CENSUS_EMBEDDINGS_LOCATION_BASE_URI}/{census_version}/{emb_metadata['id']}"
```
bkmartinjr, line 122: shouldn't these use urljoin()?

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
ebezzi: Good point.

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
```diff
@@ -75,14 +89,46 @@ def get_anndata(
...
+        # If add_obs_embeddings or add_var_embeddings are defined, inject them in the appropriate slot
+        if add_obs_embeddings or add_var_embeddings:
+            from .experimental._embedding import _get_embedding, get_embedding_metadata_by_name
+
+            census_version = _extract_census_version(census)
+            experiment_name = _get_experiment_name(organism)
+            census_directory = get_census_version_directory()
+
+            if add_obs_embeddings:
+                if obsm_layers and [x for x in add_obs_embeddings if x in obsm_layers]:
```
bkmartinjr, line 114: this error check (name collision) is missing from the var axis.  Seems like you need it for both

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
ebezzi: `obs_embeddings` seems fine. `obs_contributed_embeddings` could work but it might be confusing since it is also used for collaborations embeddings. Maybe it's worth adding a paragraph at the beginning with a quick explanation on what these embeddings are?

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
ebezzi: I removed `varm_layers` from the arguments, as you suggested, since it is not in the current API. The only way you can request a varm is through `add_varm_embeddings`

api/python/cellxgene_census/src/cellxgene_census/_util.py
```diff
@@ -18,3 +20,9 @@ def _uri_join(base: str, url: str) -> str:
         p_url.fragment,
     ]
     return urllib.parse.urlunparse(parts)
+
+
+def _extract_census_version(census: soma.Collection) -> str:
```
bkmartinjr, line 25: this doesn't handle errors very well.  Previously, the API worked fine for Census builds that were local (e.g., opened with the `uri=` param), and I suspect we _really_ don't want to break that as it rules out people taking local copies.

I'd give some thought to how you know if the "version" returned is nonsense?  (e.g., the SOMA Census objects are copied into /tmp/foobar).  At least you inspect it and see if it looks like a YYYY-MM-DD, and throw an error if not.

You could also sniff the `census.uri` prefix, and see if it is in an expected location (rather than file:///tmp/foobar/)

I think raising is perfectly fine here, as it just means that some edge cases can't infer the version from the URI, and therefore are unsupported (with a nice clean error saying so)

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding.py
```diff
@@ -91,48 +149,119 @@ def get_embedding(
...
+    embeddings = []
+    manifest = response.json()
+    for _, obj in manifest.items():
+        if obj["census_version"] == census_version:
+            embeddings.append(obj)
+
+    return embeddings
+
+
+def get_all_census_versions_with_embedding(
```
bkmartinjr, line 238: What is the rationale for adding these simple "filtered" variants of "get_all_embeddings"?  Personally, I think one simple "get all" function that the user can then filter easily (e.g.  a list of dicts or whatever) is preferable as it covers more use cases, and is simpler to maintain.

api/python/cellxgene_census/src/cellxgene_census/_util.py
ebezzi: I also think raising is fine. This is an edge case because right now this is only called in `get_anndata` when adding the embeddings and nowhere else. In this case, we can't guarantee that the local copy of the Census matches the embeddings, therefore raising is the only option.

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding.py
```diff
@@ -91,48 +149,119 @@ def get_embedding(
...
+        embedding_type:
+            The type of embedding. Defaults to "obs_embedding".
+
+    Returns:
+        A list of census versions that contain the specified embedding.
+    """
+    response = requests.get(CELL_CENSUS_EMBEDDINGS_MANIFEST_URL)
+    response.raise_for_status()
+
+    versions = set()
```
bkmartinjr, line 257: this whole blob could be a simple comprehension which might make it more pythonic (this is a nit, up to you). E.g,
```
return sorted({ obj['census_version'] for obj in manifest.values() if ...  })
```

And unless there are duplicates expected, I'm not sure what the set adds?  If there are duplicates, doesn't that imply you need more filter criteria?

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
bkmartinjr: Right - gotcha!

bkmartinjr: So, I'm thinking about this previous conversation a bit more.  Coming around to a slightly different perspective.  Given:
* tiledbsoma/somacore support more functionality than is exposed here (e.g., varp_layers, etc)
* Census doesn't currently use these, due to _schema_ definition (not code)
* we want to clearly separate the arg names for clarity (the previous conversation where we decided to remove `varm_layers`)

I think we should land roughly here, to keep the code & schema modular:
* this function (`get_anndata`) should pass through all (or at least most) arguments supported by ExperimentAxisQuery.to_anndata.  That makes it future-proof and decoupled from the schema
* the newly added args should have clearly separated names (the current names are not clear per above comment)
* the error checking needs to detect collisions because AnnData only has one "dict" to shove them all in, and they might conflict.

Boiling this down, I suggest:
1. Add (and pass to query.to_anndata) the args `obsm_layers`, `varm_layers`, `obsp_layers`, and `varp_layers`.  These are just pass through.
2. Rename `add_obs_embeddings` and `add_var_embeddings` to something clear (see above discussion)
3. Do the error checks for collisions, and do them before you do any data loading.  Example below.

Example error check for obs:
```
if set(obsm_layers) & set(add_obs_embeddings):
    ... there is a collision error ...

....do same for varm ...
```

ebezzi: Sounds good.

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding.py
ebezzi: These methods were requested by @pablo-gar . I'll defer the decision to him.

api/python/cellxgene_census/src/cellxgene_census/_get_anndata.py
pablo-gar: `obs_embeddings` and `var_embeddings` are good names.

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding.py
pablo-gar: strongly prefer 1) and having a `verbose` argument

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding.py
pablo-gar: Yes, a convenience helper API goes a long way in this space

pablo-gar (approved):
@ebezzi have you considered the convenience API to get all available Census version for a given embedding name?

ebezzi:
> @ebezzi have you considered the convenience API to get all available Census version for a given embedding name?

@pablo-gar `get_all_census_versions_with_embedding` that would be the function.

Issue #839, opened 2023-11-02
[Feature request] Python 3.12 support


bkmartinjr:
Add support for Python 3.12 (release Oct 2023).

Several upstream dependencies will need 3.12 support before this can land

jdblischak:
I was investigating the support for Python 3.12 today. Some notes:

* TileDB-SOMA supports Python 3.12 (https://github.com/single-cell-data/TileDB-SOMA/issues/1849, https://github.com/single-cell-data/TileDB-SOMA/pull/2108)
* It appears that cellxgene-census supports Python 3.12 on Linux but not macOS (https://github.com/chanzuckerberg/cellxgene-census/pull/1189)

https://github.com/chanzuckerberg/cellxgene-census/blob/f32c2748e9664cc196bb1758acb22f5a460fe188/.github/workflows/py-dependency-check.yml#L26-L31

PR #1046, opened 2024-03-11
Prepare for 1.11.0 release

ebezzi:
1. Update the `tiledb` dependency for API to ~1.8.0
2. Update R's `DESCRIPTION` file

R docs do not need to be regenerated as they're now part of the CI/CD.


api/python/cellxgene_census/pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies= [
     # NOTE: the tiledbsoma version must be >= to the version used in the Census builder, to
     # ensure that the assets are readable (tiledbsoma supports backward compatible reading).
     # Make sure this version does not fall behind the builder's tiledbsoma version.
-    "tiledbsoma~=1.7.0",
+    "tiledbsoma~=1.8.0",
```
bkmartinjr, line 34: I believe tiledbsoma==1.8.0 supports python 3.12.  Do you want to support 3.12 in this release as well, or does that require additional effort?

bkmartinjr (approved):
One question regarding python 3.12 support, but otherwise LGTM

api/python/cellxgene_census/pyproject.toml
ebezzi: Looks like there are some issues with Python 3.12 in tiledbsoma so the support is marked as provisional: https://github.com/single-cell-data/TileDB-SOMA/issues/1849.

So I think it's safe to leave it out from the Census for now. I'll ask John to let us know when it's fully supported.

bkmartinjr: sounds good. Also cross-referencing #839

PR #1061, opened 2024-03-21
[misc] Prepare for 1.12.0 release

prathapsridharan:
This release of the census package is being done to include an important update for `tiledbsoma` in `1.9.0` which includes a fix to categorical columns. To ensure that this release goes smoothly, we will:

1. Run an acceptance test off `TileDB-SOMA` `main` [branch](https://github.com/single-cell-data/TileDB-SOMA/tree/main) which contains the fixes that will go into `1.9.0`. We can think of this as a _pre-release acceptance test_
2. After (1) passes and the tiledb folks officially release `tiledbsoma==1.9.0`, we will update the dependencies in this PR to point to `tiledbsoma==1.9.0` and do the official release of `cellxgene-census==1.12.0`

### DIFF
--- a/api/python/cellxgene_census/pyproject.toml
+++ b/api/python/cellxgene_census/pyproject.toml
@@ -31,6 +31,7 @@ dependencies= [
     # NOTE: the tiledbsoma version must be >= to the version used in the Census builder, to
     # ensure that the assets are readable (tiledbsoma supports backward compatible reading).
     # Make sure this version does not fall behind the builder's tiledbsoma version.
+    # TODO: When ready to release 1.12.0 update tiledbsoma version to 1.9.0 and delete this temporary comment 
     "tiledbsoma~=1.8.0",
     "anndata",
     "numpy>=1.21",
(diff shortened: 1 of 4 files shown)


bkmartinjr (approved):

api/python/cellxgene_census/pyproject.toml
```diff
@@ -31,7 +31,8 @@ dependencies= [
     # NOTE: the tiledbsoma version must be >= to the version used in the Census builder, to
     # ensure that the assets are readable (tiledbsoma supports backward compatible reading).
     # Make sure this version does not fall behind the builder's tiledbsoma version.
-    "tiledbsoma~=1.8.0",
+    # TODO: When ready to release 1.12.0 update tiledbsoma version to 1.9.x and delete this temporary comment
```
bkmartinjr, line 34: time to delete the comment?

BTW, I believe `~=1.9.1` means anything >1.9.1 and <1.10.0, so this pin should be fine (1.9.0 never existed in the wild)

bkmartinjr (approved):
One minor point/nit, but otherwise LGTM

api/python/cellxgene_census/pyproject.toml
prathapsridharan: will remove comment.

> I believe ~=1.9.1

Yea that is my understanding too

PR #1066, opened 2024-03-26
[python] fix hvg_vs_scanpy test

ebezzi:
Due to [this change](https://github.com/scverse/scanpy/commit/82f84129ed895dfdded3111b98aa4dc2ae5d4887#diff-a3319950a357c8d24de0f78d7e64852aa3c8a5b59d3a633e960d088c86528f53R137) in Scanpy's implementation, there is now a new `gene_name` field that gets added to the returned dataframe. This PR adds the same field to our implementation.

### DIFF
--- a/api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_highly_variable_genes.py
+++ b/api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_highly_variable_genes.py
@@ -206,7 +206,6 @@ def _highly_variable_genes_seurat_v3(
     median_ranked = np.ma.median(ma_ranked, axis=0).filled(np.nan)  # type: ignore[no-untyped-call]
 
     var_df = var_df.assign(
-        gene_name=var_df.index,
         highly_variable_nbatches=pd.Series(num_batches_high_var, index=var_df.index),
         highly_variable_rank=pd.Series(median_ranked, index=var_df.index),
         variances_norm=pd.Series(np.mean(norm_gene_vars, axis=0), index=var_df.index),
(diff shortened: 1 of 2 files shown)


api/python/cellxgene_census/src/cellxgene_census/experimental/pp/_highly_variable_genes.py
```diff
@@ -206,6 +206,7 @@ def _highly_variable_genes_seurat_v3(
     median_ranked = np.ma.median(ma_ranked, axis=0).filled(np.nan)  # type: ignore[no-untyped-call]
 
     var_df = var_df.assign(
+        gene_name=var_df.index,
```
bkmartinjr, line 209: I don't quite follow this code.  It appears to assign the soma_joinid of each var column to the `gene_name`.  While this puts something in the column, it doesn't really follow the spirit of the ScanPy implementation or our CxG schema.

In the CxG schema, the var index contains the ENSEMBL ID of the gene (aka `feature_id` in the Census var dataframe). This is similar to most (*not* enforced) AnnData usage.

For ease of use, I argue that we should assign `gene_name` to be the `feature_id` value, not the `soma_joinid`.  You can look up feature_id from soma_joinid (and vice versa), but the feature_id is more likely what the user expects.

This is an API design issue for @pablo-gar and @hthomas-czi

bkmartinjr (changes requested):
I don't think the contents of `gene_name` are correct - see inline comment

api/python/cellxgene_census/tests/experimental/pp/test_hvg.py
```diff
@@ -117,7 +117,7 @@ def test_hvg_vs_scanpy(
     scanpy_hvg.index.name = "soma_joinid"
     scanpy_hvg.index = scanpy_hvg.index.astype(int)
     assert len(scanpy_hvg) == len(hvg)
-    assert all(scanpy_hvg.keys() == hvg.keys())
+    assert [k for k in scanpy_hvg.keys() if k != "gene_name"] == list(hvg.keys())
```
bkmartinjr, line 120: more pythonic to do something like:

> assert set(scanpy_hvg.drop('gene_name').keys()) == set(hvg.keys())

this also removes the column order dependency that will trip up the above code at some point

PR #1067, opened 2024-03-27
[builder] improve unit tests

bkmartinjr:
Changes:
* Add additional unit tests and fix one small bug found by those new tests
* Enable coverage to be collected more easily, by running tests in a single process (i.e., change to Dask cluster config for tests)
* Run with and without Numba JIT, to fully capture coverage _and_ continue testing in the as-used mode (numba enabled).
* Reduce memory budget for consolidation to allow testing in small instances (does not affect normal Census build)
* Additional unit tests to cover float precision reduction in more detail
* Update to latest tiledbsoma version
* Fix bug in HVG uncovered by enabling categoricals in the Census build

This should significantly increase Builder coverage (some due to new tests, some due to more accurate code coverage)


tools/cellxgene_census_builder/tests/anndata/test_anndata.py
```diff
@@ -232,3 +237,180 @@ def test_AnnDataProxy_indexing(census_build_args: CensusBuildArgs, slices: Any)
 
         adata = adata[slc]
         original_X = original_X[slc]
+
+
+@pytest.mark.parametrize("h5ad_simple", ("csr", "csc", "dense"), indirect=True)
+def test_estimated_density(tmp_path: pathlib.Path, h5ad_simple: str) -> None:
+    with open_anndata(h5ad_simple, base_path=tmp_path.as_posix()) as ad:
+        density = ad.get_estimated_density()
+        assert density == ad[1:].get_estimated_density()
```
prathapsridharan, line 246: I don't quite understand the reasoning behind the assert in `line 246`:

1. Are you trying to check some character of the data where if the bulk of the density resides in row 0, then you want the assertion to raise? If that is the case, then density could be concentrated in some other row - then what? OR

2. Are you trying to check the character of the implementation of `get_estimated_density()` function in some way? Perhaps that it doesn't fit to some strange non-linear function? I doubt that is your intent but just stating it since I am just guessing here.
3. Something else entirely?

prathapsridharan (approved):
LGTM. I just have a question about one of the tests. Thanks!

tools/cellxgene_census_builder/tests/anndata/test_anndata.py
bkmartinjr: I am trying to confirm that the estimated density is not affected by the choice of slice.  For example, if I pick a particularly dense slice, am I getting a different value than the average density for the entire file?

tools/cellxgene_census_builder/tests/anndata/test_anndata.py
```diff
@@ -232,3 +237,180 @@ def test_AnnDataProxy_indexing(census_build_args: CensusBuildArgs, slices: Any)
...
+    Conditions:
+    * the filter has a target species (e.g., NCBITaxon:9606, NCBITaxon:10090, etc)
+    * the obs and var axis species may have no matches, partial matches or all matches with the filter target
+
+    Expected results:
+    * when both axes are multi-species, return empty result
+    * when one or both axes are single-species, return slice that matches the filter
+
+    IMPORTANT: if the filter results in no cells (obs) matching, the result is empty on the
+    obs axes, and the var aixs IS NOT FILTERED (as there is no point to doing so). So test
```
ebezzi, line 390:
```suggestion
    obs axis, and the var axis IS NOT FILTERED (as there is no point to doing so). So test
```

ebezzi (approved):
LGTM. A typo in comments.

bkmartinjr:
@ebezzi @prathapsridharan - given the HVG fix in this PR, you may want to consider pushing a patch release once this lands.

PR #1069, opened 2024-03-29
Add support for custom encoders

ridvan-eksi:
Current version creates a label encoder by default for each string column queried. This PR extends this capability to provide custom encoders for columns including non-string types.

### DIFF
--- a/api/python/cellxgene_census/pyproject.toml
+++ b/api/python/cellxgene_census/pyproject.toml
@@ -31,7 +31,7 @@ dependencies= [
     # NOTE: the tiledbsoma version must be >= to the version used in the Census builder, to
     # ensure that the assets are readable (tiledbsoma supports backward compatible reading).
     # Make sure this version does not fall behind the builder's tiledbsoma version.
-    "tiledbsoma~=1.8.0",
+    "tiledbsoma~=1.9.1",
     "anndata",
     "numpy>=1.21",
     "requests",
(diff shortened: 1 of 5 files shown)


atolopko-czi:
@ridvan-eksi Thanks for the contribution!  Can you tell us a little more about how you're using the custom encoders for non-string columns? In case these non-numeric encoders might be useful to others, by default.

ridvan-eksi:
@atolopko-czi Main use case for users providing encoders would be to use fix mappings across different datapipe instances. In the current version, a new encoder is generated if the data release or query changes, but we would like to use consistent mappings even if the data or query expands. 

A second use would be using encoders other than one hot encoding for string columns, like ordinal encoder. Ordinal encoding is required for using learnable embeddings.

A third, and maybe a more hacky use case for these custom encoders would be allowing more complex functions rather than encoders to process non-string/string columns. This would be similar to transform concept in torchvision datasets and datapipes. For example, we might have a function to parse/extract some information from `development_stage` column. We can do that by providing the parse function as a custom encoder. 

I hope it's clear. As far as I can see, these encoder class should have a `transform` and `inverse_transform` methods which should be implemented in all custom encoders. (even though in some cases `inverse_transform` may not be possible)

atolopko-czi:
Reviving PR; we still intend to review and incorporate this change.

pablo-gar:
Thanks for PR @ridvan-eksi. All of those use cases are valuable. We'd like to incorporate the changes, and it looks like some error checking _may be_ needed and along some unit tests.

For the unit test, we'd welcome any suggestions you may have, otherwise we'll discuss this internally.

ivirshup:
@ridvan-eksi, if you're still interested in seeing this merged, would you mind merging main into this branch? Then we'll be able to see all the CI checks run.

It'd also be good to see a test for this, since at the moment it looks to me like any custom encoders passed in here would be overwritten during initialization.

ridvan-eksi:
Hi @ivirshup. I will merge main into this branch, and I will add a test. 

If I am not mistaken: If user provides a custom encoder for a specific column, that encoder is used for that column. For the remaining columns, the default encoder is used.

PR #1070, opened 2024-03-29
New embeddings API notebooks

ebezzi:
Notebooks for the new embedding API usage. 

[Notebook 1 preview](https://github.com/chanzuckerberg/cellxgene-census/blob/10aec4e4737c1c17baf3008ee42d99d706438977/api/python/notebooks/api_demo/census_access_maintained_embeddings.ipynb)
[Notebook 2 preview](https://github.com/chanzuckerberg/cellxgene-census/blob/10aec4e4737c1c17baf3008ee42d99d706438977/api/python/notebooks/api_demo/census_embedding.ipynb)

Notes:
1. I removed the `Loading embeddings into a Scipy matrices and NumPy arrays` part from Notebook 1, and I replaced it with `Load an embedding into a dense NumPy array`. Using sparse matrices isn't straightforward with the new API, since we don't expose direct access to the underlying SparseNDArray, and I don't think the example was particularly useful in the first place, since all the embeddings we have are essentially dense.
2. This PR also contains the docsite changes for the new convenience methods.


atarashansky (approved):
looks great!

pablo-gar (approved):
Looks to me, can you please fix this small item before merging?

-  Delete the last empty cell for `api/python/notebooks/api_demo/census_access_maintained_embeddings.ipynb`

PR #1072, opened 2024-04-01
[misc] turn off fast-fail for weekly CI checks

bkmartinjr:
The weekly checks that explore the full matrix of version/census build/etc. had GHA fast-fail enabled.  For these periodic checks, it is better to not fast-fail, and see which pats of the test matrix work and which are in error.  They are not run during the normal dev cycle, so there is little downside, and this will help diagnose actual issues faster.


prathapsridharan (approved):
LGTM

ebezzi (approved):

PR #1078, opened 2024-04-03
[R] Fix acceptance tests via gc()

### LINKED ISSUE #1060: [r] Add limits to acceptance tests in R
Acceptance tests in R require too much memory even on a 1TB runner now:

https://github.com/chanzuckerberg/cellxgene-census/actions/runs/8304274116/job/22729690851

Add coords bound (like in Python) to limit the amount of required memory.

ebezzi:
Fixes #1060 

Adding a call to gc() seems to fix the issue.


pablo-gar (changes requested):
Best practice and style is to use another `on.exit()` call after the `census$close()` one

```
on.exit(census$close(), add = TRUE)
on.exit(gc(verbose = TRUE, reset = FALSE, full = TRUE), add = TRUE)
```

see https://testthat.r-lib.org/articles/test-fixtures.html

pablo-gar (approved):

PR #1094, opened 2024-04-05
[builder] cellxgene_ontology_guide integration

ebezzi:
Integrate cellxgene_ontology_guide in place of the old TissueMapper.

### DIFF
--- a/tools/cellxgene_census_builder/pyproject.toml
+++ b/tools/cellxgene_census_builder/pyproject.toml
@@ -36,6 +36,7 @@ dependencies= [
     #    https://github.com/TileDB-Inc/TileDB/blob/dev/format_spec/FORMAT_SPEC.md
     "tiledbsoma==1.9.3",
     "cellxgene-census==1.12.0",
+    "cellxgene-ontology-guide==0.4.0",
     "scipy==1.12.0",
     "fsspec[http]==2024.3.1",
     "s3fs==2024.3.1",
--- a/tools/cellxgene_census_builder/pyproject.toml
+++ b/tools/cellxgene_census_builder/pyproject.toml
@@ -36,7 +36,7 @@ dependencies= [
     #    https://github.com/TileDB-Inc/TileDB/blob/dev/format_spec/FORMAT_SPEC.md
     "tiledbsoma==1.9.3",
     "cellxgene-census==1.12.0",
-    "cellxgene-ontology-guide==0.4.0",
+    "cellxgene-ontology-guide==0.6.1",
     "scipy==1.12.0",
     "fsspec[http]==2024.3.1",
     "s3fs==2024.3.1",
--- a/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/tissue_mapper.py
+++ b/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/tissue_mapper.py
@@ -1,117 +1,16 @@
 """NOTE: In the future, this code will be part of an ontology service library."""
 
-import owlready2
-
-from .globals import CXG_UBERON_ONTOLOGY_URL
+from cellxgene_ontology_guide import curated_ontology_term_lists
+from cellxgene_ontology_guide.entities import CuratedOntologyTermList
+from cellxgene_ontology_guide.ontology_parser import OntologyParser
 
 
 class TissueMapper:
-    # Name of anatomical structure, used to determine the set of ancestors for a given
-    # entity that we"re interested in.
-    ANATOMICAL_STRUCTURE_NAME = "UBERON_0000061"
-
-    # List of high level tissues, ORDER MATTERS. If for a given tissue there are multiple high-level tissues associated
-    # then `self.get_high_level_tissue()` returns the one that appears first in th this list
-    HIGH_LEVEL_TISSUES = [
-        "UBERON_0000178",  # blood
-        "UBERON_0002048",  # lung
-        "UBERON_0002106",  # spleen
-        "UBERON_0002371",  # bone marrow
-        "UBERON_0002107",  # liver
-        "UBERON_0002113",  # kidney
-        "UBERON_0000955",  # brain
-        "UBERON_0002240",  # spinal cord
-        "UBERON_0000310",  # breast
-        "UBERON_0000948",  # heart
-        "UBERON_0002097",  # skin of body
-        "UBERON_0000970",  # eye
-        "UBERON_0001264",  # pancreas
-        "UBERON_0001043",  # esophagus
-        "UBERON_0001155",  # colon
-        "UBERON_0000059",  # large intestine
-        "UBERON_0002108",  # small intestine
-        "UBERON_0000160",  # intestine
-        "UBERON_0000945",  # stomach
-        "UBERON_0001836",  # saliva
-        "UBERON_0001723",  # tongue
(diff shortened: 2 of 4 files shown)


tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/tissue_mapper.py
```diff
@@ -1,117 +1,16 @@
...
-        "Thing",
-        "UBERON_0000465",  # material anatomical entity
-        "UBERON_0001062",  # anatomical entity
-    ]
-
     def __init__(self) -> None:
-        self._cached_tissues: dict[str, str] = {}
-        self._cached_labels: dict[str, str] = {}
-        self._uberon = owlready2.get_ontology(CXG_UBERON_ONTOLOGY_URL).load()
+        self.ontology_parser = OntologyParser()
```
nayib-jose-gloria, line 10: you might want to pin to a particular schema version `OntologyParser("v5.0.0")` since the builder pins the version when it fetches the datasets from the index endpoint

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/tissue_mapper.py
```diff
@@ -1,117 +1,16 @@
 """NOTE: In the future, this code will be part of an ontology service library."""
```
nayib-jose-gloria, line 1: remove this?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/tissue_mapper.py
ebezzi: Does the version match the DP schema?

nayib-jose-gloria: yes, with a "v" prefix

tools/cellxgene_census_builder/pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies= [
     #    https://github.com/TileDB-Inc/TileDB/blob/dev/format_spec/FORMAT_SPEC.md
     "tiledbsoma==1.9.3",
     "cellxgene-census==1.12.0",
+    "cellxgene-ontology-guide==0.4.0",
```
Bento007, line 39: remove "owlready==0.44"

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/tissue_mapper.py
```diff
@@ -123,60 +22,19 @@ def get_high_level_tissue(self, tissue_ontology_term_id: str) -> str:
...
-
-        # List ancestors for this entity, including itself. Ignore any ancestors that
-        # are not descendents of UBERON_0000061 (anatomical structure).
-        ancestors = [entity.name]
-        branch_ancestors: list[str] = []
-        for is_a in entity.is_a:
-            branch_ancestors = self._list_ancestors(is_a, branch_ancestors)
-
-        # Include this branch of ancestors is under anatomical structure
-        if self.ANATOMICAL_STRUCTURE_NAME in branch_ancestors:
```
Bento007, line 148: Is this functionality covered by COG?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/tissue_mapper.py
```diff
@@ -193,53 +51,3 @@ def reformat_ontology_term_id(ontology_term_id: str, to_writable: bool = True) -
...
-            ancestors.append(entity.value.name.replace("obo.", ""))
-
-            # Check for ancestors of restriction
-            self._list_ancestors(entity.value, ancestors)
-            return ancestors
-
-        elif self._is_entity(entity) and not self._is_and_object(entity):
-            # Entity is a superclass, check for is_a relationships
-
-            if entity.name in self.DENY_LIST:
```
Bento007, line 220: COG doesn't have a deny list filter. Is this a problem?

Bento007 (commented):
COG should also be able to replace [ontology_mappy.py](https://github.com/chanzuckerberg/cellxgene-census/blob/e37605f86000d86e2a0f10a74f1e36d20fc9c0ba/tools/models/geneformer/helpers/ontology_mapper.py)

Bento007 (changes requested):
See comments

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/tissue_mapper.py
ebezzi: Not sure?

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/tissue_mapper.py
ebezzi: Most likely. I was expecting this function to be fully replaceable by `get_high_level_terms`, if it's not then I think we're not ready for using it in the Census.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/tissue_mapper.py
nayib-jose-gloria: this will not affect functionality. This is subsetting the ontology graph being built, but COG pre-builds the ontology graph and packages it in the library so it's not necessary, and is a superset of this graph. Census only matches terms to one of the terms in the curated list of tissue general terms anyway, so there isn't a concern around over-listing ancestors to users.

tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/tissue_mapper.py
nayib-jose-gloria: we already filter out non-UBERON terms. I don't foresee issues with the two UBERON terms in the deny list, since they are high-level terms closer to the root node than anything in the curated tissue general list, so skipping fetching their ancestors will not affect traversing the graph used for mapping to one of the curated tissue general terms. I believe this would only become an issue if a term higher-level than "material anatomical entity" or "anatomical entity" is added to the curated tissue list--which doesn't fit the use-case of the list, and further wouldn't make sense if the goal is to filter out ancestor terms of these terms

ebezzi:
> COG should also be able to replace [ontology_mappy.py](https://github.com/chanzuckerberg/cellxgene-census/blob/e37605f86000d86e2a0f10a74f1e36d20fc9c0ba/tools/models/geneformer/helpers/ontology_mapper.py)

This is lower priority as it's not part of the core builder. 

> COG should also be able to replace [ontology_mappy.py](https://github.com/chanzuckerberg/cellxgene-census/blob/e37605f86000d86e2a0f10a74f1e36d20fc9c0ba/tools/models/geneformer/helpers/ontology_mapper.py)

Are its functionalities covered by COG? This is only used by Geneformer, which isn't part of core Census, so I wouldn't be too worried about it.

Bento007 (commented):
Update to COG 0.6.1 and this PR is good to go.

Bento007 (approved):

prathapsridharan (approved):

PR #1098, opened 2024-04-11
[ci] Test new GH runners

ebezzi:
New Arc Runners cluster.


.github/workflows/py-unittests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   unit_tests_python_api:
     strategy:
       matrix:
-        os: [single-cell-8c64g-runner, macos-latest]
+        os: [sc-dev-census-64g-runner, macos-latest]
```
ebezzi, line 14:
```suggestion
        os: [sc-dev-64g-runner, macos-latest]
```

ivirshup (commented):
Should we also change this in:

* `docsite-build-deploy.yml`
*  `profiler.yml`
* `py-dependency-check.yml`
* `py-unittests.yml:64` (a little lower in this file, for the `unit_tests_builder`

.github/workflows/py-formatting.yml
```diff
@@ -13,7 +13,7 @@ env:
 jobs:
   pre_commit_checks:
     name: pre-commit checks
-    runs-on: [self-hosted, Linux, X64]
+    runs-on: [sc-dev-64g-runner]
```
ebezzi, line 16:
```suggestion
    runs-on: [self-hosted, Linux, X64]
```

.github/workflows/py-unittests.yml
```diff
@@ -96,7 +96,7 @@ jobs:
     needs:
       - unit_tests_builder
       - unit_tests_python_api
-    runs-on: [self-hosted, Linux, X64]
+    runs-on: [sc-dev-64g-runner]
```
ivirshup, line 99: This probably also doesn't need such a big machine

.github/workflows/docsite-build-deploy.yml
```diff
@@ -13,7 +13,7 @@ env:
 jobs:
   build-and-deploy:
     concurrency: ci-${{ github.ref }}
-    runs-on: single-cell-8c64g-runner
+    runs-on: sc-dev-64g-runner
```
ivirshup, line 16: This may also be do-able on a smaller machine, but I think okay to leave for now since we were already using a big one.

.github/workflows/py-unittests.yml
ivirshup:
```suggestion
    runs-on: [self-hosted, Linux, X64]
```

ivirshup (approved):

Issue #477, opened 2023-05-12
Add unit tests for PyTorch classes


atolopko-czi:

atolopko-czi:
Added unit tests for various batching cases, sparse/dense output, and encoders

prathapsridharan:
@pablo-gar (cc: @atolopko-czi @ebezzi ) - Is this intentionally left open to serve as a placeholder for unit test tasks for pytorch? I ask because the ticket is was created in _May 2023_.

If so, should we create tickets for specific test cases for `pytorch`? For example, we don't have any testing for the `scatter-gather-shuffle` algorithm - It is a non-deterministic algorithm and testing it needs some thought. Should we make a specific ticket for this case and link it to _this ticket_?

atolopko-czi:
This is definitely an old ticket! 

There are two stubbed/ignored tests that are intended to test coverage for https://github.com/chanzuckerberg/cellxgene-census/blob/96c0499cd83521d169220f37555724bc9c67f114/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py#L547-L555. But honestly, we need to _eliminate_ the multi-worker support entirely (it's redundant w/TileDB threaded I/O), so IMO those test stubs can be deleted.

Regarding the other two test suggestions, above:
```
 The obs / X data is joined correctly (X rows correspond to their respective obs soma_joinids)
```
This appears to be implicitly tested by various methods such as `test_batching__all_batches_full_size`, `test_batching__exactly_one_batch`, and `test_experiment_dataloader__batched`, where both the obs and X contents of a batch are asserted in tandem.  Could still be worth adding a test that is named appropriately to explicitly capture the requirement, e.g. `test_obs_and_X_are aligned`.

```
 The var / X data joined correctly (gene ordering matches var query filter)
```
Similarly, this is implicitly tested by tests that assert the ordered values within a row of X, but I would add an explicit test for this. And particularly for sparse tensors, I would test that the expected data at a given var coordinate is correct.

PR #1103, opened 2024-04-16
[python] Shuffle multiple SOMA chunks

atolopko-czi:
Adds a `shuffle_chunk_count` parameter.

Improves randomness of shuffling, while allowing for explicit tuning of memory usage vs I/O performance.


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -121,23 +121,23 @@ def __init__(
...
-        self.obs_joinids_chunks_iter = self._maybe_local_shuffle_obs_joinids(obs_joinids_chunked, shuffle_rng)
+        if shuffle_chunk_count:
+            assert shuffle_rng is not None
+            splits = len(obs_joinids_chunked) // min(len(obs_joinids_chunked), shuffle_chunk_count)
+            self.obs_joinids_chunks_iter = (
+                shuffle_rng.permutation(np.concatenate(shuffle_chunks))
+                for shuffle_chunks in np.array_split(obs_joinids_chunked, splits)
+            )
+        else:
+            self.obs_joinids_chunks_iter = iter(obs_joinids_chunked)
```
atolopko-czi, line 138: should factor this out into a method

api/python/cellxgene_census/tests/experimental/ml/test_pytorch.py
```diff
@@ -570,6 +570,33 @@ def test__shuffle(soma_experiment: Experiment) -> None:
...
+
+    all_rows = list(iter(dp))
+
+    soma_joinids = [row[1][0].item() for row in all_rows]
+    X_values = [row[0][0].item() for row in all_rows]
+
+    # same elements
+    assert set(soma_joinids) == set(range(16))
+    # not ordered! (...with a `1/16!` probability of being ordered)
+    assert soma_joinids != list(range(16))
```
atolopko-czi, line 594: TODO: Could assert that the first and second half of the `soma_joinids` are each formed from exactly 2 quarters of the full data set (without knowing which ones specifically, since it's random).

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
prathapsridharan: If I am reading this correctly, if `shuffle_chunk_count = 2`, then this block of code would split the list globally shuffled chunks into 2 lists of chunks where each such list contains say N/2 chunks. It would then concatenate these N/2 chunks in memory and then shuffle this concatenated ndarray in memory.

This might exceed the memory budget and cause an OOM crash.

I think the `splits` calculation in line 132 should be removed, and `shuffle_chunk_count` should be used in place of `splits` in line 135.

So if `shuffle_chunk_count = 2`, it would take 2 chunks at a time and concatenate them and then shuffled the concatenated ndarray which would fit into the memory budget.

prathapsridharan (changes requested):

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
prathapsridharan: @atolopko-czi - Please disgard the above comment. I confused the math. Assuming There are `16 chunks` and `shuffle_chunk_count = 2`, you would need `splits = 16/2 = 8` to split the `16 chunks` into 8 partitions of `2 chunks` each.

This makes sense now.

prathapsridharan (approved):

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
atolopko-czi: yeah, probably worth refactoring this to improve clarity; i'm pushing some variable renames now fwiw

prathapsridharan:
For informational purposes, a description of a general algorithm is captured in this github issue:

https://github.com/chanzuckerberg/cellxgene-census/issues/1146

api/python/cellxgene_census/tests/experimental/ml/test_pytorch.py
```diff
@@ -570,6 +570,37 @@ def test__shuffle(soma_experiment: Experiment) -> None:
...
+    # randomizes X in same order as obs
+    # note: X values were explicitly set to match obs_joinids to allow for this simple assertion
+    assert X_values == soma_joinids
+
+    # If shuffle_chunk_count is defined, each batch should contain elements from different chunks
+    batches = [soma_joinids[i : i + 4] for i in range(0, len(all_rows), 4)]
+    assert any(max(batch) - min(batch) > 3 for batch in batches)
+
```
prathapsridharan, line 602: I would recommend against assertions like these when randomness is part of the function. Testing randomness is pretty complex topic and I think these types of assertions make the test flakey (because of randomness) and doesn't give the reader of the test much information about what is going on.

Instead of lines `599-601`, I might recommend accessing [self.soma_chunk_iter](https://github.com/chanzuckerberg/cellxgene-census/blob/0220e06462b29219df2925831d037c34244197e1/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py#L256) of the [_ObsAndXIterator](https://github.com/chanzuckerberg/cellxgene-census/blob/0220e06462b29219df2925831d037c34244197e1/api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py#L605) object and checking:

1. `assert len(soma_chunk_iter) == obs_range // (soma_chunk_size * shuffle_chunk_count)`
2. Each element of `soma_chunk_iter` is of size `(soma_chunk_size * shuffle_chunk_count)` _except possibly the last element_
3. _Nice to have:_ If `batch_size > 1` and `batch_size` does not evenly divide `soma_chunk_size` , then check that a `batch_size` length of `soma_joinids` or `X_values` in `lines 588-589` correctly cross the boundary of `soma_chunk_iter`

(3) is probably not necessary right now but I think (1) and (2) tests the crux of the algorithm and assume randomizing the concatenated chunks just work.

I agree that doesn't test how random the scatter gather algorithm is and that is definitely worth testing but I think it involves more thought and should be a separate test altogether.

IMHO, tests of randomness are more like _stress tests_ than unit tests and they require quite a large amount of data and will _probably be slower than what is acceptable for unit tests_

prathapsridharan (changes requested):

prathapsridharan:
Per a synchronous conversation with @ebezzi we decided to scrap the test as it needs more thought. [We will make a ticket to write a better test](https://github.com/chanzuckerberg/cellxgene-census/issues/477#issuecomment-2148056462) for it but for the sake of expediency, we want to get this merged and get some of our first users to use it. Anecdotal evidence suggests that the `scatter-gather-shuffle` algorithm is performant and gives good randomness.

prathapsridharan (approved):
LGTM

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -451,6 +475,9 @@ def __init__(
                 available for processing via the iterator. This allows network (or filesystem) requests to be made in
                 parallel with client-side processing of the SOMA data, potentially improving overall performance at the
                 cost of doubling memory utilization. Defaults to ``True``.
+            shuffle_chunk_count:
+                The number contiguous blocks (chunks) of rows to read at random and then concatenated and shuffled.
+                Larger number for `shuffle_chunk_count` correspond to more randomness in the shuffling.
```
ivirshup, line 480: There are some typos here, but I think this could also be explained a little more clearly (though maybe the whole approach should be documented somewhere).

If we're okay with calling the intermediate collection a "shuffle buffer", then:

```suggestion
                The number of contiguous blocks (chunks) of rows sampled for each shuffle buffer.
                Each shuffle buffer is then shuffled again, and used a source for each batch.
                Larger numbers correspond to more randomness per training batch.
```

Otherwise:

```suggestion
                The number of contiguous blocks (chunks) of rows sampled to then concatenate and shuffle.
                Larger numbers correspond to more randomness per training batch.
```

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -184,6 +191,21 @@ def __next__(self) -> _SOMAChunk:
         return _SOMAChunk(obs=obs_batch, X=X_batch, stats=stats)
 
 
+def list_split(arr_list: List[Any], sublist_len: int) -> List[List[Any]]:
+    """Splits a python list into a list of sublists where each sublist is of size `sublist_len`."""
```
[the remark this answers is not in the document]
ebezzi: Let's add it.

Issue #1112, opened 2024-04-25
Census cell similarity search: pipeline to build TileDB Vector Indexes of cell embeddings


mlin:
Develop a productionizable pipeline to build the indexes for [TileDB-Vector-Search](https://github.com/TileDB-Inc/TileDB-Vector-Search) from the stored Census embeddings (starting with scVI but also UCE, Geneformer, etc.). This consists of some Python code to read the embeddings sparse arrays and build the indexes (which are themselves TileDB arrays), then packaged up for cloud deployment. It's expected to take a few hours for each set of embeddings, and the different sets of embeddings can be processed in parallel.

Unless suggested otherwise, I'll package this as a dockerized WDL pipeline since that's most familiar to me (@mlin).

PR #1122, opened 2024-05-02
Vector indexing pipeline for Census embeddings

mlin:
#1112 

Adds a Docker+WDL pipeline to build the [TileDB-Vector-Search](https://github.com/TileDB-Inc/TileDB-Vector-Search) index for each Census embedding. (Excluding NMF for now due to some minor technical issues that will be easier to deal with after future improvements to TileDB-Vector-Search.)

The pipeline launches a powerful instance to build the index for each embedding, and processes the embeddings in parallel. The longest takes 9-10 hours to build (UCE, which has the highest dimensionality). The output is an S3 folder which can be copied into the destination bucket.

### DIFF
--- a/tools/census_embeddings_indexer/README.md
+++ b/tools/census_embeddings_indexer/README.md
@@ -1,0 +1,22 @@
+# census_embeddings_indexer
+
+This is a Docker+WDL pipeline for building [TileDB-Vector-Search](https://github.com/TileDB-Inc/TileDB-Vector-Search) indexes for Census cell embeddings, supporting cell similarity search in embedding space. The pipeline consumes one or more of the existing TileDB arrays for hosted and contributed [Census embeddings](https://cellxgene.cziscience.com/census-models) stored on S3. The resulting indexes are themselves TileDB groups to be stored on S3.
+
+```
+export AWS_ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
+AWS_DEFAULT_REGION=$(aws configure get region)
+export ECR_ENDPT=${AWS_ACCOUNT_ID}.dkr.ecr.us-west-2.amazonaws.com
+
+docker build -t ${ECR_ENDPT}/omics:census_embeddings_indexer .
+aws ecr get-login-password | docker login --username AWS --password-stdin "$ECR_ENDPT"
+docker push ${ECR_ENDPT}/omics:census_embeddings_indexer
+
+miniwdl-omics-run census_embeddings_indexer.wdl \
+    embeddings_s3_uris=s3_//cellxgene-census-public-us-west-2/cell-census/2023-12-15/soma/census_data/homo_sapiens/ms/RNA/obsm/scvi \
+    s3_region=$AWS_DEFAULT_REGION \
+    docker=${ECR_ENDPT}/omics:census_embeddings_indexer \
+    --output-uri s3://mlin-census-scratch/census_embeddings_indexer/out/ \
+    --role poweromics
+```
+
+(The `embeddings_s3_uris=s3_//...` is a workaround for an AWS-side existence check that doesn't seem to work right on public buckets.)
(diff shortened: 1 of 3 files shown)


tools/census_embeddings_indexer/README.md
```diff
@@ -0,0 +1,32 @@
+# census_embeddings_indexer
+
+This is a Docker+WDL pipeline to build [TileDB-Vector-Search](https://github.com/TileDB-Inc/TileDB-Vector-Search) indexes for Census cell embeddings, supporting cell similarity search in embedding space. It's meant to run on the AWS HealthOmics workflow service using the [miniwdl-omics-run](https://github.com/miniwdl-ext/miniwdl-omics-run) launcher (assuming account setup documented there).
+
+The pipeline consumes one or more of the existing TileDB arrays for hosted and contributed [Census embeddings](https://cellxgene.cziscience.com/census-models) stored on S3. The resulting indexes are themselves TileDB groups to be stored on S3.
+
+```bash
+export AWS_ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
+AWS_DEFAULT_REGION=$(aws configure get region)
```
ebezzi, line 9: Missing an `export`?

tools/census_embeddings_indexer/README.md
```diff
@@ -0,0 +1,32 @@
...
+export ECR_ENDPT=${AWS_ACCOUNT_ID}.dkr.ecr.us-west-2.amazonaws.com
```
ebezzi, line 10: Kinda of a nitpick, but if we want to keep this region customizable, I think we should interpolate the region here as well.

tools/census_embeddings_indexer/README.md
```diff
@@ -0,0 +1,32 @@
...
+    embeddings_s3_uris=s3_//cellxgene-contrib-public/contrib/cell-census/soma/2023-12-15/CxG-czi-1 \
+    embeddings_s3_uris=s3_//cellxgene-contrib-public/contrib/cell-census/soma/2023-12-15/CxG-czi-4 \
+    embeddings_s3_uris=s3_//cellxgene-contrib-public/contrib/cell-census/soma/2023-12-15/CxG-czi-5 \
+    embeddings_s3_uris=s3_//cellxgene-contrib-public/contrib/cell-census/soma/2023-12-15/CxG-contrib-1 \
+    embeddings_s3_uris=s3_//cellxgene-contrib-public/contrib/cell-census/soma/2023-12-15/CxG-contrib-2 \
+    embeddings_s3_uris=s3_//cellxgene-contrib-public/contrib/cell-census/soma/2023-12-15/CxG-contrib-3 \
+    census_version=2023-12-15 \
+    s3_region=$AWS_DEFAULT_REGION \
+    docker=${ECR_ENDPT}/omics:census_embeddings_indexer \
+    --output-uri s3://mlin-census-scratch/census_embeddings_indexer/out/ \
```
ebezzi, line 26: Parametrize the output bucket?

ebezzi (approved):
Few nitpicks but LGTM.

ivirshup (commented):
Very minor, but should `input_vectors_per_work_item` and `memory_budget` also be derived from `cpu`, since requested amount of memory is derived form number of CPUs?

And a question: Is there some previous discussion on how this indexing strategy was decided on?

pablo-gar (commented):
I defer to other engineers on approving this PR.

ivirshup:
> Starting out with TileDB-Vector-Index was kind of the obvious initial choice since all the other Census data and embeddings are stored with TileDB

I was actually wondering why the IVF_FLAT over the other index options in the tiledb library.

mlin:
@ivirshup Ah, that one's easy. When we initially scoped this, `IVFFlatIndex` was the only real choice (`FlatIndex` being a strawperson exhaustive-search implementation). I see they've added a newer `VamanaIndex`. I'll ask the TileDB developer if that'd be interesting in the future for our Census embeddings, but I don't think we need to do a deep dive on it right now. For future-proofing, it seems like a good idea to specify the index type in the manifest JSON we discussed with @ebezzi this morning though!

PR #1129, opened 2024-05-06
[python] Fix pytorch imports

ebezzi:
Some of the imports in test_pytorch.py were causing the tests to fail even if they were not invoked (as in, not in experimental mode). This should fix the issue.


prathapsridharan (approved):
LGTM! Thank you!

PR #1150, opened 2024-05-17
[python] Fix git-lfs issue when cloning Geneformer

ebezzi:
Attemps to fix an issue caused by git-lfs which is used by Geneformer tests.

### DIFF
--- a/.github/workflows/py-unittests.yml
+++ b/.github/workflows/py-unittests.yml
@@ -33,7 +33,7 @@ jobs:
         run: |
           python -m pip install -U pip setuptools wheel
           pip install --use-pep517 accumulation-tree # Geneformer dependency needs --use-pep517 for Cython
-          pip install -r ./api/python/cellxgene_census/scripts/requirements-dev.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/py-unittests.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         run: |
           python -m pip install -U pip setuptools wheel
           pip install --use-pep517 accumulation-tree # Geneformer dependency needs --use-pep517 for Cython
-          pip install -r ./api/python/cellxgene_census/scripts/requirements-dev.txt
+          GIT_CLONE_PROTECTION_ACTIVE=false pip install -r ./api/python/cellxgene_census/scripts/requirements-dev.txt
```
ivirshup, line 36: It looks like this fixes it! Or, at least it doesn't fail to install anymore.

Should this get  a link to the issue in git-lfs here, or an issue tracking the removal of this workaround?

ebezzi: I'm thinking that maybe these tests should not belong here, as the specific dependency is quite obscure (it even points to a specific commit). If we decide to purge them, this won't be needed, otherwise we should definitely create an issue.

ivirshup (approved):

Issue #1033, opened 2024-02-28
Communicate download progress to Census API users


hthomas-czi:
[Visibility of system status](https://www.nngroup.com/articles/ten-usability-heuristics/#1-visibility-of-system-status) is an important usability heuristic. Users want to know how long a data fetching call will take and how much progress is being made.

<Start Pablo's edit>
Applies to

`SOMA.DataFrame.read().concat()`
`SOMA.SparseNDArray.read().concat()`
`cellxgene_census.download_source_h5ad()`
<End Pablo's edit>

**User Quote**
> I’d like to see progress bars to understand how long to wait for download to finish

pablo-gar:
@ebezzi for the SOMA-related task let's ask TileDB if there is any possibility to implement anything on this.

ivirshup:
This is very straightforward for `cellxgene_census.download_source_h5ad`. Basically just:

```python
from fsspec.callbacks import TqdmCallback

...

    fs = s3fs.S3FileSystem(
        anon=True,
        cache_regions=True,
    )
    fs.get_file(
        locator["uri"],
        to_path,
        callback=TqdmCallback(),
    )
```

![image](https://github.com/chanzuckerberg/cellxgene-census/assets/8238804/2acf241b-79fb-4d32-aece-a9556e809354)

It sounds like using this for other things is more of a SOMA issue. 

* Should it be possible to turn this reporting off?
* Any ideas on how to test this?

prathapsridharan:
In regards to progress for other APIs - Is supplying way of getting progress for programmatic APIs a common expectation? 

`download_*` seems straightforward enough but I can't recall expecting a progress indicator when programmatically invoking APIs on a terminal or script.

Would be good to have examples from other domains where programmatic usage of an API also has a mechanism to communicate progress of work.

ivirshup:
> download_* seems straightforward enough but I can't recall expecting a progress indicator when programmatically invoking APIs on a terminal or script.

I think giving the option to report progress (sometimes by default) is pretty common, `wget` for instance. `dask` also has a number of tools for reporting progress of ongoing tasks. A lot of bioinformatics tools (though especially older ones) also make this available.

It's nice when its done via stderr or some alternative stream.

pablo-gar:
I think something that can suffice the request and is simple for now for `SOMA.DataFrame.read().concat()` and `SOMA.SparseNDArray.read().concat()` is:

[...]

PR #1153, opened 2024-05-20
[python] Progress bar for download_source_h5ad

ivirshup:
For #1033

Adding a progress bar for download_source_h5ad

TODO:

- [x] Tests?
- [x] Do we want a way to turn this off?
- [x] Does this display correctly in a notebook?
  - I does a text based progress bar in a vscode notebook. There is an HTML one, but the main difference is that the html one is blue.

Currently looks like:

```python
In [3]: cellxgene_census.download_source_h5ad(small_dataset.dataset_id, "tmp.h5ad")
Downloading: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████| 5.10M/5.10M [00:00<00:00, 17.7MB/s]
```

### DIFF
--- a/api/python/cellxgene_census/src/cellxgene_census/_open.py
+++ b/api/python/cellxgene_census/src/cellxgene_census/_open.py
@@ -14,7 +14,7 @@ from typing import Any, Dict, Optional, get_args
 
 import s3fs
 import tiledbsoma as soma
-from fsspec.callbacks import TqdmCallback
+from fsspec.callbacks import NoOpCallback, TqdmCallback
 
 from ._release_directory import (
     CensusLocator,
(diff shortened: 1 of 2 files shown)


api/python/cellxgene_census/src/cellxgene_census/_open.py
```diff
@@ -334,4 +335,8 @@ def download_source_h5ad(dataset_id: str, to_path: str, *, census_version: str =
         anon=True,
         cache_regions=True,
     )
-    fs.get_file(locator["uri"], to_path)
+    fs.get_file(
+        locator["uri"],
+        to_path,
+        callback=TqdmCallback(),
```
prathapsridharan, line 341: Since `TqdmCallback()` invocation is a side effect would suffice to mock `TqdmCallback()` in the test and simply assert that it is called?

prathapsridharan (approved):

pablo-gar (approved):

ebezzi (approved):

PR #1155, opened 2024-05-21
Embrace the __future__ of typing

ivirshup:
Turns out this needed a little bit of manual work. For now, here is a demo of what this looks like. I've covered the main census API, but not the tests or experimental yet.


ivirshup:
I need to look into the failure, iirc this may be fixable with `typing-extensions`

PR #1158, opened 2024-05-22
[python] Don't kill all tests on failure

ivirshup:
It's working! You can see that on commit fafb3ff the actions all quit after one failure, but after b2058a0 they continue to run. 

~~I'm starting this with a test that should fail on some jobs, to confirm that it kills the others. Then I will modify the workflow file to make it so that doesn't happen.~~

- [x] Get rid of test that is expected to fail
- [x] Push commit with comments explaining the changes

### DIFF
--- a/.github/workflows/py-unittests.yml
+++ b/.github/workflows/py-unittests.yml
@@ -7,6 +7,10 @@ on:
   push:
     branches: [main]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   unit_tests_python_api:
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/py-unittests.yml
```diff
@@ -7,9 +7,15 @@ on:
   push:
     branches: [main]
 
+# If a new commit is pushed, cancel the jobs from previous commits.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
```
ivirshup, line 14: This change is actually unrelated to the initial purpose of this PR. This change makes it so: if you push a commit, existing jobs get cancelled. I like this one for keeping runners available/ reducing usage. But it is unrelated to this PR in general, and can be removed if you want.

ebezzi: I think this makes sense.

ebezzi (approved):

prathapsridharan (approved):

Issue #1115, opened 2024-04-25
Census cell similarity search: experimental Python API for metadata prediction


mlin:
Building on chanzuckerberg/cellxgene-census#1114, add another API method that predicts metadata attributes of the query cells based on the similar cells. The first pass can simply be a plurality vote of similar cells, with the vote share as the confidence score. In the future we can weight by distance or do other fancier modelling.

Issue #1114, opened 2024-04-25
Census cell similarity search: experimental Python API for searching given AnnData


mlin:
Having built the vector indexes on Census embeddings (#694 chanzuckerberg/cellxgene-census#1113), develop the Python API inside `cellxgene_census.experimental` to input an AnnData and identify the most-similar Census cells.

This necessitates running a forward pass of the embedding model (starting with scVI) on the given AnnData. In case that causes a lot of complications, we can create an initial demo that searches existing Census cells against the index and break out the forward passes into a separate issue. It might even end up involving a docker image or web service of some sort.

mlin:
Per 4/29 discussion:

For now (i) the API assumes the given AnnData will include a layer with suitable embeddings and (ii) we'll informally provide a notebook/docker showing how to do the forward pass to add them. To be revisited in H2.

mlin:
Draft PR: https://github.com/chanzuckerberg/cellxgene-census/pull/1164

(finalization pending #1116)

PR #1164, opened 2024-05-27
Embeddings search experimental API

mlin:
Adds two new functions to `cellxgene_census.experimental`:

1. `find_nearest_obs` uses TileDB-Vector-Search indexes of Census embeddings to find nearest neighbors of given embedding vectors (in an AnnData obsm layer). #1114
2. `predict_obs_metadata` uses the nearest neighbors to predict metadata attributes like cell_type and tissue_general for the query cells. Naive initial implementation is just a starting point to start experimenting with. #1115

The TileDB-Vector-Search query speed seems to be very S3-latency-sensitive, even moreso than typical Census queries. It's many times faster to run from within AWS us-west-2 than externally.


mlin:
@ebezzi Putting this up for initial review since it's working well, but we still need to plan action on #1181 -- this still copies the approach of hard-coding the base S3 URI.

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
```diff
@@ -0,0 +1,141 @@
...
+        )
+
+    # formulate index URI and run query
+    index_uri = CENSUS_EMBEDDINGS_INDEX_URI_FSTR.format(
+        census_version=embedding_metadata["census_version"], embedding_id=embedding_metadata["id"]
+    )
+    memory_vectors = memory_GiB * (2**30) // (4 * n_features)  # number of float32 vectors
+    index = vs.ivf_flat_index.IVFFlatIndex(
+        uri=index_uri,
+        config={"vfs.s3.region": CENSUS_EMBEDDINGS_INDEX_REGION, "vfs.s3.no_sign_request": "true"},
```
ivirshup, line 80: Could this config be based off `DEFAULT_TILEDB_CONFIGURATION`?

It would be nice to keep config setting centralized, especially if we are going to need to control this programmatically for analytics.

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
```diff
@@ -0,0 +1,141 @@
...
+    implementation-dependent.
+    """
+
+    neighbor_ids: npt.NDArray[np.int64]
+    """
+    obs soma_joinid's of the nearest neighbors for each query embedding (q by k).
+    """
+
+
+def find_nearest_obs(
```
ivirshup, line 37: On the API side, it would be nice if this could produce output that can be directly with sklearn style classes. For example, if this returned a KNNTransformer subclass, that could be used directly with the [KNeighborsClassifier and KNeighborsRegressor](https://scikit-learn.org/stable/modules/neighbors.html#nearest-neighbors-transformer) classes.

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
```diff
@@ -0,0 +1,141 @@
...
+    embedding_metadata: Dict[str, Any],
```
ivirshup, line 38: Why does this use a different way to specify an embedding than [`get_embedding`](https://chanzuckerberg.github.io/cellxgene-census/_autosummary/cellxgene_census.experimental.get_embedding.html#cellxgene_census.experimental.get_embedding) does?

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
```diff
@@ -0,0 +1,141 @@
...
+        # step through query cells to generate prediction for each column as the plurality value
+        # found among its neighbors, with a confidence score based on the simple fraction (for now)
+        # TODO: something more intelligent for numeric columns! also use distances, etc.
+        out: Dict[str, List[Any]] = {}
+        for i in range(neighbors.neighbor_ids.shape[0]):
+            neighbors_i = neighbor_obs.loc[neighbors.neighbor_ids[i]]
+            for col in column_names:
+                col_value_counts = neighbors_i[col].value_counts(normalize=True)
+                out.setdefault(col, []).append(col_value_counts.idxmax())
+                out.setdefault(col + "_confidence", []).append(col_value_counts.max())
```
ivirshup, line 139: I would suggest using sklearn.KNeighborsClassifier here instead of rolling our own. I'd be concerned about the performance of having this loop happen in python code.

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
mlin: `get_embedding` is a little low level in that it wants the full URI to the embeddings TileDB array, which isn't actually needed to find the index. `embedding_metadata` is the information returned from `get_embedding_metadata()` which seems like the appropriate level (especially in view of #1181 wherein we will actually put the relative URIs to the index arrays in there), although of course it'd be nice if it were more typesafe.

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
mlin: @ivirshup I like this idea very much, but I'm not quite sure it's workable (albeit I'm not as familiar with those APIs)... 

Those scikit-learn classes seem oriented around the scenario where you're providing either all the points (in the "universe") or the complete distance matrix for them. Here we're working with a more limited view of the query points and their neighbor distances; we don't have or want the complete distance matrix, and actually we don't even have the coordinates of the neighbors immediately handy.

Do you think the shoe fits? I see there's some stuff about the "K neighbors graph" that might be relevant, but I'm not personally familiar enough to use them in an unconventional/advanced way like this.

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
pablo-gar: @mlin can we change it to?

```
   embedding_name: str, 
   organism: str, 
   census_version: str
```

I think that provides an easier entry point to users and it aligns to [get_embedding_metadata_by_name](https://chanzuckerberg.github.io/cellxgene-census/_autosummary/cellxgene_census.experimental.get_embedding_metadata_by_name.html#cellxgene_census.experimental.get_embedding_metadata_by_name)

pablo-gar (changes requested):
Looks good to me except one comment in API signature

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
```diff
@@ -0,0 +1,167 @@
...
+    index = vs.ivf_flat_index.IVFFlatIndex(uri=index_uri, config=config, memory_budget=memory_vectors)
+    distances, neighbor_ids = index.query(query.obsm[embedding_name], k=k, nprobe=nprobe, **kwargs)
+
+    return NeighborObs(distances=distances, neighbor_ids=neighbor_ids)
+
+
+def _resolve_embedding_index(
+    embedding_metadata: Dict[str, Any],
+    mirror: Optional[str] = None,
+) -> Optional[Tuple[str, str]]:
```
mlin, line 101: @ebezzi new index resolution method here

mlin:
@ebezzi @pablo-gar @ivirshup Updated this to resolve indexes through mirrors/contributions json and remove the need for caller to use `get_embedding_metadata_by_name()` on their own. Please take another pass including the prior discussion. Unfortunately we have known CI issues currently but I've run the new test cases locally. 🙏

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
```diff
@@ -0,0 +1,167 @@
...
+    neighbor_ids: npt.NDArray[np.int64]
+    """
+    obs soma_joinid's of the nearest neighbors for each query embedding (q by k).
+    """
+
+
+def find_nearest_obs(
+    embedding_name: str,
+    organism: str,
+    census_version: str,
```
ivirshup, line 39: This doesn't uniquely refer to an embedding, right? It refers to the newest embedding with this name?

I think we need to be able to non-ambiguously specify an embedding here since it's really important that the embedding being queried matches what we projected our data into.

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
ivirshup: This doesn't uniquely refer to an embedding, right? It refers to the newest embedding with this name?

I think we need to be able to non-ambiguously specify an embedding here since it's really important that the embedding being queried matches what we projected our data into.

pablo-gar: At the moment the combination of the three pieces does match to a unique embedding matrix, unless there is an edge-case I'm not catching.

ivirshup: The documentation for `get_embedding_metadata_by_name` says that it can be ambiguous and returns the latest one.

I think the unique value for embeddings is in the `"id"` key of the embedding metadata, while `"embedding_name"` just says what method was used.

https://github.com/chanzuckerberg/cellxgene-census/blob/7612fb36e3504ea8886f855af0fd1cc8f67da10b/api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding.py#L171-L172

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
```diff
@@ -0,0 +1,167 @@
...
+        # TODO: something more intelligent for numeric columns! also use distances, etc.
+        out: Dict[str, List[Any]] = {}
+        for i in range(neighbors.neighbor_ids.shape[0]):
+            neighbors_i = neighbor_obs.loc[neighbors.neighbor_ids[i]]
+            for col in column_names:
+                col_value_counts = neighbors_i[col].value_counts(normalize=True)
+                out.setdefault(col, []).append(col_value_counts.idxmax())
+                out.setdefault(col + "_confidence", []).append(col_value_counts.max())
+
+    return pd.DataFrame.from_dict(out)
```
ivirshup, line 167:
```suggestion
        max_joinid = neighbor_obs.index.max()
        out: dict[str, pd.Series] = {}
        indices = np.broadcast_to(np.arange(neighbors.neighbor_ids.shape[0]), (10, neighbors.neighbor_ids.shape[0])).T
        g = sparse.csr_matrix(
            (
                np.broadcast_to(1, neighbors.neighbor_ids.shape[0] * 10),
                (
                    indices.flatten(),
                    neighbors.neighbor_ids.astype(np.int64).flatten(),
                ),
            ),
            shape=(neighbors.neighbor_ids.shape[0], max_joinid + 1)
        )
        for col in column_names:
            col_categorical = neighbor_obs[col].astype("category")
            joinid2category = sparse.coo_matrix(
                (
                    np.broadcast_to(1, len(neighbor_obs)),
                    (
                        neighbor_obs.index,
                        col_categorical.cat.codes
                    )
                ),
                shape=(max_joinid + 1, len(col_categorical.cat.categories))
            )
            counts = g @ joinid2category
            rel_counts = counts / counts.sum(axis=1)
            out[col] = col_categorical.cat.categories[rel_counts.argmax(axis=1).A.flatten()].astype(object)
            out[f"{col}_confidence"] = rel_counts.max(axis=1).A.flatten()

    return pd.DataFrame.from_dict(out)
```

This should be faster, but accomplish the same thing. Unfortunately it looks like the results are a little different due to a difference in how ties are resolved (but that's only for low confidence labels anyways).

For 2.3k cells, using the current implementation:

```python
%%time
result = cellxgene_census.experimental.predict_obs_metadata(organism=organism, census_version=census_version, neighbors=test_query, column_names=["cell_type"])
```

```
CPU times: user 16.2 s, sys: 2.06 s, total: 18.3 s
Wall time: 15.2 s
```

For the implementation suggested here:

```python
%%time
result_new = predict_obs_metadata(organism=organism, census_version=census_version, neighbors=test_query, column_names=["cell_type"])
```


```
CPU times: user 3.21 s, sys: 1.86 s, total: 5.07 s
Wall time: 1.95 s
```

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
pablo-gar: Right `"id"` is unique, but the IDs are `"cxg-czi-[\d+]"` or  `"cxg-contributed-[\d+]"` -- so no human-friendly.

The reason for  `If more embeddings match the query parameters, the most recent one will be returned.` is that we ended up having two versions of scVI for the 2023-12-15 Census version, the second effectively being a replacement of the first one (the Yosef lab recommended to that first needed some fixing).

So now refreshing my memory that one is actually the only edge-case we have, but it is intended for our users to use the latest version. I don't anticipate having to have multiple versions of the same embedding with the same name, unless one is the replacement of the other.

api/python/cellxgene_census/src/cellxgene_census/experimental/_embedding_search.py
```diff
@@ -0,0 +1,184 @@
...
+        for col in column_names:
+            col_categorical = neighbor_obs[col].astype("category")
+            joinid2category = sparse.coo_matrix(
+                (np.broadcast_to(1, len(neighbor_obs)), (neighbor_obs.index, col_categorical.cat.codes)),
+                shape=(max_joinid + 1, len(col_categorical.cat.categories)),
+            )
+            counts = g @ joinid2category
+            rel_counts = counts / counts.sum(axis=1)
+            out[col] = col_categorical.cat.categories[rel_counts.argmax(axis=1).A.flatten()].astype(object)
+            out[f"{col}_confidence"] = rel_counts.max(axis=1).A.flatten()
```
mlin, line 182: @ivirshup the test case choked on this, mind taking a look?

```
>               out[f"{col}_confidence"] = rel_counts.max(axis=1).A.flatten()
E               AttributeError: 'coo_matrix' object has no attribute 'A'
```

ivirshup:
```suggestion
            out[f"{col}_confidence"] = rel_counts.max(axis=1).toarray().flatten()
```

I think this should do it

mlin:
@ivirshup I split out the perf optimization to #1257 since I was still getting an error, will write more there -- hope you don't mind, it's only because I need to triage desperately right now!

Issue #620, opened 2023-07-13
Add search functionality to cellxgene-census docs


ayushnoori:
## Description

Add search functionality to `cellxgene-census` documentation website at https://chanzuckerberg.github.io/cellxgene-census/index.html.

## Context

See https://cziscience.slack.com/archives/C04LMG88VKJ/p1689261048038369.

ivirshup:
Some updates after an initial investigation:

* Sphinx by default includes search entirely client side search. E.g. you start a new project and run `make html`, there will be a functioning search bar
* Currently this feature is disabled in the census docs. The search index is still built, but the search bar isn't excluded by using a blank `"searchbox.html"` template. If I remove that, a search bar is visible but still seems non-functional due to a missing jquery. Still not sure how that is being excluded or how I can get it back.

ivirshup:
As an alternative to client side search, there is also Algolia DocSearch. I've had a little bit of trouble with it in the past, but think it's overall fine. No idea if it's going to give users cookies.

ivirshup:
Closing, since it was fixed in #1166

ayushnoori:
Great, thanks so much, @ivirshup!

PR #1166, opened 2024-05-28
[python] Make docs searchable

ivirshup:
This PR addresses https://github.com/chanzuckerberg/cellxgene-census/issues/620

I've re-enabled Sphinx's built-in client side search by:

* Removing the blank template for the search bar
* Removing the line:

```python
html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
```

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -66,7 +66,6 @@ source_suffix = ['.rst', '.md']
 # Inject custom css files in `/_static/css/*`
 html_static_path = ['_static']
 
-import sphinx_rtd_theme
 html_theme = "sphinx_rtd_theme"
 
 html_js_files = [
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -66,9 +66,7 @@
 # Inject custom css files in `/_static/css/*`
 html_static_path = ['_static']
 
-import sphinx_rtd_theme
 html_theme = "sphinx_rtd_theme"
-html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
```
ivirshup, line 71: @ebezzi, do you remember why you added this?

ebezzi: I believe that was a requirement for RTD.

ebezzi (approved):

pablo-gar (approved):
Looks great. Thanks @ivirshup! We'll need to check it works well once deployed in `github.io`

ivirshup:
Yeah, it would be nice if we had preview-able doc builds as part of CI. Also a hosted development version of docs.

PR #1167, opened 2024-05-29
Identify tiledbsoma version that causes builder unit tests to fail

prathapsridharan:
This PR shows the `tiledbsoma` version that causes census builder unit tests to fail


prathapsridharan:
A fix for this has been identified by `tiledbsoma` maintainers. The issue is fixed by `tiledbsoma` at commit `3869abc` and verified by passing builder unit tests as can be seen in the CI. The issue is tracked here:

https://github.com/single-cell-data/TileDB-SOMA/issues/2663

PR #1173, opened 2024-05-31
[python] Update scvi pipeline for the June LTS training

ebezzi:


tools/models/scvi/scvi-config.yaml
```diff
@@ -19,7 +21,7 @@ model:
   filename: "scvi.model"
   n_hidden: 512
   n_latent: 50
-  n_layers: 1
+  n_layers: 2
```
prathapsridharan, line 24: What is the reason for incrementing `n_layers`? Apologies if you have already addressed this in our synchronous meeting and I missed the point.

ebezzi: This was determined during the training phase last year, but I forgot to update it in the configuration file.

prathapsridharan (approved):

PR #1178, opened 2024-06-04
[python] Fix wrong parameters in ExperimentDataPipe

ebezzi:
Not sure how those ended up there in the first place.


prathapsridharan (approved):

ivirshup (approved):

atolopko-czi:
> Not sure how those ended up there in the first place.
It was lacking a proper test! 😳 Probably should add one to test the defaults since all of the existing tests explicitly set these two params.

PR #1184, opened 2024-06-05
[python] Restore soma indexer in hvg

ebezzi:
Previously removed due a memory bug, now fixed by https://github.com/single-cell-data/TileDB-SOMA/issues/2662


ivirshup (approved):

ivirshup:
The issue before was memory, right? Would you want to introduce a memory test here?

I've been using https://github.com/bloomberg/pytest-memray for this

ebezzi:
> The issue before was memory, right? Would you want to introduce a memory test here?
> 
> I've been using https://github.com/bloomberg/pytest-memray for this

That would be great. This should be part of the upcoming "expanded regression testing" work for sure.

PR #1186, opened 2024-06-05
Fix incorrect logger calls

prathapsridharan:
There were calls to `logging.info` or `logging.debug`. They should be `logger.info` and `logger.debug`.


ebezzi (approved):

ivirshup (approved):
Did a quick check for any remaining cases, but couldn't find any. lgtm

PR #1189, opened 2024-06-06
[python] Follow spec-0

### LINKED ISSUE #1177: Follow SPEC-0 for minimum supported versions of dependencies
## Motivation

https://scientific-python.org/specs/spec-0000/

It would be nice to have a clear policy on what the support window for dependencies are. This makes it clear when we can use newer features, and when we can stop supporting old versions of libraries.

As an illustration of why this is important, let's say a user comes across a bug specific to python 3.8. It could turnout that this is a bug in all the versions of pandas that support 3.8, but was fixed after pandas dropped support for 3.8. Now it's basically on us to solve bugs in dependencies since we have a larger support window that our dependencies do.

## Definition of Done

Bump minimum versions of dependencies and change policies around this going forward.

## Tasks

- [ ] Make sure no important downstream tasks are depending on an old version of a dep (there may be some tiledb thing around python 3.8?)
- [ ] Bump the minimum versions
- [ ] Remove CI/ CD infra that is specific to old versions

ivirshup:
Fixes #1177

This PR implements adoption [SPEC-0](https://scientific-python.org/specs/spec-0000/) for support windows of dependencies.

I'm going to hold off on running formatting for all files until I'm done with the manual changes to make this readable.

-----

Blocked by tiledbsoma's pinning of pyarrow to v12 (current release is v16). It looks like it's difficult to install `pyarrow~=12.0` on python 3.12 since there haven't been any wheel releases. See: https://github.com/single-cell-data/TileDB-SOMA/issues/1926


johnkerl:
Not positive it's entirely related but we had a hard fail on MacOS, beyond our control:
https://github.com/single-cell-data/TileDB-SOMA/issues/1926#issuecomment-1834695149

ivirshup:
I've gone ahead and made a few changes here.

Right now I've followed spec-0, but since we can't do 3.12 while tiledbsoma pins it I've just left off support there.

My last commit is just letting the formatter run + a couple manual fixes. I can revert it if other changes are needed.

ivirshup:
This is ready for review. I think the main question is whether we want to drop support for python 3.8 and 3.9 if we can't support 3.12 while tiledbsoma pins it.

ivirshup:
no

.github/workflows/py-dependency-check.yml
```diff
@@ -25,10 +25,10 @@ jobs:
       fail-fast: false  # don't fail-fast, as errors are often specific to a single cell in the matrix
       matrix:
         os: [sc-dev-64g-runner, macos-latest]
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.10", "3.11", "3.12"]
         exclude:
           - os: macos-latest
-            python-version: "3.8"
+            python-version: "3.10"
```
ivirshup, line 31:
```suggestion
            python-version: "3.12"
```

api/python/notebooks/README.md
```diff
@@ -9,7 +9,7 @@ Demonstration notebooks for the CZ CELLxGENE Discover Census. There are two kind
 
 You must be on a Linux or MacOS system, with the following installed:
 
-* Python 3.8 to 3.11
+* Python 3.10 to 3.11
```
ivirshup, line 12:
```suggestion
* Python 3.10 to 3.12
```

PR #1191, opened 2024-06-12
[python] Support custom obs encoders

ebezzi:
Add support for custom obs encoders. Example use case: when multiple obs columns need to be batched together before the tensor creation.


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -164,6 +212,8 @@ def __next__(self) -> _SOMAChunk:
         )
         assert obs_batch.shape[0] == obs_joinids_chunk.shape[0]
 
+        # print("obs_batch", obs_batch)
+
```
atolopko-czi, line 216: rm

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -36,6 +36,54 @@
...
+    @abc.abstractmethod
+    def register(self, obs: pd.DataFrame) -> None:
+        """Register the encoder with obs."""
+        pass
+
+    @abc.abstractmethod
+    def transform(self, df: pd.DataFrame) -> pd.DataFrame:
+        """Transform the obs DataFrame into a DataFrame of encoded values."""
+        pass
```
atolopko-czi, line 58: ⚠️ should also have an `inverse_transform` method, as LabelEncoder provided.. See example usage in `pytorch.ipynb`.  This was a hidden requirement...

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -36,6 +36,54 @@
...
+
+    def transform(self, df: pd.DataFrame) -> pd.DataFrame:
+        return self._encoder.transform(df[self.col])  # type: ignore
+
+    @property
+    def name(self) -> str:
+        return self.col
+
+    @property
+    def classes_(self):  # type: ignore
```
atolopko-czi, line 83: Worth pulling up into `Encoder` base class?

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -480,6 +533,9 @@ def __init__(
             shuffle_chunk_count:
                 The number of contiguous blocks (chunks) of rows sampled to then concatenate and shuffle.
                 Larger numbers correspond to more randomness per training batch.
+            encoders:
+                Specify custom encoders to be used. If not specified, a LabelEncoder will be created and
```
atolopko-czi, line 537:
```suggestion
                Specify custom encoders to be used. If not specified, a DefaultEncoder will be created and
```

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -700,7 +766,8 @@ def obs_encoders(self) -> Encoders:
         self._init()
         assert self._encoders is not None
 
-        return self._encoders
+        # return self._encoders
```
atolopko-czi, line 769: rm

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -640,16 +698,24 @@ def __len__(self) -> int:
...
-                encoders[col] = enc
+        encoders = []
+        obs = query.obs(column_names=self.obs_column_names).concat().to_pandas()
+
+        if self._custom_encoders:
+            # Register all the custom encoders with obs
+            for enc in self._custom_encoders:
+                enc.register(obs)
+                encoders.append(enc)
+        else:
```
atolopko-czi, line 712: ⚠️ Should this still register default encoders for columns that are not the source of any custom encoder? This complicates matters, I realize, since each encoder would then have to publicize its source columns explicitly.  But without it, any `object`-typed columns that do not have a custom encoder will not be usable in a Tensor. Minimally, the class should present a clear error if after transform, any non-numeric columns remain (with a hint to specify an encoder for each unencoded column).

atolopko-czi (changes requested):
Noted two issues that need to be addressed (see ⚠️ comments). Otherwise, just some nits.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ebezzi: Right, I saw that but I forgot! Thanks for the catch!

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ebezzi: My idea here is that, if the user specifies custom encoders, then they are responsible for ensuring that all the columns are transformed properly. With the new changes (see lined 346-347), if a column isn't explicitly registered by the encoder, it won't be added to the final obs_tensor. I believe this addresses this concern, but feel free to let me know if I am still missing something 😄 

That said, I think we should expand the docstrings to better explain the consequences of using custom encoders.

atolopko-czi: In that case, I suppose the `obs_columns` and `encoders` could be mutually exclusive. That is, it would be nice if one could specify just the `encoders` without having to also specify the source `obs_columns`, since this could lead to errors.  But here again, the `Encoder` would need to be able to report what columns it depends on so that the `ExpDataPipe` can retrieve them correctly.

But I also agree this could just be handled via clear documentation and useful error messages.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ebezzi: I am not sure if this is something that necessarily applies to all encoders - I think we can keep this scoped to our DefaultEncoder.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ebezzi: Added.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -36,6 +36,54 @@
...
+        """Register the encoder with obs."""
+        pass
+
+    @abc.abstractmethod
+    def transform(self, df: pd.DataFrame) -> pd.DataFrame:
+        """Transform the obs DataFrame into a DataFrame of encoded values."""
+        pass
+
+    @property
+    def name(self) -> str:
```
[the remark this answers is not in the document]
ebezzi: I agree. I changed it.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ebezzi: I do agree that, if using custom encoders, `obs_columns` could be omitted. There is a slight performance increase in using them, as it doesn't require to fetch the whole `obs` though. Maybe we can just make them optional?

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -36,6 +36,74 @@
...
+
+
+class DefaultEncoder(Encoder):
+    """Default encoder based on LabelEncoder."""
+
+    def __init__(self, col: str) -> None:
+        self._encoder = LabelEncoder()
+        self.col = col
+
+    def register(self, obs: pd.DataFrame) -> None:
```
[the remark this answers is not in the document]
ivirshup: There is also an existing concept of a [`ColumnEncoder`](https://scikit-learn.org/stable/modules/compose.html#column-transformer) which keeps track of column names

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ivirshup: My 2 cents is that this should error if both are passed, and the encoders should be able to report which columns they are going to grab. This links onto my other comment about adopting a scikit-learn like (or maybe even compatible) API, where column transformers and metadata routing are used to retrieve information like column names for their transformers.

ivirshup: In my heart of hearts, I would probably even allow specifying an encoder for `X`. E.g. instead of specifying `use_sparse_X` you could have a `DenseEncoder()`.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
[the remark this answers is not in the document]
ebezzi: I agree that `fit` is a better name. Will change it.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ebezzi: I'm ok with implementing the logic where either the parameter can be defined. I also think that using an encoder for X is a great idea - I'll create a ticket for that as it will go out of scope for this PR.

api/python/cellxgene_census/tests/experimental/ml/test_pytorch.py
```diff
@@ -419,6 +421,7 @@ def test_distributed__returns_data_partition_for_rank(
             measurement_name="RNA",
             X_name="raw",
             obs_column_names=["label"],
+            encoders=[DefaultEncoder("soma_joinid"), DefaultEncoder("label")],
```
ebezzi, line 424: These test use the `soma_joinid` to assert positional conditions, so this is how you force the `soma_joinid` to be part of the encoder.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -492,13 +588,25 @@ def __init__(
...
         self._obs_joinids = None
         self._var_joinids = None
         self._shuffle_chunk_count = shuffle_chunk_count if shuffle else None
         self._shuffle_rng = np.random.default_rng(seed) if shuffle else None
         self._initialized = False
 
+        if obs_column_names and encoders:
+            raise ValueError(
+                "Cannot specify both `obs_column_names` and `encoders`. If `encoders` are specified, columns will be inferred automatically."
+            )
```
atolopko-czi, line 601: 👍

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
atolopko-czi: looks like the obs_columns vs encoder are now mutually exclusive, so the comment is nearly addressed--does the X encoder idea still need an issue created for it? (didn't see one)

api/python/cellxgene_census/tests/experimental/ml/test_pytorch.py
atolopko-czi: Ah, then this point really should be explained in the docstring for `encoders` param.

atolopko-czi: Also would `obs_column_namess=["soma_joinid", "label"]` be equivalent?

atolopko-czi (approved):
LGTM w/a couple of minor comments and maybe a couple more tests:

The various existing tests do exercise both `obs_columns` and `encoders`, but the `encoders` param doesn't have an _explicit_ test. So could add a couple of tests perhaps:
* A test similar to `test_encoders` but the for `encoders` param
* A custom encoder that reads from multiple columns (since that was a motivation for this work)

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -492,13 +588,25 @@ def __init__(
...
+
+        if encoders:
+            # Check if names are unique
+            if len(encoders) != len({enc.name for enc in encoders}):
+                raise ValueError("Encoders must have unique names")
+
+            self.obs_column_names = list(itertools.chain(*[enc.columns for enc in encoders]))
```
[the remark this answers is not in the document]
ebezzi: I'll add a dedup here.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ebezzi: Added: https://github.com/chanzuckerberg/cellxgene-census/issues/1219

api/python/cellxgene_census/tests/experimental/ml/test_pytorch.py
ebezzi: Not the same - `soma_joinid` is ignored because by default only string columns are encoded when using the default behavior. Using custom encoders override this.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -605,7 +524,7 @@ def __init__(
             if len(encoders) != len({enc.name for enc in encoders}):
                 raise ValueError("Encoders must have unique names")
 
-            self.obs_column_names = list(itertools.chain(*[enc.columns for enc in encoders]))
+            self.obs_column_names = list(set(itertools.chain(*[enc.columns for enc in encoders])))
```
ivirshup, line 527: `set`s don't have a consistent iteration order, which can be a source of really hard to figure out bugs. I'd suggest deduplication in an order-preserving way.

Stack overflow says:

```suggestion
            self.obs_column_names = list(dict.fromkeys(itertools.chain(*[enc.columns for enc in encoders])))
```

https://stackoverflow.com/questions/39755464/python-unique-items-of-list-in-the-order-that-it-appears

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/__init__.py
```diff
@@ -1,9 +1,13 @@
...
+from .encoders import BatchEncoder, DefaultEncoder, Encoder
 from .pytorch import ExperimentDataPipe, Stats, experiment_dataloader
 
 __all__ = [
     "Stats",
     "ExperimentDataPipe",
     "experiment_dataloader",
+    "Encoder",
+    "DefaultEncoder",
+    "BatchEncoder",
```
ivirshup, line 12: Should these get visible in the rendered docs? I think they need to be added to `docs/python-api.rst`

ivirshup (approved):
LGTM. I'm not totally sure about the API (e.g. what happens if an encoder returns a non-1d value?), but this can be revisited in future.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/encoders.py
```diff
@@ -0,0 +1,131 @@
...
+        pass
+
+    @property
+    @abc.abstractmethod
+    def columns(self) -> List[str]:
+        """Columns in `obs` that the encoder will be applied to."""
+        pass
+
+
+class DefaultEncoder(Encoder):
```
ivirshup, line 56:
```suggestion
class DefaultEncoder(Encoder):
```

Just a thought, and this is something which is easy to fix with a renaming/ deprecation in the future. Is this a good name?

Why is this called DefaultEncoder? What about it is default?

To me, it's that it's the encoder used by default for columns passed by `obs_column_names`. But I don't think the behavior of a mutually exclusive argument is really the most informative name to give the class. 

Maybe this could be `LabelEncoder` since that name matches the behavior being wrapped?

ebezzi: I agree that `LabelEncoder` is probably a better name.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/__init__.py
ebezzi: Added.

pablo-gar (approved):
Notebook looks good!

Issue #997, opened 2024-02-14
[misc] user-agent tagging


bkmartinjr:
On Python/R API-initiated HTTP requests, set User-Agent tags:
* R/Py environment
* Package version
* Means to add additional "application name" tag (allows discrimination of CI, WMG builder, etc)

This should propagate through:
- HTTP requests to CDN, eg, fetch of release.json, mirrors.json
- TileDB requests to S3
- Other requests, e.g., H5AD fetch via API which is currently using fsspec

To do:
- [ ] Python API - TileDB requests
- [ ] R API - TiledDB requests
- [ ] Python API - all other HTTP requests
- [ ] R API - all other HTTP requests
- [ ] R unit tests set app flag
- [ ] Python unit tests set app flag

ivirshup:
We've merged #1193, which checks a few of the boxes here. I've edited the top comment to show this.

pablo-gar:
Other non-Census usage is now tracked here

https://github.com/chanzuckerberg/single-cell-data-portal/issues/7298

pablo-gar:
R work is now tracked here
https://github.com/chanzuckerberg/cellxgene-census/issues/1249

pablo-gar:
@ivirshup if this has been completed

| Builder fetch of datasets manifest needs to set header

Then we can close this ticket

PR #1193, opened 2024-06-12
[python] set user-agent

ivirshup:
This PR sets a user agent on requests from the python api, addressing part of #997

* Adds a new user agent to all requests made by this library. Format is `cellxgene-census-python/{VERSION}`
* Defines a new environment variable `CELLXGENE_CENSUS_USERAGENT` whose value will be added to the user agent if defined. This allows us to set the flag on internal systems and figure out internal vs external usage of the resource.
* Tests for all this are in `test_user_agent`. This is the most complicated part.
  * For each test session we set up an ephemeral proxy server that intercepts and logs all requests
  * We turn off SSL verification for all requests so that we can do the interception
  * This is exposed through a fixture that gives access to the logged requests

Could do

- [ ] Some centralization of environment variables
- [ ] Attention payed to mypy. 

Most of the typing problems are in the tests. I think there's very little value to adding more type info here, and it gets quite complicated since most of it is in fixtures. I don't think doing much more is worth it, so would like to ignore these.

Probably should do:

- [x] Set a custom user agent in the testing environment (though the version number should be `.dev` for all of these). Idk what this should be.


ivirshup:
For test coverage on the `logger_proxy.py` file, that code isn't executed by the test process so I wouldn't expect coverage.

api/python/cellxgene_census/src/cellxgene_census/_testing/logger_proxy.py
```diff
@@ -0,0 +1,36 @@
+"""This module defines a plugin class that logs each request to a logfile."""
```
mlin, line 1: It seems like py.proxy necessitates including this under the main source tree, is that correct? If so, please add a bit of explanation in this docstring (or a README.md) for posterity.

api/python/cellxgene_census/tests/test_user_agent.py
```diff
@@ -0,0 +1,265 @@
+from __future__ import annotations
```
mlin, line 1:
```suggestion
# mypy: ignore-errors
from __future__ import annotations
```

api/python/cellxgene_census/src/cellxgene_census/_testing/logger_proxy.py
ivirshup: I believe it has to be importable, so in a module that is on the python path. I don't feel super attached to having it here, so am open to other suggestions if you think there is a better way to do this.

```suggestion
"""This module defines a plugin class that logs each request to a logfile.

This class needs to be importable by the proxy server which runs in a separate process.
See the user agent tests for usage.
"""
```

ebezzi (approved):

PR #1200, opened 2024-06-21
[builder] Add collection_doi_label to the datasets dataframe

### LINKED ISSUE #1180: Implement `collection_doi_label` in builder
- [ ] Implement in builder #1179 
- [ ] Modify schema as indicated in #1179

ebezzi:
Fixes #1180


tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/manifest.py
```diff
@@ -90,6 +90,7 @@ def load_manifest_from_CxG() -> list[Dataset]:
             collection_id=dataset["collection_id"],
             collection_name=null_to_empty_str(dataset["collection_name"]),
             collection_doi=null_to_empty_str(dataset["collection_doi"]),
+            collection_doi_label=null_to_empty_str(dataset.get("collection_doi_label")),
```
prathapsridharan, line 93: Here you are using `dataset.get("...")` which does simply returns `None` if the key doesn't exist - that is, it does not raise an exception. The other calls use `dataset["..."]` which raise on exception if the key doesn't exist.

Should this semantics be made consistent? Maybe the other key lookups in `null_to_empty_str()` should use `dataset.get(...)` instead?

prathapsridharan (changes requested):

prathapsridharan (approved):

PR #1223, opened 2024-07-03
[ci] Test GH concurrency

ebezzi:


ivirshup:
I think I was wrong and failures are unrelated to the concurrency rule. This is what an action cancelled by the concurrency rule looks like:

* https://github.com/chanzuckerberg/cellxgene-census/actions/runs/9788168759/job/27026703847

It gets a grey icon instead of a red one.

-----

Have you ever seen this happen on a Mac job? If it's only happening on the self hosted runners (i.e. running on custom AWS instances), maybe the jobs are being pre-empted?

ebezzi:
Closing as the problem is not related to concurrency, but rather to an issue with our GH runners.

PR #1224, opened 2024-07-03
[python] dataloader optimization (picking up 1169)

ivirshup:
Building on top of #1169

This PR removes the new argument `method` (which was undocumented and had no tests atm) and instead using `return_sparse_X` to control conversion. Previously `return_sparse_X` controlled whether densification happened in a wrapping iterator, while now a dense array is created without having to make a CSR intermediate.


api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -36,6 +37,10 @@
 The Tensors are rank 1 if ``batch_size`` is 1, otherwise the Tensors are rank 2."""
 
 
+# "Chunk" of X data, returned by each `Method` above
+ChunkX = Union[npt.NDArray[Any], sparse.csr_matrix]
```
bkmartinjr, line 41: nit (optional): why `npt.NDArray[Any]` and not `npt.NDArray[np.number]` (or something even more specific?)

bkmartinjr: Also, is the API signature really saying that it could return _any_ sparse type?  If it is actually constrained to a more specific set of sub-types (e.g, csr_matrix), it would be useful to further specify this type alias (even if just for the static type checking...)

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -101,6 +106,19 @@ def _open_experiment(
         yield exp
 
 
+def _tables_to_np(
+    tables: Iterator[Tuple[Table, Any]], shape: Tuple[int, int]
+) -> typing.Generator[Tuple[npt.NDArray[Any], Any, int], None, None]:
+    for tbl, indices in tables:
+        row_indices_np = np.array(tbl.columns[0])
```
bkmartinjr, line 113: This creates a copy of the Arrow array which is wasteful/unnecessary.

Recommend instead doing a zero-copy conversion:
> row_indices_np = tbl.columns[0].to_numpy()

Same issue for all three columns.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -173,18 +193,28 @@ def __next__(self) -> _SOMAChunk:
...
-            self.X.read(coords=(obs_joinids_chunk, self.var_joinids))
-            .blockwise(axis=0, size=len(obs_joinids_chunk), eager=False)
-            .scipy(compress=True)
+        blockwise_iter = self.X.read(coords=(obs_joinids_chunk, self.var_joinids)).blockwise(
+            axis=0, size=len(obs_joinids_chunk), eager=False
         )
-        X_batch, _ = next(scipy_iter)
+
+        if not self.return_sparse_X:
+            batch_iter = _tables_to_np(blockwise_iter.tables(), shape=(obs_batch.shape[0], len(self.var_joinids)))
```
bkmartinjr, line 201: IMHO, this motivates a dense option on the underlying blockwise iterator (i.e, in addition to the Table and scipy options). I don't think we should hold up this PR, but ideally just file a feature request GH issue in the TileDB-SOMA repo.

@ryan-williams - thoughts?

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
```diff
@@ -173,18 +193,28 @@ def __next__(self) -> _SOMAChunk:
...
+        else:
+            batch_iter = blockwise_iter.scipy(compress=True)
+
+        res = next(batch_iter)
+        X_batch: ChunkX = res[0]
+        if isinstance(X_batch, np.ndarray):
```
bkmartinjr, line 207: nit: I would make this conditional on the same value used to select the iterator (e.g., `self.return_sparse_X`), and assert it is an ndarray.

Just for clarity - no other reason (i.e, the above code works AFAIK)

bkmartinjr (commented):
Overall looks good - no major concerns. A few minor points of clarity and refinment noted inline.

In at least one case, there is a redundant data copy where zero-copy would suffice. Recommend tracing through the code to ensure no others (perhaps after this PR lands?), as reducing memory pressure is important in this use case.

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ivirshup: Yeah, I wasn't totally happy with the logic on this.

The type of the value should be invariant with the value of `return_sparse_X`. But that's not how it was in the original PR or the previous code. But I can go for consistency with other branches here.

ivirshup: Its sorta weird that we make an iterator here but only ever use the first iteration

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ivirshup: 👍

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ivirshup: Hmm. This is opening up a big can of `typing` worms. There wouldn't be as much of an issue once we drop support for python 3.8.

I think I will go back to `Any` until we move to SPEC-0

ebezzi (approved):

api/python/cellxgene_census/src/cellxgene_census/experimental/ml/pytorch.py
ryan-williams: Makes sense to me; we'd basically just move `_tables_to_np` into TileDB-SOMA?

PR #1234, opened 2024-07-10
[python] Reintroduce embeddings tests

ebezzi:
Tests were adapted to account for temporary lack of embeddings in the new LTS. Now that they have been published, tests can be reintroduced.

### DIFF
--- a/api/python/cellxgene_census/tests/test_get_anndata.py
+++ b/api/python/cellxgene_census/tests/test_get_anndata.py
@@ -195,7 +195,7 @@ def test_get_anndata_obs_embeddings(lts_census: soma.Collection, obs_embeddings:
         obs_embeddings=obs_embeddings,
     )
 
-    assert len(ad.obsm.keys()) == 3
+    assert len(ad.obsm.keys()) == 2
     assert len(ad.varm.keys()) == 0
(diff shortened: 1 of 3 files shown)


api/python/cellxgene_census/tests/test_get_anndata.py
```diff
@@ -197,7 +195,7 @@ def test_get_anndata_obs_embeddings(lts_census: soma.Collection, obs_embeddings:
         obs_embeddings=obs_embeddings,
     )
 
-    assert len(ad.obsm.keys()) == 3
+    assert len(ad.obsm.keys()) == 2
```
ivirshup, line 198: What is the behavior if we generate multiple scvi embeddings for a release? And do we plan to add embeddings to old LTS releases?

I'm wondering if this test should just run on the older release.

ivirshup (approved):

api/python/cellxgene_census/tests/test_get_anndata.py
ebezzi: We don't really support having multiple embeddings per release. In this case, we would yank the previous version and replace it with the new one.

PR #1235, opened 2024-07-10
[python] Bump tiledbsoma version to 1.12.3

ivirshup:
Bump `tiledbsoma` dep to 1.12.3.

I believe the remaining deprecations being thrown are actually coming from `tiledbsoma`, since this PR removes all use of `tiledb_ctx`.


ivirshup:
@ebezzi were there other checks you wanted to run before bumping this?

ebezzi (approved):

PR #1244, opened 2024-07-12
Fix LTS compat check bug

ebezzi:


ivirshup (approved):
I think this will work, but my `bash` isn't that great.

ivirshup (commented):
@ebezzi, I tried running the check again on this branch, but it looks like the same error is occurring:

* https://github.com/chanzuckerberg/cellxgene-census/actions/runs/9944509425/job/27470721380

Any idea what's up there? It looks like it's pointing at the right version of the workflow file.

.github/workflows/lts-compat-check.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           python -m pip install -U pip setuptools wheel
           GIT_CLONE_PROTECTION_ACTIVE=false pip install -r ./api/python/cellxgene_census/scripts/requirements-dev.txt
 
-          if [ {{matrix.py-pkg-version}} = "head-of-main" ]; then
+          if [ {{matrix.py-pkg-version}} == "head-of-main" ]; then
```
ivirshup, line 48:
```suggestion
          if [ ${{matrix.py-pkg-version}} == "head-of-main" ]; then
```

PR #1252, opened 2024-07-29
[docs] Remove similarity search notebook

ebezzi:
This notebook isn't ready for the docsite yet.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -27,7 +27,6 @@ Access Census embeddings.
    
     notebooks/api_demo/census_access_maintained_embeddings.ipynb
     notebooks/api_demo/census_embedding.ipynb
-    notebooks/api_demo/census_embedding_search.ipynb
     
 Use the Census trained models.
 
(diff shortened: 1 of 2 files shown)


ivirshup (changes requested):
I think this would also need to be removed from `docs/examples.rst:30`

Would it make sense to just revert and reopen the PR that added this? It would also be good to make sure that re-adding this was tracked for the next release.

cc: @pablo-gar

ebezzi:
> I think this would also need to be removed from `docs/examples.rst:30`
> 
> Would it make sense to just revert and reopen the PR that added this? It would also be good to make sure that re-adding this was tracked for the next release.
> 
> cc: @pablo-gar

Thanks, forgot about that.

I think it makes sense to keep the notebook if we don't link it in the docsite.

ivirshup:
As discussed in person, we can probably do a doc build from this PR and not need to re-add things later.

ebezzi:
Closing this - the branch was enough for patching the docsite.

PR #1257, opened 2024-08-08
[python] similarity search API: optimize predict_obs_metadata

mlin:
diff is to shrink once https://github.com/chanzuckerberg/cellxgene-census/pull/1164 merges to main


mlin:
@ivirshup It looks like the last thing we fixed helped but [broke on the python 3.8 runner only](https://github.com/chanzuckerberg/cellxgene-census/actions/runs/10308890286/job/28537326318#step:7:335), maybe b/c of an older numpy/scipy version?

ivirshup:
We'll drop 3.8 support in the next release so I think it's fine.

ivirshup (approved):

PR #1292, opened 2024-10-01
[Python] Correct docs on default behaviour for DataLoader no longer including soma_joinid

### LINKED ISSUE #1291: Fix docs for data loaders
In https://github.com/chanzuckerberg/cellxgene-census/issues/1289#issuecomment-2386551003 @yanwu2014 noted that the docs still say we are returning the `soma_join_ids` per batch in all cases, even though we removed that behavior. We can see that here:

image

We should fix that.

ivirshup:
Fixes #1291

A couple places in the docs still said we include the join id by default. I think this catches all the remaining places.


ebezzi (approved):
LGTM, thanks Isaac for spotting the issue.

PR #1294, opened 2024-10-02
Static R docsite

ebezzi:
Restores the R static docsite.


ivirshup (commented):
@ebezzi, this looks like an older version of the R docs than what we have up currently.

Should we just grab the R docs from the current gh-pages branch instead?

image

image

ebezzi:
> @ebezzi, this looks like an older version of the R docs than what we have up currently.
> 
> Should we just grab the R docs from the current gh-pages branch instead?
> 
> image image

I used an outdated branch. I pushed an update.

ivirshup (approved):

PR #1319, opened 2024-11-25
tools/models/scvi docs adjustments

mlin:
A few minor updates to the scvi docs; mainly, I hit weird ImportErrors without using a clean virtualenv.

### DIFF
--- a/tools/models/scvi/scvi-init.sh
+++ b/tools/models/scvi/scvi-init.sh
@@ -26,8 +26,7 @@ python3 -m venv scvi_venv
 source scvi_venv/bin/activate
 
 pip install --upgrade "jax[cuda11_pip]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
-pip install pathlib torch click ray hyperopt
-pip install git+https://github.com/scverse/scvi-tools.git
+pip install torch click ray hyperopt scvi-tools
 
 
 # pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu118


tools/models/scvi/scvi-init.sh
```diff
@@ -22,7 +22,10 @@ sudo cp /usr/lib/python3/dist-packages/apt_pkg.cpython-38-x86_64-linux-gnu.so /u
 
 sudo apt -y install libnvidia-gl-535 libnvidia-common-535 libnvidia-compute-535 libnvidia-encode-535 libnvidia-decode-535 nvidia-compute-utils-535  libnvidia-fbc1-535 nvidia-driver-535
 
-ipip install --upgrade "jax[cuda11_pip]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
+python3 -m venv scvi_venv
+source scvi_venv/bin/activate
+
+pip install --upgrade "jax[cuda11_pip]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
 pip install pathlib torch click ray hyperopt
 pip install git+https://github.com/scverse/scvi-tools.git
```
ivirshup, line 30: Not really the topic of this PR, but:

* Should we at least install a released version of scvi? If not pinning the version of scvi?
* `pathlib` is in the standard library since `3.4`. The [`pathlib` on pypi](https://pypi.org/project/pathlib/) is a currently unmaintained backport. 

```suggestion
pip install torch click ray hyperopt scvi-tools
```

ivirshup (commented):
One minor comment, but generally looks good. Would like @ebezzi to take a look before merging though.

Any chance you have the traceback from those `ImportError`s?

tools/models/scvi/scvi-init.sh
ebezzi: Yeah, we can probably replace it with just `pip install scvi-tools`. I believe that's what we did for the last run. We can also remove `ray` and `hypteropt` as they're not used by any of these scripts.

mlin:
@ivirshup I found one in my notes, `ImportError: cannot import name 'html5lib' from 'pip._vendor'`. I presume it was something to do with using a different python version than shipped with ubuntu 20.04, but I didn't look too closely once I found it worked in the virtualenv.

PR #1320, opened 2024-11-26
Fix presence matrix calculation (and derived values)

### LINKED ISSUE #1284: feature_dataset_presence_matrix should include genes with 0 counts
## Describe the bug

The current `feature_dataset_presence_matrix` only reports genes that have a total expression count > 0. Instead, it should report  all genes that were listed in the original dataset.

## To Reproduce

For dataset 0895c838-e550-48a3-a777-dbcd35d30272, 13696 genes are nnz

## Expected behavior

For dataset 0895c838-e550-48a3-a777-dbcd35d30272, 33363 should be nnz

ivirshup:
Fixes #1284

This PR addresses issues with presence matrix calculation. Previously only genes with non-zero values were included in the presence matrix, while the intent was to include all genes recorded in the original anndata.

Previously, the test cases had all non-zero values. This meant that the "non-zero" and "present" genes overlapped completely. We now include some zeros in the test cases, which caused the existing tests to fail before I implemented a fix.

Evidence this works: https://gist.github.com/ivirshup/4c09193e30534bdcfe9710efed4f2b16

----

I think there is also a broader code cleanup that could happen. Broadly, I find the statistics calculation hard to follow (many classes, deferred calculations). But what could be in scope is cleaning up unused attributes, classes, and code related to presence matrix construction.


ivirshup:
So, I was adding a little more validation to https://github.com/chanzuckerberg/cellxgene-census/pull/1320 and ended up in an odd situation. The validation I added seemed to contradict some existing validation code which was (AFAICT) validating the previous incorrect state. Yet: no validation errors.

After banging my head against it, I paired with @ebezzi and we figured out that [this line](https://github.com/chanzuckerberg/cellxgene-census/blob/fbb00fee91f3bbb5c6a0e7b640e9b0b84bd43cb5/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/validate_soma.py#L580) ends up meaning that this whole validation step is never actually run (though it is logged as running, it just ends up being a no-op).

This is now fixed in https://github.com/chanzuckerberg/cellxgene-census/pull/1320/commits/4a7ed8f3891dccb9b1a2f00a38c1bada31f21685

ebezzi (approved):

PR #1323, opened 2024-12-07
Fix reading manifest file when it's only one line and also served by GitHub

ivirshup:
For some reason changing the size of our manifest file has broken our code to read it. It has something to do with how the file is encoded. I believe it is related to https://github.com/fsspec/filesystem_spec/issues/389.

The issue only comes up due to some handling of partial reads. So I switched to doing a full read since the file is never that large.

We use fsspec instead of requests since sometimes we read a local file or one from s3.

### DIFF
--- a/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/manifest.py
+++ b/tools/cellxgene_census_builder/src/cellxgene_census_builder/build_soma/manifest.py
@@ -111,15 +111,18 @@ def load_blocklist(dataset_id_blocklist_uri: str | None) -> set[str]:
         logger.error(msg)
         raise ValueError(msg)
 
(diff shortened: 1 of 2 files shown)


ebezzi (approved):

ivirshup:
Cleaner version:

```python
blocklist = (
    # Figure out protocol to open file
    fsspec.filesystem(
        fsspec.utils.infer_storage_options(dataset_id_blocklist_uri)["protocol"]
    )
    # Read whole file as bytes
    .cat_file(dataset_id_blocklist_uri)
    # Decode to string and split into lines
    .decode("utf-8").strip().split("\n")
)
```

ivirshup:
I did not touch the code that mypy is complaining about so I am okay with skipping it for this PR

PR #1343, opened 2025-01-24
Allow numpy 2

### LINKED ISSUE #1240: Support numpy 2
We should try to remove the pin on numpy 2

ivirshup:
Fixes #1240

Unpinning numpy 2 since the test suite now runs fine with numpy 2.1.3.


ivirshup:
Hmm, that didn't fail so badly locally. Will investigate further

ebezzi (approved):

PR #1349, opened 2025-02-03
advance to tiledb-vector-search~=0.11

mlin:


mlin:
@ivirshup If you see fit to include this in new package release, it may save us from needing another patch release for the embeddings. (Feel free to merge without me)

FTR, TileDB-Vector-Search 0.11 uses TileDB 2.27, which matches the version used in tiledbsoma 1.15.

ivirshup (approved):

PR #1376, opened 2025-03-19
Try using new managed runners

ivirshup:
Our self-hosted runners seem busted. In this PR I am attempting to switch over to some more centrally managed runners.

### DIFF
--- a/.github/workflows/docsite-build-deploy.yml
+++ b/.github/workflows/docsite-build-deploy.yml
@@ -13,7 +13,7 @@ env:
 jobs:
   build-and-deploy:
     concurrency: ci-${{ github.ref }}
-    runs-on: sc-dev-64g-runner
+    runs-on: x64
     steps:
       - name: Checkout 🛎️
         uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


mlin:
@ivirshup on the self-hosted runner is "No space left on device", is there some accumulation of junk we can just clear out, or is the bustedness more profound than that? sorry i'm not up to speed on this

ivirshup:
@mlin, we could definitely do something about that!

I'm not immediately seeing that error, could you point me to where you're seeing that?

mlin:
@ivirshup This is what I'm seeing on #1365:

- https://github.com/chanzuckerberg/cellxgene-census/actions/runs/13978546584/job/39138295841?pr=1365#step:5:469
- https://github.com/chanzuckerberg/cellxgene-census/actions/runs/13978546584/job/39138294782?pr=1365#step:4:459

I don't have a clue about how to get into the hosted runner, but I wouldn't mind seeing what I can do if pointed in the right direction (cc @ebezzi)

Issue #1378, opened 2025-03-26
take Geneformer out of automatic CI


mlin:
The Geneformer dependency occasionally causes CI breakage with its complicated dependency chain. It should be fine to run its tests on an as-needed basis, since we mainly just use it for LTS releases. Disable from our GitHub CI, delete CI steps that try to install Geneformer, and add a few readme notes on how to run the tests at the command line when needed.

mlin:
@ivirshup fyi

ivirshup:
Btw, I am running into more problems with geneformer. The pinned commit now needs pinning of the `transformers` library, since the `AdamW` class has been removed

Here's an issue describing the problem from `transformers`:

* https://github.com/huggingface/transformers/issues/36954

Here's the test failure on our side. Since we clobber the import error here it's not easy to see the root cause.

* https://github.com/chanzuckerberg/cellxgene-census/actions/runs/14095286993/job/39481281917

But if you remove the clobbering, you can see that import geneformer with `transformers>=4.50` throws:

```pytb
ImportError: cannot import name 'AdamW' from 'transformers' (/home/ubuntu/miniforge3/envs/census-api-dev-experimental/lib/python3.11/site-packages/transformers/__init__.py)
```

PR #1381, opened 2025-03-27
take Geneformer out of CI

mlin:
Mark Geneformer tests to disable them by default, and avoid installing Geneformer in CI due to dependency issues (#1378).


mlin:
@ivirshup updated as we discussed, with deprecation notices and still-fewer dependencies. I will make a separate PR next week to add the code & tests to tools/models/geneformer.

ivirshup (approved):

Issue #1373, opened 2025-03-17
Incorrect number of donors in `census["census_info"]["summary"]`


ivirshup:
## Describe the bug

For the latest LTS, the total number of donors is incorrectly reported in the `census["census_info"]["summary"]` table.

## To Reproduce

```python
import cellxgene_census

census = cellxgene_census.open_soma(census_version="2025-01-30")
census["census_info"]["summary"].read().concat().to_pandas()
```

|    |   soma_joinid | label                      | value                                      |
|---:|--------------:|:---------------------------|:-------------------------------------------|
|  0 |             0 | census_schema_version      | 2.1.0                                      |
|  1 |             1 | census_build_date          | spatial-build-rc3-io2-2025-01-30T21:20:53Z |
|  2 |             2 | dataset_schema_version     | 5.2.0                                      |
|  3 |             3 | total_cell_count           | 154437194                                  |
|  4 |             4 | unique_cell_count          | 85809959                                   |
|  5 |             5 | number_donors_homo_sapiens | 390                                        |
|  6 |             6 | number_donors_mus_musculus | 172                                        |

```python
len(
    census["census_data"]["homo_sapiens"]
    .obs.read(column_names=["donor_id"])
    .concat().to_pandas()
    ["donor_id"]
    .unique()
)
```

```
8510
```

## Expected behavior

The number should be at least 8510, though maybe should be greater than that if we want to include spatial and single cell donors together.

## Additional context

Originally noticed while looking at: https://github.com/chanzuckerberg/cellxgene-census/pull/1356

ivirshup:
We have discussed internally.

These numbers are definitely wrong and while we can report something *more* accurate (at least group by collection), we probably cannot accurately report this number since we simply don't know accurately when donors are shared across studies.

We don't know why this metric was originally included, but was likely due to some measurements that were taken to measure velocity of data being added to census. We are not using these fields from census for anything downstream (e.g. analytics).

Proposed plan of action:

- [ ] Remove the incorrect donor counts from the [census data release](https://chanzuckerberg.github.io/cellxgene-census/cellxgene_census_docsite_data_release_info.html) page
- [ ] Add an [errata statement](https://chanzuckerberg.github.io/cellxgene-census/cellxgene_census_docsite_data_release_info.html#errata) saying this field had been calculated incorrectly
- [ ] Stop the builder from adding this field in future
	- [ ] Update spec to remove the field

cc: @brianraymor @BAevermann @jahilton

PR #1382, opened 2025-03-27
Remove summarized donor count from docs and builds

ivirshup:
Addressing part of https://github.com/chanzuckerberg/cellxgene-census/issues/1373

This PR removes the summarized donor counts fields from builds and docs.

It looks like we missed a bump on census version number, so that needs to be corrected. We also need to decide what level of version bump this change is.

We should also re-run notebooks to point at newer versions of census so that the fields don't show up when loading older versions.


mlin:
@ivirshup If it's alright with you, I'm planning to adopt this into "increment 2" of the work @brianraymor and I are currently doing toward census schema 2.4.0. Aside from what you already mentioned above, is there anything else you'd want me to know about it? Thanks!

Issue #1092, opened 2024-04-05
Draft Census schema for support of Visium and Spatial data 


pablo-gar:
**LAST EDITED: Aug, 29, 2024**

See parent Epic for further information. 
https://github.com/chanzuckerberg/single-cell/issues/644

See current draft for spatial support in SOMA https://docs.google.com/document/d/1S48pD5XTzDcaLGlq6YVYCoUjptR93PHHHmG79TiJzsA/edit

# TODOs

- [ ] Change all version mentions of datasset schema to 5.1.0
- [ ] Update `./census_accepted_assays.csv` to include:
   - [ ] ` EFO:0010961` - `Visium Spatial Gene Expression`
   - [ ]  `EFO:0030062` - `Slide-seqV2` 
   - [ ] ` EFO:0009920` - `Slide-seq` maybe?
- [ ] Define a diameter for slide-seq data to be stored in `spatial[scene_id].obsl["loc"]["soma_geometry"]`
- [ ] Define the units of the positions data frame for slide-seq (visium is pixels)

# Schema changes

-----

**Version**: 2.2.0

**Last edited**: April, 2024.

-----

### Data included

All datasets included in the Census MUST be of [CELLxGENE dataset schema version 5.1.0](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.1.0/schema.md). The following data constraints are imposed on top of the CELLxGENE dataset schema.

> _**Editor's note:** do this change in all other places where the CELLxGENE dataset schema version is mentioned. For simplicity all other changes are omitted here._

-----

#### Assays

[...]

The Census MUST include all cells from the list of [accepted assays](./census_accepted_assays.csv).

These assays were selected with the following criteria:

> Only children "EFO:0002772" or "EFO:0010183" are shown as this is a constraint imposed by the CELLxGENE dataset schema >3.0.0.
>
> * Must measure gene expression via RNA sequencing.
> * Can be done at the single-cell level.
> * May include nascent or elongating RNA data.
> * May be targeted to specific genes in an assay-specific manner.
> * May include spatial data only from Visium or Slide-seq.
> * Doesn’t measure spatial data from other assays,
> * Doesn't measure other non-RNA molecules concurrently.
> * Doesn’t require author metadata for correct interpretability (e.g. perturbation-based technologies).
> * Doesn’t intend to primarily measure RNA structure, RNA fusions, RNA modifications, or RNA interactions.
[...]

PR #1387, opened 2025-04-02
Add spatial to census schema

ivirshup:
Based on the schema from Pablo in https://github.com/chanzuckerberg/cellxgene-census/issues/1092, but updated in the following ways:

* `obs_scene` is now `obs_presence_matrix`
	* Also this was described a little inconsistently, but that's been corrected
* Fixed typo, only one matrix is in `X`
* There are a few TODOs left
	* Slide-seq radius: This was left as a default value from Prathaps old PR. Discussed with @mlombardo, but we didn't have any better ideas.
* Locations are stored as `PointCloudDataFrame`
	* Radius is now stored in metadata
* We ended up not including the full res image, since we don't actually always know if it's aligned to the fullres image/ coordinates.
	* I removed the "MAY include fullres". Let me know if we should keep it.
* We also ended up keying images by library id, so that we don't need to mangle the names on export to spatialdata.

### DIFF
--- a/docs/cellxgene_census_schema.md
+++ b/docs/cellxgene_census_schema.md
@@ -1,8 +1,8 @@
 # CZ CELLxGENE Discover Census Schema
 
-**Version**: 2.2.0
+**Version**: 2.3.0
 
-**Last edited**: January, 2025.
+**Last edited**: April, 2025.
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED" "MAY", and "OPTIONAL" in this document are to be interpreted as described in [BCP 14](https://tools.ietf.org/html/bcp14), [RFC2119](https://www.rfc-editor.org/rfc/rfc2119.txt), and [RFC8174](https://www.rfc-editor.org/rfc/rfc8174.txt) when, and only when, they appear in all capitals, as shown here.
 
@@ -126,8 +126,9 @@ These assays were selected with the following criteria:
 > * Can be done at the single-cell level.
 > * May include nascent or elongating RNA data.
 > * May be targeted to specific genes in an assay-specific manner.
+> * May include spatial data only from Visium or Slide-seq.
+> * Doesn’t measure spatial data from other assays,
 > * Doesn't measure other non-RNA molecules concurrently.
-> * Doesn’t measure spatial information.
 > * Doesn’t require author metadata for correct interpretability (e.g. perturbation-based technologies).
 > * Doesn’t intend to primarily measure RNA structure, RNA fusions, RNA modifications, or RNA interactions.
 > * Doesn’t intend to primarily measure non-mRNA (e.g. tRNA, rRNA, small RNAs).
@@ -140,6 +141,15 @@ These assays were selected with the following criteria:
 From the list of accepted assays, this list of [full-gene sequencing assays](./census_accepted_assays_full_gene.csv) are those that when used at the single-cell level will always perform full-gene sequencing.
 
 These data need to be normalized by gene length for downstream analysis.
+
+##### Spatial Assays
+
+Only observations from Visium and Slide-seq assays MUST be included in Census, as indicated in the list of [accepted assays](./census_accepted_assays.csv).  Per the CELLxGENE dataset schema, datasets with spatial observations can be identified with the presence of the slot `uns["spatial"]`. For these assays, only observations from datasets that contain "one Space Ranger output for a single tissue section" MUST be included in Census.
+
+The full logic above can be asserted as follows:
+
+* If a dataset has the slot `uns["spatial"]` and `True` in `uns["spatial"]["is_single"]`, then all observations MUST be included.
+* If a dataset has the slot `uns["spatial"]` and `False` in `uns["spatial"]["is_single"]`, then all observations MUST be excluded.
 
 #### Data matrix types
 
@@ -198,10 +208,10 @@ This `SOMADataFrame` MUST have the following rows:
 3. Dataset schema version:
    1. label: `"dataset_schema_version"`
    2. value: The CELLxGENE Discover schema version of the source H5AD files.
-4. Total number of cells included in this Census build:
+4. Total number of cells or spatial spots included in this Census build:
    1. label: `"total_cell_count"`
    2. value: Cell count
-5. Unique number of cells included in this Census build (is_primary_data == True)
+5. Unique number of cells or spatial spots included in this Census build (is_primary_data == True)
    1. label: `"unique_cell_count"`
    2. value: Cell count
 6. Number of human donors included in this Census build. Donors are guaranteed to be unique within datasets, not across all Census.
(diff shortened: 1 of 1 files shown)


docs/cellxgene_census_schema.md
```diff
@@ -1,8 +1,8 @@
 # CZ CELLxGENE Discover Census Schema
 
-**Version**: 2.2.0
+**Version**: 2.3.0
```
brianraymor, line 3: I don't feel strongly about this - but note that the semver explanation in the census schema mentions "new modality" as a major schema change ...

Major: any schema changes that make the Census incompatible with the Census API or SOMA API. Examples:
Column deletion in Census obs
**Addition of new modality**

docs/cellxgene_census_schema.md
```diff
@@ -126,8 +126,9 @@ These assays were selected with the following criteria:
 > * Can be done at the single-cell level.
 > * May include nascent or elongating RNA data.
 > * May be targeted to specific genes in an assay-specific manner.
+> * May include spatial data only from Visium or Slide-seq.
+> * Doesn’t measure spatial data from other assays,
```
brianraymor, line 130: Editorial.

Editorial. assays, -> assays.

or you could cut line 130 and rewrite line 129 to:

May include spatial data exclusively from Visium or Slide-seq.

130 reads a bit redundant to me.

docs/cellxgene_census_schema.md
```diff
@@ -141,6 +142,15 @@ From the list of accepted assays, this list of [full-gene sequencing assays](./c
...
 
+##### Spatial Assays
+
+Only observations from Visium and Slide-seq assays MUST be included in Census, as indicated in the list of [accepted assays](./census_accepted_assays.csv).  Per the CELLxGENE dataset schema, datasets with spatial observations can be identified with the presence of the slot `uns["spatial"]`. For these assays, only observations from datasets that contain "one Space Ranger output for a single tissue section" MUST be included in Census.
+
+The full logic above can be asserted as follows:
+
+* If a dataset has the slot `uns["spatial"]` and `True` in `uns["spatial"]["is_single"]`, then all observations MUST be included.
+* If a dataset has the slot `uns["spatial"]` and `False` in `uns["spatial"]["is_single"]`, then all observations MUST be excluded.
+
```
brianraymor, line 153: There is only support for Visium and Slide-seq in the dataset schema, so it's unclear why uns['spatial'] is the identifier versus the actual assay_ontology_term_id?

In addition, the example of "one Space Ranger output" is only specific to Visium. See the language related to pucks for Slide-seq in https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.2.0/schema.md#is_single:

<li> if <code>assay_ontology_term_id</code> is <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0010961"><code>"EFO:0010961"</code></a> for <i>Visium Spatial Gene Expression</i> and the dataset represents one Space Ranger output for a single tissue section
      </li>
      <li> if <code>assay_ontology_term_id</code> is <a href="https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0030062"><code>"EFO:0030062"</code></a> for <i>Slide-seqV2</i> and the dataset represents the output for a single array on a puck </li>
<br>
You could simply reuse this language here?

docs/cellxgene_census_schema.md
ivirshup: I saw that, but I put the emphasis on **incompatible** as the qualifier. We did this in a way that does not break anything. I read the example as having the assumption that adding a new modality would require a breaking change.

docs/cellxgene_census_schema.md
ivirshup: How about:

Only observations from Visium and Slide-seq assays MUST be included in Census, as indicated in the list of [accepted assays](./census_accepted_assays.csv).  Per the CELLxGENE dataset schema, datasets with spatial observations can be identified with the presence of the slot `uns["spatial"]`. For these assays, observations from datasets that contain more than one tissue section MUST NOT be included in Census.

The full logic above can be asserted as follows:

* if `assay_ontology_term_id` is [`"EFO:0010961"`](https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0010961) for Visium Spatial Gene Expression and the dataset represents one Space Ranger output for a single tissue section (e.g.the dataset has `True` in `uns["spatial"]["is_single"]`)
* if `assay_ontology_term_id` is [`"EFO:0030062"`](https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0030062) for Slide-seqV2 and the dataset represents the output for a single array on a puck (e.g. the dataset has `True` in `uns["spatial"]["is_single"]`)

docs/cellxgene_census_schema.md
ivirshup: Made this change in 12a467602b47

docs/cellxgene_census_schema.md
```diff
@@ -872,8 +881,185 @@ Cell metadata MUST be encoded as a `SOMADataFrame` with the following columns:
...
+  * Feature dataset presence matrix – `census_obj["census_spatial_sequencing"][organism].ms["RNA"]["feature_dataset_presence_matrix"]` – `SOMASparseNDArray`
+* Obs to spatial data mapping:
+  * Obs to spatial data – `census_obj["census_spatial_sequencing"][organism].obs_spatial_presence`. It indicates the link between an observation and a scene, it MUST have three columns: 1) `soma_joinid` corresponding to `soma_joinid` of `obs` and 2) `scene_id` corresponding to the associated scene and 3) `data` which is a boolean value indicating presence.
+* Spatial data  –  `census_obj["census_spatial_sequencing"][organism].spatial` – `SOMACollection`.
+  * Spatial Scenes with spatial data  –  `census_obj["census_spatial_sequencing"][organism].spatial[scene_id]`  – `SOMAScene`.  There will be as many as Spatial Scenes as  spatial datasets. Each`SOMAScene` MUST contain the following:
+    * MUST contain a positions array – `census_obj["census_spatial_sequencing"][organism].spatial[scene_id].obsl["loc"]` – `PointCloudDataFrame`. This will contain the spatial array positions for each observation, the geometry points associated to them, and additional metadata.
```
brianraymor, line 904: Editorial - This contains the spatial array positions for each observation, their associated geometry points, and additional metadata.

docs/cellxgene_census_schema.md
```diff
@@ -872,8 +881,185 @@ Cell metadata MUST be encoded as a `SOMADataFrame` with the following columns:
...
+    * MUST contain a high resolution image  – `census_obj["census_spatial_sequencing"][organism].spatial[scene_id].img[library_id]["highres_image"]` – `MultiscaleImage`.
+
+#### Matrix Data, count (raw) matrix – `census_obj["census_spatial_sequencing"][organism].ms["RNA"].X["raw"]` – `SOMASparseNDArray`
+
+Same as non-spatial data. See the corresponding section [here](#matrix-data-count-raw-matrix--census_objcensus_dataorganismmsrnaxraw--somasparsendarray).
```
brianraymor, line 909: Editorial. For all the "Same as non-spatial data" with [here] as the link, I might rewrite to this template:

```
Spatial and non-spatial data share the [same requirements](#matrix-data-count-raw-matrix--census_objcensus_dataorganismmsrnaxraw--somasparsendarray).

```

docs/cellxgene_census_schema.md
```diff
@@ -872,8 +881,185 @@ Cell metadata MUST be encoded as a `SOMADataFrame` with the following columns:
...
+    <td>integer</td>
+    <td>It MUST be valid <code>soma_joinid</code> from <code>census_obj["census_spatial_sequencing"][organism].obs</code>.</td>
+  </tr>
+  </tr>
+  </tr>
+ </tbody>
+</table>
+<!-- markdownlint-enable reference-links-images -->
+
+If Visium ("EFO:0010961") the units for the spatial array positions are pixels from the high-resolution image (`spatial[scene_soma_joinid].img[library_id]`). Otherwise TBD.
```
brianraymor, line 1017: "TBD" ? I think this needs to be more specific for this version.

docs/cellxgene_census_schema.md
```diff
@@ -872,8 +881,185 @@ Cell metadata MUST be encoded as a `SOMADataFrame` with the following columns:
...
+    <td>float</td>
+    <td>Radius of points: <code>dimeter/2</code>. If Visium ("EFO:0010961") <code>diameter</code> MUST be <code>.uns["spatial"][library_id]['spot_diameter_fullres']</code>.  As defined in the CELLxGENE dataset schema. Otherwise TBD-TODO (else for Slide-seq it should be  0.003% of the radius occupied by the full cloud of points).</td>
+  </tr>
+ </tbody>
+</table>
+<!-- markdownlint-enable reference-links-images -->
+
+#### Images of a Scene - `census_obj["census_spatial_sequencing"][organism].spatial[scene_soma_joinid].img[library_id]` – `Collection` of `MultiscaleImage`
+
+Images of a Visium ("EFO:0010961") scene MUST adhere to the following specifications. Other assays MUST NOT have images, and MUST NOT include the `img` collection.
```
brianraymor, line 1047: You should be able to rely on / reuse the dataset schema requirement for uns.spatial[library_id] which guarantees that census builder receives validated data. For schema 5.2.0:

_Curator MUST annotate if assay_ontology_term_id is "EFO:0010961" for Visium Spatial Gene Expression and uns['spatial']['is_single'] is True; otherwise, this key MUST NOT be present._

So you could write:

This MUST only be present when there is a corresponding value in the source H5AD slot `.uns["spatial"][library_id]`, as defined in the CELLxGENE dataset schema.

docs/cellxgene_census_schema.md
```diff
@@ -872,8 +881,185 @@ Cell metadata MUST be encoded as a `SOMADataFrame` with the following columns:
...
+
+`library_id` MUST be the corresponding value in the source H5AD slot `.uns["spatial"][library_id]`, as defined in the CELLxGENE dataset schema.
+
+##### High resolution image of a Scene  – `census_obj["census_spatial_sequencing"][organism].spatial[scene_soma_joinid].img[library_id]["highres_image"]` – `MultiscaleImage`
+
+The high resolution image of a Visium ("EFO:0010961") scene MUST be included and MUST be encoded as a `SOMAImageNDArray`.
```
brianraymor, line 1053: If you're going to use this style - Visium ("EFO:0010961") - to reference to assays, then I'd recommend using the correct label - Visium Spatial Gene Expression.

docs/cellxgene_census_schema.md
brianraymor, line 904: RE `PointCloudDataFrame` - should this be `SOMAPointCloudDataFrame`

docs/cellxgene_census_schema.md
```diff
@@ -872,8 +881,185 @@ Cell metadata MUST be encoded as a `SOMADataFrame` with the following columns:
...
+  <tr>
+    <td>data</td>
+    <td>bool</td>
+    <td>It MUST be <code>True</code> if the scene contains spatial information about the oberservation, otherwise it MUST be <code>False</code>.</td>
+  </tr>
+  </tbody>
+</table>
+<!-- markdownlint-enable reference-links-images -->
+
+#### Positions array of a Scene – `census_obj["census_spatial_sequencing"][organism].spatial[scene_id].obsl["loc"]` – `PointCloudDataFrame`
```
brianraymor, line 980: RE `PointCloudDataFrame` - should this be `SOMAPointCloudDataFrame`

docs/cellxgene_census_schema.md
```diff
@@ -872,8 +881,185 @@ Cell metadata MUST be encoded as a `SOMADataFrame` with the following columns:
 </tbody>
 </table>
 
+### Census Spatial Sequencing Data – `census_obj["census_spatial_sequencing"][organism]` – `SOMAExperiment`
+
+Only Visium and Slide-seq are supported for spatial data. [See the "assays included" section above](assays).
```
brianraymor, line 886: Editorial - more specific assay references.

docs/cellxgene_census_schema.md
```diff
@@ -872,8 +881,185 @@ Cell metadata MUST be encoded as a `SOMADataFrame` with the following columns:
...
+
+Spatial data for *Homo sapiens* MUST be stored as a `SOMAExperiment` in `census_obj["census_spatial_sequencing"]["homo_sapiens"]`.
+
+Spatial data for *Mus musculus* MUST be stored as a `SOMAExperiment` in `census_obj["census_spatial_sequencing"]["mus_musculus"]`.
+
+For each organism the `SOMAExperiment` MUST contain the following:
+
```
brianraymor, line 893: Editorial - I would cut much/all of this section since it summarizes requirements that are stated with more clarity below while requiring reviewers/readers to read information twice in different forms.

For example:

* Obs to spatial data mapping:
  * Obs to spatial data – `census_obj["census_spatial_sequencing"][organism].obs_spatial_presence`. It indicates the link between an observation and a scene, it MUST have three columns: 1) `soma_joinid` corresponding to `soma_joinid` of `obs` and 2) `scene_id` corresponding to the associated scene and 3) `data` which is a boolean value indicating presence.

and then:

#### Obs to spatial mapping –  `census_obj["census_spatial_sequencing"][organism].obs_presence_matrix` – `SOMADataFrame`

It indicates the link between an observation and a scene.  Each row corresponds to an observation with the following columns:


<table>
<thead>
  <tr>
    <th>Column</th>
    <th>Encoding</th>
    <th>Description</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>soma_joinid</td>
    <td>int</td>
    <td>It MUST be valid <code>soma_joinid</code> from <code>census_obj["census_spatial_sequencing"][organism].obs</code>.</td>
  </tr>
  </tr>
    <td>scene_id</td>
    <td>string</td>
    <td>It MUST be valid <code>scene_id</code> from <code>census_obj["census_spatial_sequencing"][organism].spatial</code>.</td>
  </tr>
  <tr>
    <td>data</td>
    <td>bool</td>
    <td>It MUST be <code>True</code> if the scene contains spatial information about the oberservation, otherwise it MUST be <code>False</code>.</td>
  </tr>
  </tbody>
</table>

docs/cellxgene_census_schema.md
ivirshup: It's a little different here since we are not including the full resolution image that might be present in the original h5ad.

docs/cellxgene_census_schema.md
ivirshup: Just to clarify, would you like me to change `Visium ("EFO:0010961")` to `Visium Spatial Gene Expression` or `Visium Spatial Gene Expression ("EFO:0010961")`?

docs/cellxgene_census_schema.md
jychien: It is my understanding that the x,y positions are pixels from the full resolution image.

docs/cellxgene_census_schema.md
```diff
@@ -872,8 +881,185 @@ Cell metadata MUST be encoded as a `SOMADataFrame` with the following columns:
...
+  * Matrix  data –  `census_obj["census_spatial_sequencing"][organism].ms["RNA"].X` – `SOMACollection`. It MUST contain exactly one layers:
+    * Count matrix – `census_obj["census_spatial_sequencing"][organism].ms["RNA"].X["raw"]` – `SOMASparseNDArray`
+  * Feature metadata – `census_obj["census_spatial_sequencing"][organism].ms["RNA"].var` – `SOMAIndexedDataFrame`
+  * Feature dataset presence matrix – `census_obj["census_spatial_sequencing"][organism].ms["RNA"]["feature_dataset_presence_matrix"]` – `SOMASparseNDArray`
+* Obs to spatial data mapping:
+  * Obs to spatial data – `census_obj["census_spatial_sequencing"][organism].obs_spatial_presence` – `SOMADataFrame`
+* Spatial data  –  `census_obj["census_spatial_sequencing"][organism].spatial` – `SOMACollection`.
+  * Spatial Scenes with spatial data  –  `census_obj["census_spatial_sequencing"][organism].spatial[scene_id]`  – `SOMAScene`.  There will be as many as Spatial Scenes as  spatial datasets. Each`SOMAScene` MUST contain the following:
```
jychien, line 903: I am not familiar with the term "scene". Is this commonly used? Or else would be nice if it was defined in "Definitions" section of this document.

docs/cellxgene_census_schema.md
```diff
@@ -872,8 +881,185 @@ Cell metadata MUST be encoded as a `SOMADataFrame` with the following columns:
...
+<tbody>
+  <tr>
+    <td>soma_geometry_type</td>
+    <td>str</td>
+    <td>MUST be "radius"</td>
+  </tr>
+  </tr>
+    <td>soma_geometry</td>
+    <td>float</td>
+    <td>Radius of points: <code>dimeter/2</code>. If Visium ("EFO:0010961") <code>diameter</code> MUST be <code>.uns["spatial"][library_id]['spot_diameter_fullres']</code>.  As defined in the CELLxGENE dataset schema. Otherwise TBD-TODO (else for Slide-seq it should be  0.003% of the radius occupied by the full cloud of points).</td>
```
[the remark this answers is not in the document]
jychien: is `Radius of points` a value that is applicable Slide-seq? It makes sense for Visium because each spot is 55 µm in diameter and contains multiple cells. For Slide-seq, the full "cloud of points" is not consistent from array to array. For example, this [dataset](https://cellxgene.cziscience.com/e/90ef2377-e4d0-41c7-924f-6f9ee7a9237d.cxg/) is not round. If applied to Slide-seq, I would say it most closely resembles the resolution. I'll research some more, if this is something of interest.

docs/cellxgene_census_schema.md
ivirshup: I am having trouble finding a concise definition, but I believe it's a common term in computer graphics. There a scene would be the entire thing being rendered. E.g. a scene graph is a tree representation of all the objects to render.

In the SOMA specification, this is defined as:

> SOMAScene: a specialization and extension of SOMACollection that stores spatially resolved data that can be registered to a single coordinate space.

So I could do:

> scene: a collection of spatially resolved data that can be registered to a single coordinate space

?

docs/cellxgene_census_schema.md
ivirshup: Ah, looks like it's from the SOMA spec

docs/cellxgene_census_schema.md
ivirshup: I would still be inclined to just not specify here, since we don't specify anything about this in the cellxgene specification. E.g. I don't think we would say it's invalid if the submitted positions had been converted to µm.

docs/cellxgene_census_schema.md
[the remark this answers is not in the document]
ivirshup: I believe that it's more of a technical artifact that we are providing a radius here. I do think it's important to make clear that the radius we are setting for slide-seq points is arbitrary, while the visium radius is something based on the actual physical assay.

docs/cellxgene_census_schema.md
jychien: I see, yes, that definition looks good to me. Thanks!

docs/cellxgene_census_schema.md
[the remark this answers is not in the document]
brianraymor: For this version, I would document the context / current implementation for Slide-Seq. 

BTW, I looked at Seurat which is also a _small constant_:

**Get Spot Radius**

```
Radius.SlideSeq <- function(object, ...) {
  return(0.005)
}
```

Editorial - <code>dimeter/2</code> -> diameter

Per @jychien's assessment and future schemas, was there a discussion whether it was applicable? Perhaps a question for the CB UXR in Q2?

docs/cellxgene_census_schema.md
brianraymor: Visium Spatial Gene Expression ("EFO:0010961") or "EFO:0010961" for _Visium Spatial Gene Expression_

docs/cellxgene_census_schema.md
jychien: Making sure I am understanding you correctly. Are you saying that you want to omit the entire line 1017? FYI, the X and Y coordinates are based off of the [tissue_positions.csv](https://www.10xgenomics.com/support/software/space-ranger/3.0/analysis/spatial-outputs#tissue-positions) file (which are in relation to the full res image) are used in adata.obsm['spatial'], per CxG schema 5.3. 

I'm not sure I am fully understanding what the statement is saying. Have the units for the positions been scaled to the high res image?

docs/cellxgene_census_schema.md
brianraymor: That's where I discovered it. May be helpful to have links to the SOMA specification for readers/reviewers where appropriate.

docs/cellxgene_census_schema.md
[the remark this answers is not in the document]
jychien: I think it would be helpful to have a high level definition of what information `soma_geometry` or `Radius of points` is trying to capture. 

- Is this the number of pixels?
- spot_diameter_fullres refers to the number of pixes that span the diameter of the spot in the full resolution image. Is the full resolution image being made available?
- Or has this been scaled from the full res image to fit the high res? 
- The beads for Slide-seq are 10-μm. And there is no images for this assay. It's still unclear to me as to how 'Radius of points' would be defined for Slide-seq.

docs/cellxgene_census_schema.md
ivirshup: Ah, no. I think we should keep what is said about Visium. I would like to not have any statement about what units are used for Slide-seq

ivirshup: > Have the units for the positions been scaled to the high res image?

Ah, also no. Good catch. We store a scaling transformation from the fullres pixel space to the hires pixel space. So this should say:

```suggestion
If Visium ("EFO:0010961") the units for the spatial array positions are pixels from the full-resolution image.
```

Though we could also just say something more like: "these are the coordinates from `.obsm["spatial"]` as defined in the cellxgene schema".

jychien: Yes, either would work. Both makes sense to me. Thanks!
On a side note, is there a particular reason the full res image is not accessible from the census?

ivirshup: > On a side note, is there a particular reason the full res image is not accessible from the census?

When we were prototyping we had some issues around alignment of the full res image and the hires image not being consistent across datasets.

jychien: Oh! I didn't realize there were alignment issues with some of the images. 
> not being consistent across datasets.

Not entirely sure I know what this means, but happy to help to see if it was a curation issue when we were wrangling the full resolution image. FWIW, there has been feedback that the full resolution image is much more useful in computational and pathology analyses. Can discuss offline if you would like.

docs/cellxgene_census_schema.md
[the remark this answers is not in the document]
ivirshup: @jychien

> Is this the number of pixels?

Radius is defined in the same units as the positions.

> Or has this been scaled from the full res image to fit the high res?

There is a scaling transformation for how to translate between fullres pixel space and hires pixel space. That is used to allow for plotting of locations on top of the image.

> The beads for Slide-seq are 10-μm. And there is no images for this assay. It's still unclear to me as to how 'Radius of points' would be defined for Slide-seq.

I believe the choice of radius for slide seq was arbitrary based on it being a reasonably good size for plotting. The specific value as is right now was chosen before I joined CZI, so I can't tell you much more than that.

----

Updated with typo fix:

```suggestion
    <td>Radius of points: <code>diameter/2</code>. If Visium ("EFO:0010961") <code>diameter</code> MUST be <code>.uns["spatial"][library_id]['spot_diameter_fullres']</code>, as defined in the CELLxGENE dataset schema. For Slide-seqV2 the radius is a small constant.</td>
```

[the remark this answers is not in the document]
jychien: Ah, I see. Very much appreciate you taking the time to answer all my questions, and sounds like it is there to help inform plotting of size of dots. The final suggestion looks good to me.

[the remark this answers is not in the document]
yyou1996: Thank you for the great discussion. @ivirshup I have a question for the slide-seq v2 data downloaded from the API: What is the unit of geometry points (GeometryArray, storing their locations)? Is it also μm?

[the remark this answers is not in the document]
yyou1996: @jychien and also apologize for bothering you. Just in case you by chance know this piece of information of the above question.

docs/cellxgene_census_schema.md
brianraymor: That makes sense. Consider this comment resolved then.

docs/cellxgene_census_schema.md
brianraymor: I think this is fine for the time being. 

This section will require a rewrite when we update the 5.3 EFO changes to Visium - descendant of ["EFO:0010961"](https://www.ebi.ac.uk/ols4/ontologies/efo/classes?obo_id=EFO%3A0010961) for _Visium Spatial Gene Expression_ so further editorial changes can occur then.

brianraymor (approved):

PR #1391, opened 2025-04-14
[docsite] fix pytorch notebook formatting

mlin:
SEE: https://chanzuckerberg.github.io/cellxgene-census/notebooks/experimental/pytorch.html

This diff avoids some issues our ipynb-to-docsite pipeline seems to have with nested formatting (code blocks inside links) although though they looked right in Jupyter and GitHub.


ivirshup (approved):
Looks like it's fixed when I do a test build.

Only change I would suggest, and up to you to implement, is using inter sphinx instead of hardcoding links. This would allow us to link to classes by name, and have the actual url be resolved at build time. The main benefit here is that if tiledb re-organizes their docs, we just need to do a rebuild instead of figuring out where things moved.

In practice this means moving links from looking like:

```
[ExperimentAxisQuery](https://tiledbsoma.readthedocs.io/en/stable/python-tiledbsoma-experimentaxisquery.html)
```

to

```
{class}`~tiledbsoma.ExperimentAxisQuery`
```

You would need to register `tiledbsoma-ml` in `docs/conf.py` to get links to that working.

mlin:
@ivirshup Thanks...I spent a little time trying to get that to work, and failed completely, and couldn't find working examples in the other notebooks. Would you care to take a look?

This is the update (different branch than this PR): https://github.com/chanzuckerberg/cellxgene-census/compare/mlin/fix-pytorch-notebook-formatting...mlin/fix-pytorch-notebook-formatting-intersphinx?expand=1

When I `make build` that locally, the references look just like they do in the GitHub render, showing e.g. {class}`~sklearn.preprocessing.LabelEncoder`

ivirshup:
Hmm. I am able to get it working when I only move the `tiledbsoma` links over to inter sphinx, but it seems a little fragile. Not totally sure what is going on.

I don't think this is worth a ton of effort at the moment, so feel free to merge as is.

PR #1416, opened 2025-08-20
fix markdownlint issues in draft schema

mlin:


mlin:
@brianraymor Just going at CI issues as I hit them. I did `brew install markdownlint-cli2`, `markdownlint-cli2 --fix dev-docs/cellxgene_census_draft_schema.md`, and manually fixed a few remaining things

dev-docs/cellxgene_census_draft_schema.md
```diff
@@ -1169,26 +1169,27 @@ The high resolution image of a Visium Spatial Gene Expression V1("EFO:0022857")
...
+  * Added _Pan troglodytes_
+  * Added _Rattus norvegicus_
+  * Added _Sus scrofa_
 * Multi-species data constraints
-    * Deleted section due to deprecated requirements for datasets containing multiple species or orthologous gene references
+  * Deleted section due to deprecated requirements for datasets containing multiple species or orthologous gene references
 * Assays
   * PENDING
-* Census table of organisms – census_obj["census_info"]["organisms"]
+* Census table of organisms – `census_obj["census_info"]["organisms"]`
```
brianraymor, line 1190: Don't agree with the change.

mlin:
@brianraymor Thanks, what we could do is add draft files like this to a `.markdownlintignore` file which works similarly to `.gitignore`. Although I too would rather not have spent time on this, I don't mind keeping it across the broader repo since it can fix most of its own complaints automatically, and what it can't AI can. (Merging this now as I endeavor to isolate all current build issues, but happy to add that afterwards)

PR #1417, opened 2025-08-20
[python] temp pin xarray, spatialdata, and pyarrow

mlin:


ivirshup (approved):
Hey! These changes look fine to me, so I am approving, especially since I will be out next week and I want to make sure you are unblocked here. I think I know where the failures are coming from, and also have a couple questions:

## Test failures

The test failures look like they are caused by changes in pyarrow (https://github.com/apache/arrow/pull/46423) where `.to_scipy` now returns a sparse array class instead of a sparse matrix class.

While it's not great that it's a breaking change, I do think this is a good change overall. The sparse arrays classes in scipy are a much needed step forward (tbh, I did have some part in pushing these classes to be developed). I am actually very pleasantly surprised that these are the only tests failing from that change.

Some thoughts on paths forward:

* This change is nice, just update the tests to do an instance check on `sparse.csr_array` or `(sparse.csr_array, sparse.csr_matrix)`. Assuming there are not more downstream failures after this check.
* Pin pyarrow, then figure out how we do a release around this.
* Maybe we want a compat preserving argument or setting for all functions that return sparse objects? Though I think maybe this should be coordinated with tiledbsoma.

## Questions

* Could you give a bit more detail on the versioning issues you're running into with spatialdata and xarray?
* What does AOTW mean here?

PR #1420, opened 2025-08-26
builder 2025 increment1

mlin:
`cellxgene_census_builder` updates for 2025Q3 "increment 1" (of several, as scoped by @brianraymor):
- Advance nominal `CENSUS_SCHEMA_VERSION` to 2.4.0
- Use CxG schema 6.0.0 (while 7.0.0 is pending)
- Replace spatial assay `EFO:0010961` with `EFO:0022857`
- Source `organism` and `organism_ontology_term_id` from `uns` instead of `obs` in the h5ad inputs
    - Triggered minor rewiring in the summary stats calculations

Unit tests pass and I've separately run a slightly larger test with 250 datasets.

Merging this to main will make upcoming weekly Census builds reflect the changes (since it uses a docker image updated from main). I'd like to go that way to leverage that automation, but an alternative is to start an epic branch and run full builds off it separately -- open to discussion


ivirshup:
I'd add that my preference would be to do a feature branch and not push this to update the public builds until:

* Increment 2 is implemented (in case any builder changes come up)
* We do at least one full build (let's us sanity check the artifact and push all the cases through the validator)

brianraymor:
@ivirshup - per my comments in slack, I have updated a copy of the census schema in **dev docs** - https://github.com/chanzuckerberg/cellxgene-census/blob/main/dev-docs/cellxgene_census_draft_schema.md - it currently contains the changes to normative requirements for the increments. I'm also planning a complete rewrite per - https://czi-sci.slack.com/archives/C08MF1AJ6F5/p1755795878818459?thread_ts=1755722007.686419&cid=C08MF1AJ6F5

I also do not like that the census schema is updated _in place_ unlike the CELLxGENE schema which has version subdirectories.

PR #1424, opened 2025-10-07
[python] fix new mypy errors

mlin:


ivirshup (approved):
LGTM, with one possible change.

PR #1440, opened 2025-11-07
[python] builder display_summary(): include spatial collections

mlin:
@jahilton noticed that the builder's summary pivot table didn't include spatial datasets, making it incomplete and also discrepant with the total_cell_count shown in the same report (which did count spatial). This diff includes spatial in the pivot table, breaking it out by collection.

Revised example summary: 

```
                 label      value
 census_schema_version      2.4.0
     census_build_date 2025-11-04
dataset_schema_version      7.0.0
      total_cell_count  217768036
     unique_cell_count  125463259
                                             unique count                                       
attribute                                          assays cell types      cells datasets tissues
organism           collection                                                                   
callithrix_jacchus census_data                          1         40    2275451       27      33
homo_sapiens       census_data                         37        898  158982719     1086     417
                   census_spatial_sequencing            2        228    3042411      361      49
macaca_mulatta     census_data                          2         54    7010229       19      29
mus_musculus       census_data                         16        473   43653561      172     101
                   census_spatial_sequencing            2         67    2645566      179       6
pan_troglodytes    census_data                          1         25     158099        1       1
```

(2275451+158982719+3042411+7010229+43653561+2645566+158099=217768036)

AFAIK this particular report was only used internally, but it was good to explore this because it does raise the question of whether the `census_obj​​["census_info"]["summary"]`, `census_obj["census_info"]["summary_cell_counts"]`, and other `census_info` tables should similarly count the spatial and non-spatial collections separately for each organism, instead of lumping them together as they do now. A separate issue we can file @brianraymor

### DIFF
--- a/tools/cellxgene_census_builder/src/cellxgene_census_builder/census_summary.py
+++ b/tools/cellxgene_census_builder/src/cellxgene_census_builder/census_summary.py
@@ -6,7 +6,7 @@ import cellxgene_census
 import pandas as pd
 from tiledbsoma import DataFrame
 
-from .build_soma.globals import CENSUS_DATA_NAME, CENSUS_INFO_NAME
+from .build_soma.globals import CENSUS_DATA_NAME, CENSUS_INFO_NAME, CENSUS_SPATIAL_SEQUENCING_NAME
 
 # Print all of the Pandas DataFrames, except the dimensions
 pd.options.display.max_columns = None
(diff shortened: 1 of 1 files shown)


pablo-gar (approved):

ivirshup (commented):
I don't think I have looked at this file before 😅 

I think I am okay with this code not being tested since AFAICT it's just for display at build time.

If there are equivalent statistics in `census_summary` I would like a follow up issue opened that removes this code and uses that instead.

mlin:
Agreed, filed #1442

PR #1446, opened 2025-12-05
docs updates for 2025-11-08 LTS

mlin:
* `docs/cellxgene_census_docsite_landing.md`: add a banner announcement for new organisms
* `docs/cellxgene_census_docsite_data_release_info.md`:
    * add section for the new build
        * call out the schema changes with pluralization of `disease` and non-unique `feature_name`
        * add commented scripts for producing the stats tables (it's more involved now with the additional organisms and pluralization)
    * adjust verbiage about release policy/cadence to be more realistic for 2026
* `api/python/notebooks/analysis_demo/comp_bio_census_info.ipynb`: rerun on new build and add a note about the `disease` pluralization
* `api/python/notebooks/api_demo/census_duplicated_cells.ipynb`: fix a historical typo (human/mouse swap), rerun on new build and add a note about the `disease` pluralization


mlin:
There were a couple notebooks that we meant to refresh on the new build, but ran into complications, and I'll hereby make the strawperson proposal to let them stand as-is. (The main reason for wanting to revise these was to use `feature_id` instead of `feature_name` where it needs to be unique.)

* [comp_bio_explore_and_load_lung_data](https://chanzuckerberg.github.io/cellxgene-census/notebooks/analysis_demo/comp_bio_explore_and_load_lung_data.html): there is a section heavily using the feature/dataset presence matrix, which was changed earlier this year to be denser than it had been when the notebook was written. See #1320. The change affects the presentability of the results (much longer than originally intended). If pressed, we could try to think of compensating adjustments in the demo scenario.
* [comp_bio_embedding_exploration](https://chanzuckerberg.github.io/cellxgene-census/notebooks/analysis_demo/comp_bio_embedding_exploration.html): written at the 'high water mark' of diverse contributed embeddings, and the notebook would not be as interesting if limited to the few embeddings available for the new build.

pablo-gar (approved):

docs/cellxgene_census_docsite_landing.md
```diff
@@ -1,8 +1,6 @@
-<span style="background-color: #f3bfcb; color; font-size: 18px"> 🚀 Now in testing: Spatial! From Jan 16th, latest builds will include data from Slide-seq and Visium assays. ⚠️ Opening these builds requires `tiledbsoma>=1.15.3` ⚠️. [Learn more](articles/2025/20250117-spatial.md)!
+<span style="background-color: #f3bfcb; color; font-size: 18px"> 🚀 Census LTS release `2025-11-08` includes CELLxGENE Discover datasets from _Macaca mulatta_, _Callithrix jacchus_, and _Pan troglodytes_ as well as _Homo sapiens_ and _Mus musculus_.</span>
```
tihuan: For my own education, for a while Census was always a subset of the full CxG corpus, namely only having human and mouse datasets. With this new version, is it still the case that Census is still a subset of the full CxG corpus?

Thank you!

tihuan (approved):

PR #1448, opened 2025-12-09
[python] fix unit tests for stable 2025-11-08

mlin:
A few test cases use the `stable` label and assumptions needed updating when we advanced that label to the new build.


ivirshup:
The assumption here was that a sparse array had more than 400 entries? What array had less than that?

kliu-czi (approved):

